PR #15, opened 2023-09-23
more tests will pass

jaytaph:
Fixed a few more tests. Still needs some additional work on the rest of the tests, mostly the token characters.


emwalker:
Would it make sense to move most of these tests onto a branch of work in progress, so that `main` compiles?  I think people might be scared away from helping if `main` is in a broken state.

PR #24, opened 2023-09-24
Remove redundant clone() call.

glomdom:
Calling clone() on a &str (str reference) will result in a noop instruction, as str doesn't implement Clone, so we'd only be copying the reference which does nothing.


jaytaph:
@glomdom Thank you for your PR!

PR #26, opened 2023-09-24
Fixup on construction active element algorithm

jaytaph:
This PR tries to get the active element reconstruction running properly. It still fails horribly on the `<a><p>X<a>Y</a>Z</p></a>` test from test1.dat in the html5lib-tests repo. There are still some function implementations missing in the parser (mod.rs).

### DIFF
--- a/src/bin/parser_test.rs
+++ b/src/bin/parser_test.rs
@@ -54,9 +54,9 @@ fn main () -> io::Result<()> {
 
         let mut test_idx = 1;
         for test in tests {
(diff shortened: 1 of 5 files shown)


tamimbost500:
fg

jaytaph:
I've updated the parser so it's in line with the main branch. I think i'm going to get rid of all the test-code so the current state as-is, is mergeable, even though not completely functioning. This way I think more people can take a look at the code and try to get it right.

PR #27, opened 2023-09-24
Add tests/tokenizer and vendor html5lib-tests

emwalker:
This PR:
* Migrates `src/bin/tokenizer_test.rs` to `tests/tokenizer.rs` to show one possible approach that could be used here
* Vendors `html5lib-tests` under `tests/data/html5lib-tests` in order to provide stable test input that works the same for everyone right after cloning

We lose a little bit of granularity in the error reporting, although it might be possible to recover this.  In turn we gain a test suite that runs with `cargo test` that can protect against regressions.

For the substantive changes without the vendoring, see https://github.com/jaytaph/gosub-browser/pull/27/commits/6e428732ed96e4c057ec18339c94292c3149f93c.


emwalker:
Errors in `test3.test`, `unicodeCharsProblematic.test` and `xmlViolation.test`, which are all commented out, represent future work that will need to be carried out until these tests pass.

jaytaph:
@emwalker  Thanks for the pr!

PR #31, opened 2023-09-25
 refactor(root): use rust library docs comments instead of regular comments

neuodev:


jaytaph:
@neuodev Thanks for the PR. very much appreciated as I'm new to rust so I don't know all the ins and outs yet :-)  This seems like a very useful addition and we should fix them (is this something we can check with a linting process during CI maybe?)

Could you maybe sign your commits, as we can only merge signed commits into the main branch. See [Learn more about signing commits.](https://docs.github.com/articles/about-gpg/)

jaytaph:
It sounds that we can automatically deploy the docs onto the site?  (ie: https://docs.dev.gosub.io or something?)

jaytaph:
@neuodev Thank you for the PR!

PR #32, opened 2023-09-25
feat(parser): exit when html5lib-tests dir is missing

neuodev:


jaytaph:
ow, I see we the commits are not signed. Could you rebase your commits using signing? See [Learn more about signing commits.](https://docs.github.com/articles/about-gpg/)

emwalker:
@neuodev note that `html5lib-tests` is now vendored under [tests/data/html5lib-tests](https://github.com/jaytaph/gosub-browser/tree/main/tests/data/html5lib-tests).

jaytaph:
So maybe this PR could include that we are not defaulting to ./html5lib-tests, but the ./tests/data/html5lib-tests directory instead?  (and worst-case give a failure if that directory is not found)

src/bin/parser_test.rs
```diff
@@ -29,6 +30,11 @@ fn main() -> io::Result<()> {
     let default_dir = "./html5lib-tests";
     let dir = env::args().nth(1).unwrap_or(default_dir.to_string());
 
+    if !Path::new(&dir).exists() {
+        println!("Missing 'html5lib-tests'. It is no located at '{dir}' \nIt should be cloned first from here https://github.com/html5lib/html5lib-tests");
```
ernest-rudnicki, line 34: Hey there, I noticed a small typo

```suggestion
        println!("Missing 'html5lib-tests'. It is not located at '{dir}' \nIt should be cloned first from here https://github.com/html5lib/html5lib-tests");
```

jaytaph (approved):

jaytaph:
@neuodev Thanks for the PR!

PR #53, opened 2023-09-27
simple user agent that only prints body

ynewmark:
Adds a simple user agent that prints just the text within the body tag, all tags within that are treated the same.

The main component of this is a function `get_node_by_path` which accepts a document and vector of tag names and returns the node at that path if it exists. It might make sense to move this function to the library. Though if we add css selectors that might replace this function.


jaytaph (approved):

jaytaph:
@ynewmark Nice pr!  Could you try and see if you can fix the CI issues (clippy / fmt)

ynewmark:
Can you rerun the workflow?

jaytaph:
All green. @ynewmark Thanks for the PR!

Issue #59, opened 2023-09-28
Make sure we cannot create recursive node trees


jaytaph:
Right now it is possible to create a node with a child that points to itself. This causes an endless loop.

We have to make sure we cannot add nodes that result in such a loop.

mihna123:
Would it just be enough to make sure that the children ids are never equal to the parent id?

mihna123:
But then i guess it doesn't work if it goes more than one level deep

lucalewin:
I would implement a arena using a `Vec` like this:

```rust
pub struct Arena<T> {
    nodes: Vec<Node<T>>
}

pub struct Node<T> {
    pub id: usize,
    pub parent: Option<usize>,
    pub children: Vec<usize>,
    pub data: T
}

impl<T> Arena<T> {
    pub fn new() -> Self {
        Self { nodes: Vec::new() }
    }

    pub fn get(&self, id: usize) -> Option<&Node<T>> {
        self.nodes.get(id)
    }

    pub fn get_mut(&mut self, id: usize) -> Option<&mut Node<T>> {
        self.nodes.get_mut(id)
    }

    pub fn add(&mut self, data: T) -> usize {
        let id = self.nodes.len();

        self.nodes.push(Node { id, parent: None, children: Vec::new(), data });

        id
    }

    pub fn attach(&mut self, parent_id: usize, child_id: usize) {
        if let Some(parent) = self.nodes.get_mut(parent_id) {
            parent.children.push(child_id);
        }

        if let Some(child) = self.nodes.get_mut(child_id) {
            child.parent = Some(parent_id);
        }
    }
}
```

Then we could create a `DomNode` struct:

```rust
pub struct DomNode {
    /// name of the node, or empty when it's not a tag
    pub name: String,
    /// namespace of the node
    pub namespace: Option<String>,
    /// actual data of the node
    pub data: NodeData,
}

/// Different type of node data
#[derive(Debug, PartialEq, Clone)]
pub enum NodeData {
    Document,
    Text {
        value: String,
    },
    Comment {
        value: String,
    },
    Element {
        name: String,
        attributes: HashMap<String, String>,
    },
}
```

This can then be used like this:

```rust
let mut arena = Arena::<DomNode>::new();

let node_id = arena.add(DomNode {
    name: "".into(),
    namespace: None,
    data: NodeData::Document
});
```

This has the advantage that the arena can be used for different node types (could be reused for the css parser)

[...]

PR #60, opened 2023-09-28
Make sure we can't add node that makes a loop

mihna123:
Added a recursive function to search for nodes that have their parent as a child. This will solve issue #59


mihna123:
Dang

mihna123:
Didn't have commits signed

PR #61, opened 2023-09-28
Make sure there are no node reference loops

mihna123:
Attempts to fix issue #59 again


jaytaph:
@mihna123 This is really nice to have in order to make sure we cannot have endless loops 👍 

Could you see if you can fix the clippy and fmt? You can use "make test" to check it locally.

jaytaph:
@mihna123 thanks for the pr!

mihna123:
@jaytaph You are welcome!

PR #62, opened 2023-09-29
Add basic benchmarking using criterion

emwalker:
This PR adds a simple benchmarking harness as an example of the kind of thing that can be done using a library called [criterion](https://github.com/bheisler/criterion.rs).

The `criterion` library is a well-known Rust library that is often used for this kind of thing.  The benchmark in this PR is just an example; hopefully it will be straightforward to extend to more substantial cases.  Its comparisons from run to run are a little noisy, but I suspect this can be dialed in.  Examples of people using `criterion` [here](https://youtu.be/3oL1xokuHBE?si=QXfqvjiszlfJ2ds1&t=10176) and [here](https://fasterthanli.me/series/advent-of-code-2022/part-17#performance-comparison-with-criterion).

To run the suite, do:

```sh
$ make bench
$ cargo bench
```

I'll leave this PR in draft since I didn't discuss it first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
+name = "atty"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
+dependencies = [
+ "hermit-abi 0.1.19",
+ "libc",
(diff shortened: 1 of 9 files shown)


jaytaph:
Awesome. Looking forward to merge

emwalker:
This PR should be ready to review and merge now.  The PR is just a placeholder.  Feel free to make significant changes, or refactor, or even revert the functionality if it doesn't prove useful.

I adjusted the settings in an attempt to make the results a little more consistent from one run to another.  The lesson is that you have to run `cargo bench` several times in order to get a picture of where things are at.

jaytaph:
@emwalker  Thanks for the PR.

Is this something we should run on each PR and deploy the results? (we have the *.developer.gosub.io domain for this for now)?

emwalker:
I doubt the benchmark results are deterministic enough to block a PR on at this point.  But I can imagine in the future failing a PR check if some time threshold is passed.  I suspect the actual benchmark might need to be modified if it was going to be used for a PR check.  So perhaps good to leave for a future investigation.

PR #65, opened 2023-10-01
ci: deploy docs to github pages

neuodev:
This patch adds a new job to the GitHub workflow to generate docs and deploy them to GH pages.

The job is only triggered when the PR is merged into the "main" branch. It needs access to the "GITHUB_TOKEN" of the repo owner. It will push the "target/doc" directory to a branch called "gh-pages" in the root directory.

The repo owner will need to update the repo settings to make GitHub deploy to this directory.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,9 +2,9 @@ name: Rust CI
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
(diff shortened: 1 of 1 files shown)


jaytaph:
@neuodev  Looks good! I'll add this for now, but we might want to deploy to different systems. I have a `docs.developer.gosub.io` reserved for this, but for now gh-pages is fine. I'll try to make this work now.

PR #71, opened 2023-10-01
Add NodeId to wrap usize

emwalker:
This PR uses the [newtype pattern](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) to make the id used for HTML5 nodes an owned type.  This will allow us to enforce invariants later on against the id type, and it will make use of Rust's type system to catch cases where we're passing something into a function that shouldn't be used to identify a node.

I think I read somewhere that the Rust compiler will see that the no memory beyond a `usize` is needed and will optimize away the `NodeId` struct during compilation.

There are still some issues that are not addressed in this PR that are left for future PRs:
- I suspect we're using `0` to represent both the root node and the case where an id hasn't been assigned yet.  In Rust, the typical approach for the latter situation is to use `Option<NodeId>`, or to move the id generation above the instantiation of the new node.
- For this reason, we'll probably want to get rid of the `Default` derive on `NodeId` at some point once this ambiguity has been addressed.
- We probably want to make a collection that wraps `Vec` and accepts `NodeId` as an argument to various getters rather than using `NodeId` directly as a `usize` index in calling code

~Keeping this PR in draft because we didn't have a chance to discuss this change.~

### DIFF
--- a/src/bin/parser_test.rs
+++ b/src/bin/parser_test.rs
@@ -1,5 +1,5 @@
 use gosub_engine::html5_parser::input_stream::InputStream;
-use gosub_engine::html5_parser::node::NodeData;
+use gosub_engine::html5_parser::node::{NodeData, NodeId};
 use gosub_engine::html5_parser::parser::document::Document;
 use gosub_engine::html5_parser::parser::Html5Parser;
 use regex::Regex;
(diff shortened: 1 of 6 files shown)


jaytaph:
:+1 I think this makes a lot of sense.

emwalker:
Ok, good.  I'll move it out of draft in that case.

PR #76, opened 2023-10-02
Implemented an initial setup for the console API

jaytaph:
This is an initial setup of the [console api](https://console.spec.whatwg.org/), since i cannot spend any time on the parser since I cannot access the code (don't ask).

So therefor, I did some reading up on the console API, and found that it's not very hard to implement. Not everything is done yet, but most is.

The idea is that anything that can be displayed (implements fmt::Display), can be added to the console functions. It also contains timers, and groups.

Ultimately, the console will write to a io::Write, which can be stdout(), a file, or a string buffer. Later on, these calls will be connected to javascript methods.

I'm not sure if this is the right way to do it, so i'm hoping from some feedback from others to see if this is the right way to do it.

(i know that the console api is not the highest priority, but it gave me some time to read into more stuff, and practice some rust)


Cargo.toml
```diff
@@ -28,7 +28,11 @@ regex = "1"
...
 typed-arena = "2.0.2"
 reqwest = { version = "0.11.12", features = ["blocking"] }
+colored = "2.0.4"
+uuid = { version = "1.4.1", features = ["v4"] }
 
 [dev-dependencies]
 criterion = { version = "0.4", features = ["html_reports"] }
 test-case = "3.2.1"
+colored = "2.0.4"
+uuid = { version = "1.4.1", features = ["v4"] }
```
emwalker, line 38: Did you want the new libraries in both the regular and dev dependencies?  I think they would go in one section or the other, but not both.

src/console/buffer_printer.rs
```diff
@@ -0,0 +1,87 @@
...
+    fn test_buffer_printer() {
+        let mut printer = BufferPrinter::new();
+        printer.print("log".into(), &[&"Hello".to_string(), &"World".to_string()], &vec![]);
+        printer.print("info".into(), &[&"Something when wrong".to_string(), &"This too!".to_string()], &vec![]);
+        printer.print("warn".into(), &[&"Something when wrong".to_string(), &"This too!".to_string()], &vec![]);
+        printer.print("error".into(), &[&"Something when wrong".to_string(), &"This too!".to_string()], &vec![]);
+
+        assert_eq!(printer.buffer, 0);
+    }
+}
```
emwalker, line 87: Whitespace.

src/console/buffer_printer.rs
```diff
@@ -0,0 +1,87 @@
...
+            groups: vec![],
+            buffer: "".to_string(),
+        }
+    }
+}
+
+impl Printer for BufferPrinter {
+    fn print(&mut self, log_level: String, args: &[&dyn fmt::Display], _options: &[&str])
+    {
+        if args.len() == 0 {
```
emwalker, line 26:
```suggestion
        if args.is_empty() {
```

src/console/buffer_printer.rs
```diff
@@ -0,0 +1,87 @@
...
+            return;
+        }
+
+        match log_level.as_str() {
```
emwalker, line 30:
```suggestion
        match &log_level {
```

src/console/buffer_printer.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub fn new() -> BufferPrinter {
+        BufferPrinter {
+            groups: vec![],
+            buffer: "".to_string(),
+        }
+    }
+}
+
+impl Printer for BufferPrinter {
+    fn print(&mut self, log_level: String, args: &[&dyn fmt::Display], _options: &[&str])
```
emwalker, line 24: `log_level` should probably be an enum.  Maybe something like:

```rust
enum LogLevel {
    Debug,
    Info,
    Warn,
    Error,
}
```

emwalker: Also, keep in mind there's a [canonical logging library](https://docs.rs/log/latest/log/) that we might want to make use of as well.

src/console/buffer_printer.rs
emwalker, line 26:
```suggestion
        if args.is_empty() {
```

src/console/formatter.rs
```diff
@@ -0,0 +1,18 @@
+use std::fmt;
+
+pub struct Formatter {}
```
emwalker, line 3: The idiomatic thing to do when you don't need fields yet is to use a unit struct.

```rust
pub struct Formatter;
```

src/console/formatter.rs
```diff
@@ -0,0 +1,18 @@
...
+
+impl Formatter {
+    pub fn new() -> Formatter {
+        Formatter {}
+    }
+
+    pub fn format(&self, args: &[&dyn fmt::Display]) -> String {
+        let mut s = String::from("");
+        for arg in args {
+            s.push_str(format!("{} ", arg).as_str());
```
emwalker, line 13:
```suggestion
            s.push_str(&format!("{} ", arg));
```

src/console/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+        let mut cnt = 1;
+        if self.counts.contains_key(&label) {
+            cnt = self.counts.get(&label).unwrap() + 1;
+        }
+
+        self.counts.insert(label.clone(), cnt + 1);
+
+        let concat = format!("{}: {}", label, cnt);
+        self.logger("count", &[&concat]);
+    }
```
emwalker, line 116:
```suggestion
    fn count(&mut self, label: &str) {
        let mut cnt = 1;
        if self.counts.contains_key(label) {
            cnt = self.counts.get(&label).unwrap() + 1;
        }

        self.counts.insert(label.to_owned(), cnt + 1);

        let concat = format!("{}: {}", label, cnt);
        self.logger("count", &[&concat]);
    }
```

src/console/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    fn count_reset(&mut self, label: String) {
+        if !self.counts.contains_key(&label) {
+            self.logger("countReset", &[&"label does not exist"]);
+            return;
+        }
+
+        self.counts.insert(label.clone(), 1);
+    }
```
emwalker, line 125:
```suggestion
    fn count_reset(&mut self, label: &str) {
        if !self.counts.contains_key(label) {
            self.logger("countReset", &[&"label does not exist"]);
            return;
        }

        self.counts.insert(label.to_owned(), 1);
    }
```

src/console/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+
+        self.timers.insert(
+            label.clone(),
+            Timer {
+                label: label.clone(),
+                start,
+                end: None,
+            },
+        );
+    }
```
emwalker, line 189:
```suggestion
    fn time(&mut self, label: &str) {
        if self.timers.contains_key(label) {
            let warning = format!("Timer '{}' already started", label);
            self.logger("warning", &[&warning]);
            return;
        }

        let start = match SystemTime::now().duration_since(UNIX_EPOCH) {
            Ok(n) => n.as_millis(),
            Err(_) => 0,
        };

        self.timers.insert(
            label.to_owned(),
            Timer {
                label: label.to_owned(),
                start,
                end: None,
            },
        );
    }
```

src/console/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    /// Log time
+    fn time_log(&self, _label: String, _data: &[&dyn fmt::Display]) {
```
emwalker, line 192:
```suggestion
    fn time_log(&self, _label: &str, _data: &[&dyn fmt::Display]) {
```

src/console/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+        todo!()
+    }
+
+    /// End the given timer
+    fn time_end(&mut self, label: String) {
```
emwalker, line 197:
```suggestion
    fn time_end(&mut self, label: &str) {
```

src/console/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+        sleep(std::time::Duration::from_millis(123));
+        c.time_end("foo".into());
+        c.group_end();
+        c.group_end();
+        c.warn(&[&"Back", &"To root"]);
+
+        c.assert(true, &[&"This assertion asserts"]);
+        c.assert(false, &[&"This assertion does not assert"]);
+        c.assert(true, &[]);
+        c.assert(false, &[]);
```
emwalker, line 267: If you go with the suggestions above, this test will change.

src/console/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+
+trait Printer {
+    fn print(&mut self, log_level: &str, args: &[&dyn fmt::Display], _options: &[&str]);
+    fn clear(&mut self);
+    fn end_group(&mut self);
+}
+
+#[cfg(test)]
+
+mod tests {
```
emwalker, line 236: Drop newline on line 235.

src/console/text_printer.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub fn new(writer: W) -> TextPrinter<W> {
+        TextPrinter {
+            writer,
+            groups: vec![],
+        }
+    }
+}
+
+impl<W: Write> Printer for TextPrinter<W> {
+    fn print(&mut self, log_level: &str, args: &[&dyn fmt::Display], _options: &[&str]) {
```
emwalker, line 24:
```suggestion
    fn print(&mut self, log_level: LogLevel, args: &[&dyn fmt::Display], _options: &[&str]) {
```

Here `LogLevel` is the new enum mentioned above.

src/console/text_printer.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        let mut buffer = Vec::new();
+        let mut printer = TextPrinter::new(Cursor::new(&mut buffer));
+        printer.print("info", &[&"Foo", &2, &false], &vec![]);
+        printer.print("warn", &[&"a", &"b"], &vec![]);
+        printer.print("error", &[], &vec![]);
+        assert_eq!(
+            String::from_utf8(buffer).expect("failed to convert"),
+            "[info] Foo 2 false\n[warn] a b\n"
+        );
```
emwalker, line 95: This test would also change.

jaytaph:
@emwalker Thanks for the feedback. I've implemented the suggestions, and will try to implement the rest of the code. I might even check if it would be possible if I could create some node.js bindings, to see if I can use it from node.js, but this depends on my time available (i really want the parser done)

emwalker (approved):

src/api/console/buffer.rs
```diff
@@ -0,0 +1,28 @@
...
+
+    pub fn to_string(&self) -> Result<String> {
+        Ok(String::from_utf8(self.buf.clone())?)
+    }
+}
+
+impl Write for Buffer {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        self.buf.extend(buf);
+        Ok(buf.len())
```
emwalker, line 21: Did you want to fix this return value before merging?  If not, perhaps add a `// TODO: return correct return value`.

emwalker:
LGTM with suggestion.

PR #82, opened 2023-10-03
Prefer &str to String for function arguments

emwalker:
This PR
* Replaces `String` function arguments with the `&str` versions when possible
* Reduces the number of allocations
* Uses `String` for char buffers instead of `Vec<chr>`, since `String` is designed to be mutable and grow easily

In a perfect world, we'd have an allocation for the input, and then just move around the current buffer without allocating until DOM nodes are needed (so-called "zero copy").  But I'm not sure to what extent that's possible here or with a streaming parser.

### DIFF
--- a/src/html5_parser/node.rs
+++ b/src/html5_parser/node.rs
@@ -704,14 +704,14 @@ mod tests {
 
         let mut node = Node::new_element("name", attr.clone(), HTML_NAMESPACE);
         assert!(node.clear_attributes().is_ok());
-        assert_eq!(node.has_attributes(), false);
+        assert!(!node.has_attributes());
     }
 
     #[test]
     fn has_attributes_non_element() {
         // if node is a non-element, will always return false
         let node = Node::new_document();
-        assert_eq!(node.has_attributes(), false);
+        assert!(!node.has_attributes());
     }
 
     #[test]
@@ -722,6 +722,6 @@ mod tests {
         assert_eq!(node.has_attributes(), false);
 
         assert!(node.insert_attribute("key", "value").is_ok());
-        assert_eq!(node.has_attributes(), true);
+        assert!(node.has_attributes());
     }
 }
--- a/src/html5_parser/node.rs
+++ b/src/html5_parser/node.rs
@@ -394,7 +394,7 @@ pub static SPECIAL_MATHML_ELEMENTS: [&str; 6] = ["mi", "mo", "mn", "ms", "mtext"
 pub static SPECIAL_SVG_ELEMENTS: [&str; 3] = ["foreignObject", "desc", "title"];
 
 #[cfg(test)]
-mod test {
+mod tests {
     use super::*;
 
     #[test]
--- a/src/html5_parser/node.rs
+++ b/src/html5_parser/node.rs
@@ -190,13 +190,13 @@ impl Node {
     }
 
     /// Check if an attribute exists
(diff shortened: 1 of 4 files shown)


src/html5_parser/node.rs
```diff
@@ -398,7 +398,7 @@ mod test {
     use super::*;
 
     #[test]
-    fn test_new_document() {
+    fn new_document() {
```
emwalker, line 401: In Rust, it's fine to avoid the `test_` prefix (probably because we're already in a `tests` module).

emwalker:
@jaytaph it would be nice to have the ability to add reviewers, if you agree.  Currently this field is locked down.



@Kiyoshika here is some of the ideas I was discussing.  Let me know what you think.

src/html5_parser/node.rs
```diff
@@ -592,167 +592,136 @@ mod test {
...
-                match value {
-                    None => assert!(false),
-                    Some(val) => assert_eq!(*val, "value".to_string()),
-                }
-            }
-        }
+
+        assert!(node.insert_attribute("key", "value").is_ok());
+        let value = node.get_attribute("key").unwrap().unwrap();
+        assert_eq!(value, "value");
```
emwalker, line 632: I had a question about whether a `Result` is needed here, since we're already using an `Option`.

src/html5_parser/node.rs
```diff
@@ -592,167 +592,136 @@ mod test {
...
         let attr = HashMap::new();
 
         let mut node = Node::new_element("name", attr.clone(), HTML_NAMESPACE);
         assert_eq!(node.has_attributes(), false);
 
-        if let Ok(_) = node.insert_attribute("key".to_string(), "value".to_string()) {
-            assert_eq!(node.has_attributes(), true);
-        }
+        assert!(node.insert_attribute("key", "value").is_ok());
+        assert_eq!(node.has_attributes(), true);
```
emwalker, line 725: Here we can have more confidence that the test will fail if we ever get an `Err`.

src/html5_parser/tokenizer/mod.rs
```diff
@@ -529,7 +529,7 @@ impl<'a> Tokenizer<'a> {
                         Element::Utf8('>') => {
                             if self.is_appropriate_end_token(&self.temporary_buffer) {
                                 let s: String = self.temporary_buffer.iter().collect::<String>();
-                                self.set_name_in_current_token(s);
+                                self.set_name_in_current_token(&s);
```
emwalker, line 532: Here we're allocating a `String` and then passing a reference.  I appreciate that this is a little weird.  I'm hoping we'll figure out a way to avoid the allocation in the future, although that might be hard in this case.

Kiyoshika:
I’m not home yet but I might be in a few hours, otherwise I will try to take a look in the morning EST

jaytaph (approved):

emwalker:
@jaytaph if it's not inconvenient, I'm going to wait for @Kiyoshika to take a look as well before merging.

Kiyoshika (approved):
I'm good with all these changes, I will follow the same approach when I implement the `ElementClass` previously discussed.

Just let me know about the `Result<Option>` if we want to update that now or I can update that separately

emwalker:
I'm thinking we can figure out what to do with `Result<Option>` later on.

Kiyoshika:
works for me, thanks!

PR #92, opened 2023-10-03
Handle classes within Node

Kiyoshika:
This PR introduces utility methods to work with class names within a node according to [this issue](https://github.com/jaytaph/gosub-browser/issues/79) and [this discussion](https://github.com/jaytaph/gosub-browser/discussions/73). This will be useful for user agents / renderers to check which/if classes need to be applied to a given element and will also be used in a future PR for querying the DOM by class name.

Note that I am not doing any "error handling" in `ElementClass` since this mimics the behaviour when using javascript.

For instance, if you try 

```javascript
var element = document.getElementById("myDIV");
element.classList.toggle("mystyle");
```

on a class that doesn't exist, it will essentially do nothing (at least, from the end user's perspective.) I'm not sure if it's worth it to throw an actual error if trying to remove/edit a class that doesn't exist, so for now they are silent (but intentional) errors.

From the parser, when a new node is created & added to the document, we check if a `class` attribute exists and take its string which is split on whitespace (extra whitespaces are ignored) and an `ElementClass` struct is created to hold the classes which is stored inside the `Node` struct. We can then toggle/add/remove classes from the node using the interface provided.

EDIT: this also simplifies the `Node::get_attribute` and `Node::get_mut_attribute` methods to return an `Option` instead of a `Result<Option>`. Even though this is an unrelated change, I don't think it was worth creating an entirely separate PR for this and snuck it into this one


Kiyoshika:
I just realised I pulled off of my outdated main (hence the conflict), going to rebase before continuing

Kiyoshika:
re-synced fork, added some starter tests but didn't have much time today to work on it, hopefully can make more progress tomorrow

Kiyoshika:
PR finished and ready to review

src/html5_parser/element_class.rs
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        ElementClass {
+            class_map: class_map_local,
+        }
+    }
+
+    /// Count the number of classes (active or inactive)
+    /// assigned to an element
+    pub fn count(&self) -> usize {
```
blaumeise20, line 41: Wouldn't it be better to stick with the Rust style here and use `len` as the method name? Especially because there is an `is_empty` method, which gets suggested by clippy when doing any comparison to zero, e.g. `list.len() == 0`.

Kiyoshika: yes, good suggestion. i'm still not entirely familiar with all of Rust's conventions yet. Will fix now

Kiyoshika: updated in [98e6130](https://github.com/gosub-browser/gosub-engine/pull/92/commits/98e61304d3836a0769df4317afe5db9006af37e3)

jaytaph (approved):

src/html5_parser/node.rs
```diff
@@ -246,32 +254,32 @@ impl Node {
...
+            return None;
         }
 
         let mut value: Option<&String> = None;
         if let NodeData::Element { attributes, .. } = &self.data {
             value = attributes.get(name);
         }
 
-        Ok(value)
+        value
```
emwalker, line 267: We'll have to keep our eye on whether there are error cases in CSS that need to be handled.  But this change feels like a good next step.  I'm thinking we can backfill error handling when we discover that it's needed.

src/html5_parser/element_class.rs
```diff
@@ -0,0 +1,157 @@
...
+        self.class_map.contains_key(name)
+    }
+
+    /// Add a new class (if already exists, does nothing)
+    pub fn add(&mut self, name: &str) {
+        // by default, adding a new class will be active.
+        // however, map.insert will update a key if it exists
+        // and we don't want to overwrite an inactive class to make it active unintentionally
+        // so we ignore this operation if the class already exists
+        if !self.contains(name) {
```
emwalker, line 61: Probably want to keep an eye on this kind of substring matching.  Might be fine indefinitely, or might cause problems later on.

src/html5_parser/parser/mod.rs
```diff
@@ -3679,4 +3692,68 @@ mod test {
...
+
+    #[test]
+    fn element_no_classes() {
+        let mut stream = InputStream::new();
+        stream.read_from_str("<div></div>", Some(Encoding::UTF8));
+
+        let mut parser = Html5Parser::new(&mut stream);
+        let (doc, _) = parser.parse();
+
+        assert_eq!(doc.get_root().classes.is_none(), true);
```
emwalker, line 3704: This and elsewhere would be
```rust
assert!(doc.get_root().classes.is_none());
```

I think you can get this automatically by running `cargo clippy --fix`.

src/html5_parser/parser/mod.rs
```diff
@@ -3679,4 +3692,68 @@ mod test {
...
+        if let Some(&ref classes) = div.classes.as_ref() {
+            assert_eq!(classes.len(), 3);
+
+            assert_eq!(classes.contains("one"), true);
+            assert_eq!(classes.contains("two"), true);
+            assert_eq!(classes.contains("three"), true);
+
+            assert_eq!(classes.is_active("one"), true);
+            assert_eq!(classes.is_active("two"), true);
+            assert_eq!(classes.is_active("three"), true);
```
emwalker, line 3728: See comment above.

emwalker (approved):
LGTM with suggestions

src/html5_parser/parser/mod.rs
Kiyoshika: I'll add the suggestions this and others, although `cargo clippy --fix` doesn't complain about anything for me (cargo 1.72.1, MacOS 13.5)

Kiyoshika: Oh, now I understand how it works, it just made the changes for me

Kiyoshika:
@emwalker I ran clippy fix which also adjusted some assertions outside my original changes but thought it's probably best to stay consistent. I'll get your final sign off before merging

emwalker (approved):

emwalker:
@Kiyoshika I'm thinking you have the ability to merge when you're ready.  Let me know if you don't.

emwalker:
Looks like our PRs overlapped, so there's a conflict you'll have to fix.

Issue #88, opened 2023-10-03
Update node_id to use pub(crate)


jaytaph:
See conversation:

https://github.com/jaytaph/gosub-browser/pull/80#pullrequestreview-1655123562

![image](https://github.com/jaytaph/gosub-browser/assets/241458/48aad7ee-a56f-4f13-b20c-e122fe562813)

manish7017:
Maybe we should consider updating the visibility of all NodeID instances to pub(crate). This is because the codebase currently relies on NodeID being pub, and changing it to pub(crate) would break this dependency. However, before we proceed with this change, we need to ensure that it won’t introduce any unforeseen issues. Let’s discuss this further.

blaumeise20:
I think this got misunderstood, I was talking about the field inside which has been made public, not the entire struct.
```rs
pub struct NodeId(pub(crate) usize);
```

PR #95, opened 2023-10-04
Update node_id to use pub(crate)

manish7017:
Issue #88


jaytaph:
@manish7017 Thanks for the PR.. would you mind signing the commits so we can merge it? See [Learn more about signing commits.](https://docs.github.com/articles/about-gpg/) for more information

jaytaph:
@blaumeise20 @emwalker Seems that using a pub(crate) doesn't work. Any idea's on how to fix this, or should we leave it as pub for now?

manish7017:
I will attempt to find the appropriate solution for pub(crate). In the meantime, let’s keep it as pub.

blaumeise20:
@jaytaph pub(crate) only makes the field visible to the current crate, so if you try to use it in a different crate, it won't work. It only works if you have the tests directly in the same crate (preferably even the same file).

blaumeise20:
Oh I see now that I looked at the code, I didn't mean the entire *struct* but the *field* of the struct.

jaytaph:
This seems to work! 👍 If you could sign the commits, i'm happy to merge!

emwalker:
I suspect we can use `let index: usize = node_id.into()` (or the equivalent) where we want access to the inner field.  But using `pub(crate)` is totally fine at this stage.  This PR is an improvement.

manish7017:
Hello, can someone guide me on how to sign the remaining commits? This is my first time signing a commit.
I’ve been trying to sign the commit for the past 2-3 hours, but there’s still one left. Can someone guide me on this?

Kiyoshika:
> Hello, can someone guide me on how to sign the remaining commits? This is my first time signing a commit.
> I’ve been trying to sign the commit for the past 2-3 hours, but there’s still one left. Can someone guide me on this?

Are you on MacOS? I just had to do this recently

jaytaph:
There is a lot going on in this PR :-)

My suggestion would be not to try to fix it, but close this PR, and submit another one, since the change is small. The new commits seem to be signed, so that should not be an issue.

jaytaph:
@manish7017 see previous comment

emwalker:
To get it to work, you'll need to add a signature to GitHub and then either rebase your earlier commits or start a new PR.  (I'm away from my home laptop so I can't look at what might be going on with this PR until later today.)

emwalker:
I agree with @jaytaph.  Probably best to just close this PR and create a new one.

PR #97, opened 2023-10-05
Enable tests/data/html5lib-tests/tokenizer/test3.test and other changes

emwalker:
This PR
* Enables another tokenization test file from the vendored test suite
* Replaces the conversion of UTF-16 characters using `unsafe` with one that doesn't use `unsafe`
* Runs `cargo clippy --fix`

### DIFF
--- a/src/html5_parser/parser/mod.rs
+++ b/src/html5_parser/parser/mod.rs
@@ -3458,10 +3458,10 @@ mod test {
         node_create!(parser, "div");
         node_create!(parser, "p");
         node_create!(parser, "button");
-        assert_eq!(parser.is_in_scope("p", Scope::Regular), true);
-        assert_eq!(parser.is_in_scope("p", Scope::Button), false);
-        assert_eq!(parser.is_in_scope("p", Scope::ListItem), true);
-        assert_eq!(parser.is_in_scope("p", Scope::Select), false);
+        assert!(parser.is_in_scope("p", Scope::Regular));
+        assert!(!parser.is_in_scope("p", Scope::Button));
+        assert!(parser.is_in_scope("p", Scope::ListItem));
+        assert!(!parser.is_in_scope("p", Scope::Select));
     }
 
     #[test]
@@ -3470,10 +3470,10 @@ mod test {
         let mut parser = Html5Parser::new(&mut stream);
 
         parser.open_elements.clear();
-        assert_eq!(parser.is_in_scope("p", Scope::Regular), false);
-        assert_eq!(parser.is_in_scope("p", Scope::Button), false);
-        assert_eq!(parser.is_in_scope("p", Scope::ListItem), false);
-        assert_eq!(parser.is_in_scope("p", Scope::Select), false);
+        assert!(!parser.is_in_scope("p", Scope::Regular));
+        assert!(!parser.is_in_scope("p", Scope::Button));
+        assert!(!parser.is_in_scope("p", Scope::ListItem));
+        assert!(!parser.is_in_scope("p", Scope::Select));
     }
 
     #[test]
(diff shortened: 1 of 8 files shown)


src/html5_parser/parser/mod.rs
```diff
@@ -3458,10 +3458,10 @@ mod test {
...
         node_create!(parser, "p");
         node_create!(parser, "button");
-        assert_eq!(parser.is_in_scope("p", Scope::Regular), true);
-        assert_eq!(parser.is_in_scope("p", Scope::Button), false);
-        assert_eq!(parser.is_in_scope("p", Scope::ListItem), true);
-        assert_eq!(parser.is_in_scope("p", Scope::Select), false);
+        assert!(parser.is_in_scope("p", Scope::Regular));
+        assert!(!parser.is_in_scope("p", Scope::Button));
+        assert!(parser.is_in_scope("p", Scope::ListItem));
+        assert!(!parser.is_in_scope("p", Scope::Select));
```
emwalker, line 3464: These changes are from `cargo clippy --fix`.  I've added a `make fix` task to help with this.

src/testing/tokenizer.rs
```diff
@@ -107,17 +110,261 @@ impl Test {
...
+        let n = u16::from_str_radix(&cap[1], 16).unwrap();
+        // There are UTF-16 characters that the following will not decode into UTF-8, so we might
+        // be dropping characters when a DecodeUtf16Error error is encountered.
+        std::char::decode_utf16([n])
+            .filter_map(|r| r.ok())
+            .collect::<String>()
     })
-    .into_owned()
+    .to_string()
 }
```
emwalker, line 368: Here we avoid using `unsafe` and rely on `std::char::decode_utf16` instead, possibly dropping `char`s if they don't convert to UTF-8.  See also the `TODO` comment at the top of the function.

src/testing/tokenizer.rs
```diff
@@ -107,17 +110,261 @@ impl Test {
...
+        // Try and find an error that matches the code, but has a different line/pos. Even though
+        // it's not always correct, it might be a off-by-one position.
+        for actual in tokenizer.get_error_logger().get_errors() {
+            if actual.message == expected.code
+                && (actual.line as i64 != expected.line || actual.col as i64 != expected.col)
+            {
+                panic!(
+                    "[{}]: wanted {:?}, got {:?}",
+                    self.description, expected, actual
+                );
```
emwalker, line 172: It was convenient to move these `assert_` methods to `impl Test` in order to make it easier to print out additional context (`self.description`, in this case).

tests/tokenizer.rs
```diff
@@ -233,14 +38,28 @@ fn assert_doctype(
 #[test_case("pendingSpecChanges.test")]
 #[test_case("test1.test")]
 #[test_case("test2.test")]
-// #[test_case("test3.test")]
+#[test_case("test3.test")]
```
emwalker, line 41: This is the heart of this PR.

src/testing/tokenizer.rs
```diff
@@ -107,17 +110,261 @@ impl Test {
...
+            sys_identifier
+        );
+    }
 }
 
-pub fn escape(input: &str) -> String {
-    let re = Regex::new(r"\\u([0-9a-fA-F]{4})").unwrap();
-    re.replace_all(input, |caps: &regex::Captures| {
-        let hex_val = u32::from_str_radix(&caps[1], 16).unwrap();
+pub fn from_utf16_lossy(input: &str) -> String {
```
Kiyoshika, line 351: just for my understanding, what does this function do? Is this attempting to convert UTF16 hex codes into UTF8 (`String`) and ignoring any that aren't able to be converted? I'm not really familiar with character encodings tbh

Kiyoshika: oh, or is this replacing the "escaped" characters with the actual UTF16 characters? I may have misread at first, didn't notice the double slash

jaytaph (approved):

src/testing/tokenizer.rs
emwalker: > just for my understanding, what does this function do? Is this attempting to convert UTF16 hex codes into UTF8 (String) and ignoring any that aren't able to be converted? 

Important caveat — I know very little about UTF-16 and UTF-8.

Some of our third-party test cases appear to have UTF-16 escape sequences in them (e.g., `\uDC00`, a hex value), and the test data specifies line and column numbers that you should be at in the input stream after attempting to tokenize these characters in order to pass the test.  Rust runs into a challenge here because its internal strings are UTF-8, and support for UTF-16 is pretty minimal.  So we end up failing a lot of tests because we're not getting the results the third-party test suite expects.

This function makes a best effort attempt to convert some UTF-16 escape sequences into UTF-8 characters.  But some escape sequences are special and can't be represented in UTF-8.  In these cases, we just drop those characters.  (But that doesn't help us pass some of the tests in the the third-party test suite, so we disable them.)

emwalker:
I'll have to troubleshoot commit signing in the new repo.  I must not have it set up globally.

src/testing/tokenizer.rs
Kiyoshika: I see, thanks for the explanation

PR #102, opened 2023-10-05
Replace older mod.rs style with newer <module>.rs style

Sharktheone:
I think, it is better to use the newer style, since it is mostly pretty annoying when having many `mod.rs` files open in the editor.
Also the older style makes it sometimes difficult, when searching a module file. At the current state of the project, this might not be such an issue, but when the project grows, it will be better to use the newer style.

### DIFF
--- a/src/html5_parser/node.rs
+++ b/src/html5_parser/node.rs
@@ -89,6 +89,8 @@ impl NodeId {
 pub struct Node {
     /// ID of the node, 0 is always the root / document node
     pub id: NodeId,
+    /// Named ID of the node, from the "id" attribute on an HTML element
+    pub named_id: Option<String>,
     /// parent of the node, if any
     pub parent: Option<NodeId>,
     /// children of the node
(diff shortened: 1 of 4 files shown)


emwalker:
@Sharktheone I have not heard of one of the styles superseding the other.  The way I've seen the two styles used depends upon how complex the module is.  Do you have a link that further discusses what you have in mind?

Sharktheone:
It is mentioned in the [Rust-book](https://doc.rust-lang.org/book/ch07-05-separating-modules-into-different-files.html#alternate-file-paths).

emwalker:
I'll take a look and refresh my memory. It's been a little while since I read that.

emwalker:
@Sharktheone it seems you are correct.  I vaguely remember reading that section, but it seems I forgot it.  Most of us, including myself, are at a fairly early stage in learning Rust.

Can you bring your PR up to date with latest `main`?

Sharktheone:
Github does't let me update the branch, i reopen this

Sharktheone:
I reopened this as #107 I didn't thinked about this, that i shouldn't push something like this to main 🤦

PR #103, opened 2023-10-05
Add local file support

### LINKED ISSUE #81: Add ability to pass a local file into `gosub-browser` executable
Right now we can pass a URL, but I think it'll be useful for testing if we can do something like:
`gosub-browser file:///path/to/myfile.html`

We already have an ability to read a file with `InputStream` so I think this should be relatively easy.

psuedomoth:
Closes #81

### DIFF
--- a/src/bin/gosub-browser.rs
+++ b/src/bin/gosub-browser.rs
@@ -1,4 +1,5 @@
 use std::process::exit;
+use std::fs;
 
 use gosub_engine::html5_parser::input_stream::Confidence;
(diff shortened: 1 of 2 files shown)


Kiyoshika:
Thanks for the PR!

Sorry to get to this late. I tested it out, seems to work fine, although can you fix the CI errors? If you run `make test` locally it will point out some formatting issues. Looks to be some trailing whitespace at the end of a couple lines

Also, for now I think we want to rename the executable from `gosub-browser` to `gosub-parse` in this early stage since we're nowhere near a functioning browser yet.

PR #104, opened 2023-10-06
Add own error type

emwalker:
This PR
* Adds our own error type using `thiserror`
* Handles top-level errors with `anyhow`
* Replaces `reqwest` with `ureq`

The main issue with `reqwest` is that it is primarily an async library and brings in the `tokio` runtime as a dependency.  Since we're only doing simple synchronous execution at this point, there's no need for all of that.  In these cases, `ureq` is often the way to go.  If we decide to go with `tokio` later on, it should be easy enough to bring back in `reqwest` if it makes sense.


src/html5_parser/parser/adoption_agency.rs
```diff
@@ -340,43 +341,43 @@ mod test {
...
             parser
-                .find_furthest_block_idx(NodeId(1))
+                .find_furthest_block_idx(1)
                 .expect("furthest element not found"),
             1
         );
         assert_eq!(
             parser
-                .find_furthest_block_idx(NodeId(2))
+                .find_furthest_block_idx(2)
```
emwalker, line 357: We're still working with `NodeId`, but our interface is now a little more ergonomic.

src/html5_parser/parser/mod.rs
```diff
@@ -2254,7 +2257,7 @@ impl<'a> Html5Parser<'a> {
                         // ignore token
                         return;
                     }
-                    let node_id = node_id.unwrap();
+                    let node_id = node_id.expect("node_id");
```
emwalker, line 2260: Convention: use `expect(...)` when we know from earlier code that a panic shouldn't happen.

src/html5_parser/tokenizer/mod.rs
```diff
@@ -2271,16 +2272,22 @@ impl<'a> Tokenizer<'a> {
...
                 *name = new_name;
             }
             Token::EndTagToken { name, .. } => {
                 *name = new_name;
             }
-            _ => panic!("trying to set the name of a non start/end tag token"),
+            _ => {
+                return Err(Error::Parsing(
+                    "trying to set the name of a non start/end tag token".into(),
+                ))
```
emwalker, line 2286: This is just an example of the kind of error handling we can introduce as an alternative to panicking.

src/types.rs
```diff
@@ -0,0 +1,15 @@
...
+    Request(#[from] Box<ureq::Error>),
+
+    #[error("io error: {0}")]
+    IO(#[from] std::io::Error),
+
+    #[error("parsing error: {0}")]
+    Parsing(String),
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
```
emwalker, line 15: Our new `Error` and `Result` types.

Cargo.toml
```diff
@@ -27,7 +27,9 @@ serde_derive = "1.0"
 regex = "1"
 lazy_static = "1.4"
 typed-arena = "2.0.2"
-reqwest = { version = "0.11.12", features = ["blocking"] }
+thiserror = "1.0.49"
+ureq = "2.8.0"
+anyhow = "1.0.75"
```
emwalker, line 32: The `thiserror` and `anyhow` crates are by the same author, who is also the author of the `serde` crate.  The `anyhow` crate is for application-level error handling, and the `thiserror` crate is for library-level error handling.

src/html5_parser/node.rs
```diff
@@ -38,7 +38,7 @@ pub enum NodeData {
 
 /// Id used to identify a node
 #[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq, Display)]
-pub struct NodeId(pub(crate) usize);
+pub struct NodeId(usize);
```
emwalker, line 41: I took a look at this, and it wasn't hard to make private again.  I did this just to illustrate how this can be done, but it's not a big deal for it to be crate-public.

src/html5_parser/parser/adoption_agency.rs
```diff
@@ -340,43 +340,43 @@ mod test {
 
         parser
             .active_formatting_elements
-            .push(ActiveElement::Node(NodeId(3)));
+            .push(ActiveElement::Node(3.into()));
```
emwalker, line 343: You can also use `NodeId::from(3)`:
```rust
.push(ActiveElement::Node(NodeId::from(3)));
```

emwalker: With these changes, it is possible to remove `pub(crate)` from the inner field of `NodeId`:
```rust
pub struct NodeId(usize);
```

But I didn't do this because it's not urgent, and because I didn't want to undo https://github.com/gosub-browser/gosub-engine/pull/100.

Kiyoshika: What are the differences between `NodeId(X)`, `X.into()` and `NodeId::from(X)`? Are these mainly readability or is there overhead with the current `NodeId(X)`?

emwalker: `NodeId(X)` can only be done if you're in the module, or the inner field is `pub` or `pub(crate)`.  `X.into()` and `NodeId::from(X)` are equivalent, and both become available if you implement [From<usize>](https://github.com/gosub-browser/gosub-engine/blob/6fb63ba7083b0280224e84b33b7560ca7af260a4/src/html5_parser/node.rs#L49-L53) in this case.

Kiyoshika: oh ok, got it

emwalker: I believe the Rust compiler is smart enough to know that you're using `NodeId` as a `usize` and will optimize it away.  The [newtype pattern](https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html) is pretty common in Rust and is used to take ownership of external types so that you can use Rust's type system to ensure that you're not passing something that shouldn't be a node id into a function argument.

Kiyoshika:
Nice, I haven't really checked but do we have a bunch of `panic`s lying around in the code base? Checking if we'll need to do some cleanup work with the new error types post implementation

Kiyoshika (approved):

emwalker:
> Nice, I haven't really checked but do we have a bunch of panics lying around in the code base?

Yes, quite a few panics and unwraps.  Our browser would crash if we hit one of these cases.  Some of them are benign because you can see from the code that a panic will never happen.  Others we should deal with.

src/html5_parser/tokenizer/mod.rs
```diff
@@ -2271,16 +2272,22 @@ impl<'a> Tokenizer<'a> {
     }
 
     // Sets the given name into the current token
-    fn set_name_in_current_token(&mut self, new_name: String) {
-        match &mut self.current_token.as_mut().unwrap() {
+    fn set_name_in_current_token(&mut self, new_name: String) -> Result<()> {
+        match &mut self.current_token.as_mut().expect("current token") {
```
emwalker, line 2276: I don't think this should be an `expect`, now that I think about it.  I wasn't sure what to do in this case.

Kiyoshika:
> > Nice, I haven't really checked but do we have a bunch of panics lying around in the code base?
> 
> Yes, quite a few panics and unwraps. Our browser would crash if we hit one of these cases. Some of them are benign because you can see from the code that a panic will never happen. Others we should deal with.

Gotcha, I kind of figured. It may have been useful for debugging at first but we'll definitely need to clean those up and my guess would be any errors may be sent to the console API that Joshua is working on (in the end state) to actually record that error somewhere or however we decide to handle it.

Nice work!

jaytaph:
I figure that http requests will be handled by a separate network layer/component instead of using ureq/reqwest. For now we only use it in some tooling so im fine whichever we use

PR #105, opened 2023-10-06
Named Node IDs and basic DOM querying

Kiyoshika:
This PR introduces "named" IDs and basic DOM querying.

Currently we have the numerical ID system but this aims to address the actual `id` attribute of an element. Honestly I wasn't sure what to call them to differentiate from the current `node.id` we have, so `named_id` was my best idea I came up with - let me know if there's a better name.

This introduces a new struct member within the `Document` that keeps track of which nodes have a named ID; this allows us to fetch a node via named ID easily instead of trying to iterate through the node arena. This is mainly for the convenience of performing something like `document.getElementById("myid")`

I've implemented some basic validation rules according to the HTML5 spec, so any spaces, IDs containing no characters or empty IDs are ignored.

One special caveat that we might have to revisit later, the HTML spec states that an ID (named ID in this case) should only correspond to ONE element within the DOM. The way I've implemented this currently, if you try to create two nodes with the exact same id, the second node is completely unsearchable (via `get_node_by_named_id`) within the DOM (but the `Node` itself still has the `id` attribute). This is *technically* not how most browsers behave so I'm not sure if we want to follow the spec very strictly like this, or bend the rules a bit.

For a more concrete example of what I mean, consider this:

```html
<!DOCTYPE html>
<html>
<p id="a">line one</p>
<p id="a">line two</p>
<script>
    document.getElementById("a").innerHTML = "modified one";
    document.getElementById("a").id = "b";
    document.getElementById("a").innerHTML = "modified two";
</script>
</html>
```
These two nodes have the exact same id. This is technically illegal according to the spec but this still works and prints "modified one" and "modified two" on separate lines (at least, in Chrome and Safari).

The way I've implemented this, this would not behave the same. After the ID is changed to "b", the original "a" is lost and completely inaccessible.

If we want to go that route, than instead of storing a single `NodeId` as the value, we could store a vector of `NodeId`s and if one changes (like above example) we could pop that one off. But for now, I've implemented strict rules following the spec. Curious to know your guys' thoughts on this


src/html5_parser/node.rs
```diff
@@ -212,6 +219,38 @@ impl Node {
...
+        self.named_id.is_some()
+    }
+
+    /// Set named ID (only applies to Element type, does nothing otherwise)
+    pub fn set_named_id(&mut self, named_id: &str) {
+        if self.type_of() == NodeType::Element {
+            self.named_id = Some(named_id.to_owned());
+            // TODO: log a warning/error if this fails for some reason
+            let _ = self.insert_attribute("id", named_id);
+        }
```
emwalker, line 237: Looking at the implementation of `self.insert_attribute`, I see that the function could be infallible (never return an error) if we never failed the [type check at the top of the function](https://github.com/gosub-browser/gosub-engine/blob/6fb63ba7083b0280224e84b33b7560ca7af260a4/src/html5_parser/node.rs#L273-L275).  Then I realized that we're doing that type check in a number of functions.  That tells me we're using a single struct where we should be using several.  One struct doesn't have these functions defined, and another has them defined but no `self.type_of()` test.

emwalker: I'm not exactly sure how to split these structs apart, yet, though, so I don't have a concrete suggestion at this point.

Kiyoshika: Yes, I had the same thought, but wasn't sure at the moment how we wanted to split them apart. The multiple type checking is a bit annoying but I would like to break those apart at some point

src/html5_parser/parser/document.rs
```diff
@@ -32,23 +37,97 @@ impl Document {
...
     }
 
+    /// Fetches a node by named id (string) or returns None when no node with this ID is found
+    pub fn get_node_by_named_id(&self, named_id: &str) -> Option<&Node> {
+        let node_id = self.named_id_elements.get(named_id)?;
+        self.arena.get_node(*node_id)
+    }
+
+    /// Fetches a mutable node by named id (string) or returns None when no node with this ID is found
+    pub fn get_mut_node_by_named_id(&mut self, named_id: &str) -> Option<&mut Node> {
```
emwalker, line 63: How about `_mut` as the final suffix instead of being in the middle of the function name, here and elsewhere?

emwalker (approved):
I don't want to block this PR, but I think a future change should be to figure out how to get rid of these tests, presumably by delegating to another struct that doesn't need the test because the test is always true:

```rust
            if node.type_of() != NodeType::Element {
                return;
            }
```

Kiyoshika:
> I don't want to block this PR, but I think a future change should be to figure out how to get rid of these tests, presumably by delegating to another struct that doesn't need the test because the test is always true:
> 
> ```rust
>             if node.type_of() != NodeType::Element {
>                 return;
>             }
> ```

Agreed, I can open a discussion and brainstorm some ideas on how to restructure the current `Node` struct to be more versatile with the different types we have. Especially as the engine starts getting developed more and more, we will inevitably introduce some jank like this unless we find a smarter way to handle it

src/html5_parser/parser/document.rs
Kiyoshika: yes, i'll adjust these

emwalker:
> Curious to know your guys' thoughts on this

I'd just keep an eye on it and not paint ourselves into a corner if we figure out that the browsers are doing this for a good reason.

jaytaph (approved):

jaytaph:
Would it make sense from a web developer point of view to have besides the standard developer tools, also a parser details where issues like these will be logged. Also incorrect nested tags etc should be found here. I think it would be a good first step in getting awareness on correct html5 usage (im not sure if this is already available in dev tools though)

Kiyoshika:
I see there are conflicts with the new error types introduces, I will resolve these when I get home and merge

PR #115, opened 2023-10-07
Split `Node.data` into structs

Kiyoshika:
Following #108, this PR creates dedicated structs to hold the node data. This is to remove some bloat from the `Node` struct itself as well as simplify the methods by avoiding type checks at the top.

Currently these new structs are basic encapsulation over a standard type (such as `String` or `HashMap<String, String>`). I was torn between whether or not to keep this very thin wrapper over the underlying private type or to expose the underlying type as public. The benefit of keeping it private is this allows us to define a very explicit interface of how this data should be used, the downside is "duplicating" method calls since most of these are 1:1 with the underlying type's methods.

When it comes to usage, you can either use `if let` if you want to verify a type before performing an operation:

```rust
let node = document.get_node_by_id(NodeId::from(4));

// ignored if node is not this type
if let NodeData::Text(TextData { value, .. }) = &node.data {
    // do something with "value"
}
```

Or you can use `match` if you want to do some type of error handling:
```rust
let node = document.get_node_by_id(NodeId::from(4));

match &node.data {
    NodeData::Text(TextData { value, .. }) => {
        // do something with "value"
    }
    _ => /* oh no, this is not a text node, do something! */
}
```

I've also ran `parser_test` to make sure this somehow didn't break the parser and verified we are still getting 76 passing tests (the same value as Joshua posted in Slack earlier)

NOTE: we may end up moving both the  `named_id` and `class` members & methods into the new `ElementData` struct if these are not applicable to other nodes, but I let them be for now.

EDIT: forgot to mention, this also removed a few duplicate tests that I felt were redundant. I also removed the `non_element` tests in `Node` since those checks are no longer required with this new API

### DIFF
--- a/src/html5_parser/node.rs
+++ b/src/html5_parser/node.rs
@@ -146,7 +146,7 @@ impl Node {
             named_id: None,
             parent: None,
             children: vec![],
-            data: NodeData::Element(ElementData::new_with_name_and_attributes(name, attributes)),
+            data: NodeData::Element(ElementData::with_name_and_attributes(name, attributes)),
             name: name.to_string(),
             namespace: Some(namespace.into()),
             classes: Some(ElementClass::new()),
@@ -160,7 +160,7 @@ impl Node {
             named_id: None,
             parent: None,
             children: vec![],
-            data: NodeData::Comment(CommentData::new_with_value(value)),
+            data: NodeData::Comment(CommentData::with_value(value)),
             name: "".to_string(),
             namespace: None,
             classes: None,
@@ -174,7 +174,7 @@ impl Node {
             named_id: None,
             parent: None,
             children: vec![],
-            data: NodeData::Text(TextData::new_with_value(value)),
+            data: NodeData::Text(TextData::with_value(value)),
             name: "".to_string(),
             namespace: None,
             classes: None,
@@ -381,14 +381,15 @@ mod tests {
         assert!(node.children.is_empty());
         assert_eq!(node.name, "div".to_string());
         assert_eq!(node.namespace, Some(HTML_NAMESPACE.into()));
-        match &node.data {
-            NodeData::Element(element) => {
-                assert_eq!(element.get_name(), "div");
-                assert!(element.attributes.contains("id"));
-                assert_eq!(element.attributes.get("id").unwrap(), "test");
-            }
-            _ => panic!(),
-        }
+        let NodeData::Element(ElementData {
+            name, attributes, ..
+        }) = &node.data
+        else {
+            panic!()
+        };
+        assert_eq!(name, "div");
+        assert!(attributes.contains("id"));
+        assert_eq!(attributes.get("id").unwrap(), "test");
     }
 
     #[test]
@@ -399,12 +400,10 @@ mod tests {
         assert!(node.children.is_empty());
         assert_eq!(node.name, "".to_string());
         assert_eq!(node.namespace, None);
-        match &node.data {
-            NodeData::Comment(comment) => {
-                assert_eq!(comment.get_value(), "test");
-            }
-            _ => panic!(),
-        }
+        let NodeData::Comment(CommentData { value, .. }) = &node.data else {
+            panic!()
+        };
+        assert_eq!(value, "test");
     }
 
     #[test]
@@ -415,12 +414,10 @@ mod tests {
         assert!(node.children.is_empty());
         assert_eq!(node.name, "".to_string());
         assert_eq!(node.namespace, None);
-        match &node.data {
-            NodeData::Text(text) => {
-                assert_eq!(text.get_value(), "test");
-            }
-            _ => panic!(),
-        }
+        let NodeData::Text(TextData { value }) = &node.data else {
+            panic!()
+        };
+        assert_eq!(value, "test");
     }
 
     #[test]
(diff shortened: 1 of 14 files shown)


src/html5_parser/node_data/comment_data.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub fn new_with_value(value: &str) -> Self {
+        let mut comment_data = CommentData::new();
+        comment_data.set_value(value);
+
+        comment_data
+    }
+
+    pub fn get_value(&self) -> String {
+        self.value.clone()
+    }
```
emwalker, line 28:
```suggestion
    pub fn value(&self) -> &str {
        &self.value
    }
```

Rust knows the difference between when a field on a struct is being accessed and when a function with the same name is being called. I have not seen readers with the `get_` prefix very often.

In this case, it would also be nice to let the caller opt into cloning the string instead of preemptively doing it.

src/html5_parser/node_data/element_data.rs
```diff
@@ -0,0 +1,104 @@
...
+        self.attributes.is_empty()
+    }
+
+    pub fn iter(&self) -> Iter<'_, String, String> {
+        self.attributes.iter()
+    }
+
+    pub fn clone_attributes(&self) -> HashMap<String, String> {
+        self.attributes.clone()
+    }
```
emwalker, line 60: An alternative might be to return a reference and allow the caller to opt into cloning:

```suggestion
    pub fn attributes(&self) -> &HashMap<String, String> {
        &self.attributes
    }
```

I noticed a lot of the methods are `pub`.  We might want to default to private or crate-public as much as possible, so that we keep the exposed API surface minimal for later on.  Doing this has the advantage of letting `rust-analyzer` tell you when a function isn't being used.  (It doesn't do this when the function is marked `pub`.)

emwalker (commented):
I left some suggestions; feel free to ignore them.

My questions as I read through this PR:
* How about dropping the `get_` prefix for the readers?  Do we readers at this point?
* Can we drop the setters at this point and backfill them when we have some real complexity?
* How about using `Self` in the constructors?
* How about having callers opt into cloning instead of doing this preemptively or exposing `clone_` functions?  E.g., returning `&str` instead of `String`, and returning `&Hash<String, String>`, etc.
* Would it make sense to use private or `pub(crate)` functions and structs where possible in favor of `pub`?  In some cases this won't be possible.

In internal crate code, I doubt we need to use the readers, and instead you can pattern match on the fields:

```rust
            NodeData::Element(ElementData { name, attributes, .. }) => {
                _ = write!(f, "{}<{}", buffer, name);
                for (key, value) in attributes.iter() {
                    _ = write!(f, " {}={}", key, value);
                }
                _ = writeln!(f, ">");
```

This doesn't violate encapsulation because we're still in internal code.  The main motivation for making the internal field of `NodeId(usize)` private was so that people don't revert to using `usize` as an index into maps and arrays, which seems like a likely thing to happen.

Just some thoughts.  As I said, feel free to ignore.

src/html5_parser/node.rs
```diff
@@ -1,4 +1,8 @@
 use crate::html5_parser::element_class::ElementClass;
+use crate::html5_parser::node_data::comment_data::CommentData;
+use crate::html5_parser::node_data::document_data::DocumentData;
+use crate::html5_parser::node_data::element_data::ElementData;
+use crate::html5_parser::node_data::text_data::TextData;
```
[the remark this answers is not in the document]
emwalker: Not so sure about this suggestion.  I'm seeing different approaches to this kind of thing.  I thought the suggestion was idiomatic, but it looks like it [might not necessarily be](https://github.com/servo/servo/blob/c4ba2970b543ef268dde2461dc662c56c82edaef/components/script/dom/servoparser/html.rs#L18-L31).

emwalker:
> How about using Self in the constructors?

I also thought this was idiomatic, but I'm now seeing [different approaches](https://github.com/servo/servo/blob/c4ba2970b543ef268dde2461dc662c56c82edaef/components/script/dom/servoparser/html.rs#L78).

emwalker:
> How about dropping the get_ prefix for the readers? Do we need readers at this point?

I thought this was idiomatic, but I'm now seeing [different approaches](https://github.com/servo/servo/blob/c4ba2970b543ef268dde2461dc662c56c82edaef/components/bluetooth/lib.rs#L369).

emwalker:
> Can we drop the setters at this point and backfill them when we have some real complexity?

I thought this was idiomatic, but I'm now seeing [different approaches](https://github.com/servo/servo/blob/c4ba2970b543ef268dde2461dc662c56c82edaef/components/script/dom/videotrack.rs#L75-L77).

Probably a lot of these conventions are a function of house style.  I'm not sure that Servo is necessarily something to emulate, although it's definitely providing more data points.

Kiyoshika:
> I left some suggestions; feel free to ignore them.
> 
> My questions as I read through this PR:
> 
> * How about dropping the `get_` prefix for the readers?  Do we need readers at this point?
> * Can we drop the setters at this point and backfill them when we have some real complexity?
> * How about using `Self` in the constructors?
> * How about having callers opt into cloning instead of doing this preemptively or exposing `clone_` functions?  E.g., returning `&str` instead of `String`, and returning `&Hash<String, String>`, etc.
> * Would it make sense to use private or `pub(crate)` functions and structs where possible in favor of `pub`?  In some cases this won't be possible.
> 
> In internal crate code, I doubt we need to use the readers, and instead you can pattern match on the fields:
> 
> ```rust
>             NodeData::Element(ElementData { name, attributes, .. }) => {
>                 _ = write!(f, "{}<{}", buffer, name);
>                 for (key, value) in attributes.iter() {
>                     _ = write!(f, " {}={}", key, value);
>                 }
>                 _ = writeln!(f, ">");
> ```
> 
> This doesn't violate encapsulation because we're still in internal code. The main motivation for making the internal field of `NodeId(usize)` private was so that people don't revert to using `usize` as an index into maps and arrays, which seems like a likely thing to happen.
> 
> Just some thoughts. As I said, feel free to ignore.

Thanks for the thorough review!

I did take a lot of your suggestions into account. I changed visibility of most things to crate visibility instead of full public and removed setters until/unless we need them later. Also removed the get prefix for the publicly-accessible fields (outside of the crate).

Also, I do like the pattern matching on fields better anyways so that was a good suggestion, I think it looks cleaner; thank you

Kiyoshika:
Ah, I did not see your other comments while I was working. I think we are fine with the new suggestions as of now; if we start needing more getter/setter style stuff in the future as things get more complicated then maybe we can revisit

emwalker:
@Kiyoshika thanks for being gracious.  I'm still learning what is and isn't idiomatic, apparently.  Walking through the Servo code today has been helpful.

emwalker (approved):

jaytaph:
Looks good to me. I'm not the person to ask for rust issues, so i'm relying on the suggestions of @emwalker.

PR #121, opened 2023-10-09
Node file structure

lucalewin:
PR for issue #120

While doing this I thought that maybe the files in the now `node/data` directory could be combined into the `node/data.rs` file since they are quite short. This could reduce overall complexity. Or are these files growing in size in the future?

PR #122, opened 2023-10-09
move classes member into ElementData

Kiyoshika:
Small cleanup PR. Previously we were holding the `classes` member inside `Node` as an optional and only used if it's an HTML element (since only an element can have classes.) With the introduction of `ElementData` it makes more sense to move this here. Also removed one test that is no longer applicable since it's no longer an `Option<>`.


emwalker (approved):

jaytaph (approved):

jaytaph:
@Kiyoshika small rebase is needed because of the node-class file changes of #121

Kiyoshika:
> @Kiyoshika small rebase is needed because of the node-class file changes of #121

yup, I knew that would happen haha. Will rebase and merge when I'm free

Kiyoshika:
Geez, this is the messiest rebase of all time. I really hate git sometimes

PR #126, opened 2023-10-10
Create doc comments

lucalewin:
There were still a few files where normal comments (`//`) instead of doc comments (`///`) were used. I replaced (hopefully all of) them with doc comments where ever it was suitable.
Otherwise nothing should have changed.


jaytaph:
@lucalewin Thanks for the PR! I've did some documentation cleanup at another PR as well, so that one will conflict with this one, but I'll merge yours first.

PR #127, opened 2023-10-10
Added more documentation

jaytaph:
Added additional documentation to structs and modules. Still a lot of room for improvement

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1,2 +1,8 @@
+//! This module contains all the APIs that are exposed through javascript
+//!
+//! A website can control many of the aspects of the useragent with javascript through different APIs.
+//! For instance, there is the console api, that allows sites to log information to the console. The
+//! window api, that allows sites to control the window, and the document api, that allows sites to
+//! make changes in the DOM.
 #[allow(dead_code)]
 pub mod console;
--- a/src/api/console.rs
+++ b/src/api/console.rs
@@ -1,3 +1,4 @@
+//! This module contains the console api as described by <https://console.spec.whatwg.org/>
 mod buffer;
 mod formatter;
 mod writable_printer;
--- a/src/html5_parser.rs
+++ b/src/html5_parser.rs
@@ -1,3 +1,7 @@
+//! This module contains the HTML5 parser as specified by the W3C.
+//!
+//! The parser's job is to take a stream of bytes and turn it into a DOM tree. The parser is
+//! implemented as a state machine and runs in the current thread.
 pub mod dom;
 pub mod element_class;
 pub mod error_logger;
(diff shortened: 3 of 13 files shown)


blaumeise20 (changes requested):
I have no idea if I'm "socially" allowed to review stuff in here, but I have one suggestion: don't use a "This blabla" style but a more present ... way (I forgot the name). So instead of writing
```
This module contains the HTML5 parser as specified by the W3C.
```
I would write
```
The HTML5 parser implemented according to the W3C specification.
```
Similar things apply for some other things like macros and structs. Essentially it should be consistent with the way methods/functions are documented, e.g.:
```
Returns the current offset in the stream
```

jaytaph:
> I have no idea if I'm "socially" allowed to review stuff in here

everyone is :-)

I think it is a good idea. If you willing to help us out on this, that would be even better :)

emwalker:
> I have no idea if I'm "socially" allowed to review stuff in here

That's how we all got here so far. :)

src/api.rs
```diff
@@ -1,2 +1,8 @@
+//! This module contains all the APIs that are exposed through javascript
```
emwalker, line 1:
```suggestion
//! All the APIs that are exposed through JavaScript
```

Just to run with @blaumeise20's thought.  He might have additional suggestions.

src/api/console.rs
```diff
@@ -1,3 +1,4 @@
+//! This module contains the console api as described by <https://console.spec.whatwg.org/>
```
emwalker, line 1:
```suggestion
//! The console API, as described by <https://console.spec.whatwg.org/>
```

src/html5_parser.rs
```diff
@@ -1,3 +1,7 @@
+//! This module contains the HTML5 parser as specified by the W3C.
```
emwalker, line 1:
```suggestion
//! The HTML5 parser, following the W3C specification
```

src/html5_parser/input_stream.rs
```diff
@@ -41,29 +46,36 @@ impl fmt::Display for Position {
...
     Utf8(char),
-    /// Surrogate character (since they cannot be stored in <char>)
+    /// Surrogate character (since they cannot be stored in char)
     Surrogate(u16),
     /// End of stream
     Eof,
 }
 
 impl Element {
+    /// Returns true when the element is an EOF
```
emwalker, line 63:
```suggestion
    /// Whether the element is an EOF
```

You get the idea.  Similar suggestions apply further down.

Kiyoshika (approved):

jaytaph:
Lets merge as-is, and change later when we have something more formalized

PR #130, opened 2023-10-11
replace macros with functions + remove unnecessary clones

lucalewin:
I completely missed the macros in the tokenizer. It's mostly the same as in the parser. I left the `to_lowercase` macro as it was because it does not use `self`. `char` has a method called `to_ascii_lowercase` which could be used but it does a check if the char is actually an uppercase ascii letter. The `to_lowercase` macro just assumes this so it should be minimally faster.

I also removed some unnecessary clones in the parser and tokenizer.

### DIFF
--- a/src/html5_parser/parser.rs
+++ b/src/html5_parser/parser.rs
@@ -1596,7 +1596,7 @@ impl<'a> Html5Parser<'a> {
             .name
             != name
         {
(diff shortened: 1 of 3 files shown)


emwalker (approved):

jaytaph (approved):

jaytaph:
@lucalewin thanks for the pr

PR #132, opened 2023-10-13
Fix func pop util

CharlesChen0823:
current pop_util implementation is not correctly through the html [docs](https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inhead)


jaytaph (approved):

jaytaph:
@CharlesChen0823  Thanks for the pr

PR #134, opened 2023-10-13
feat(css): add basic css parser (rdp)

neuodev:
This merge request implements basic CSS parser using [recursive descent parser](https://en.wikipedia.org/wiki/Recursive_descent_parser) algorithm and [Backus-Naur form](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) to defined the CSS grammar. 

Here is a simple example on how the parser generate the AST: 
```rs
        let mut parser = CSS3Parser::new();
        let style_sheet = parser.parse(
            r#"
            /* Shoud skip comments */

            .header {
                width: 10px;
                font-size: 1rem;
            } 

            #nav {
                height: 200px;
            }

            @media screen, printer, all and (max-width: 600px) {
                .header {
                    display: flex;
                }

                body {
                    max-height: 100vh;
                }
            }


            @media all {}
        "#,
        );

        assert_eq!(
            style_sheet,
            CSSStyleSheet {
                css_rules: vec![
                    CSSRule::CSSStyleRule(CSSStyleRule {
                        selector: CSSSelector::from_str_slice(vec![".", "header"]),
                        style_declarations: vec![
                            CSSStyleDeclaration {
                                name: "width".to_string(),
                                value: "10px".to_string()
                            },
                            CSSStyleDeclaration {
                                name: "font-size".to_string(),
                                value: "1rem".to_string()
                            }
                        ],
                    }),
                    CSSRule::CSSStyleRule(CSSStyleRule {
                        selector: CSSSelector::from_str_slice(vec!["#", "nav"]),
                        style_declarations: vec![CSSStyleDeclaration {
                            name: "height".to_string(),
                            value: "200px".to_string()
                        }],
                    }),
                    CSSRule::CSSMeidaRule(CSSMeidaRule {
                        conditions: vec![
                            MediaCondition::Type(MediaType::Screen),
                            MediaCondition::Type(MediaType::Printer),
                            MediaCondition::Type(MediaType::All),
                            MediaCondition::Feature(MediaFeature {
                                name: "max-width".to_string(),
                                value: Some("600px".to_string())
                            })
                        ],
                        css_rules: vec![
                            CSSRule::CSSStyleRule(CSSStyleRule {
                                selector: CSSSelector::from_str_slice(vec![".", "header"]),
                                style_declarations: vec![CSSStyleDeclaration {
                                    name: "display".to_string(),
                                    value: "flex".to_string()
                                }],
                            }),
                            CSSRule::CSSStyleRule(CSSStyleRule {
                                selector: CSSSelector::from_str_slice(vec!["body"]),
                                style_declarations: vec![CSSStyleDeclaration {
                                    name: "max-height".to_string(),
                                    value: "100vh".to_string()
                                }],
                            })
                        ],
                    }),
                    CSSRule::CSSMeidaRule(CSSMeidaRule {
                        conditions: vec![MediaCondition::Type(MediaType::All)],
                        css_rules: Vec::new(),
                    })
                ]
            }
        )
```

### DIFF
--- a/src/css3_parser/parser.rs
+++ b/src/css3_parser/parser.rs
@@ -8,6 +8,12 @@ pub struct CSSStyleSheet {
     css_rules: Vec<CSSRule>,
 }
 
+impl Default for CSSStyleSheet {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl CSSStyleSheet {
     pub fn new() -> CSSStyleSheet {
         CSSStyleSheet {
@@ -28,6 +34,12 @@ pub struct CSSStyleDeclaration {
 #[derive(Debug, PartialEq)]
 pub struct CSSSelector {
     path: Vec<String>,
+}
+
+impl Default for CSSSelector {
+    fn default() -> Self {
+        Self::new()
+    }
 }
 
 impl CSSSelector {
@@ -85,17 +97,26 @@ pub enum CSSRule {
     CSSMeidaRule(CSSMeidaRule),
 }
 
+/// # CSS3 Parser
+/// The parser using the Recursive Descent Parser algorithm (predictive parser).
+/// The grammer rules is defined using Backus–Naur form (BNF)
 #[derive(Debug, PartialEq)]
-struct CSS3Parser {
+pub struct CSS3Parser {
     tokenizer: CSSTokenizer,
     lookahead: Option<Token>,
     raw: String,
 }
 
+impl Default for CSS3Parser {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl CSS3Parser {
     pub fn new() -> CSS3Parser {
         CSS3Parser {
-            tokenizer: CSSTokenizer::new(),
+            tokenizer: CSSTokenizer::default(),
             lookahead: None,
             raw: "".to_string(),
         }
@@ -106,7 +127,7 @@ impl CSS3Parser {
         self.tokenizer.init(raw);
         self.lookahead = self.tokenizer.get_next_token();
 
-        return self.css_style_sheet();
+        self.css_style_sheet()
     }
 
     /// ```txt
@@ -134,7 +155,7 @@ impl CSS3Parser {
             rules.push(self.css_rule());
         }
 
-        return rules;
+        rules
     }
 
     /// ```txt
@@ -196,7 +217,7 @@ impl CSS3Parser {
             }
         }
 
-        return conditions;
+        conditions
     }
 
     /// ```txt
@@ -210,7 +231,7 @@ impl CSS3Parser {
             return MediaCondition::Feature(self.media_feature());
         };
 
-        return MediaCondition::Type(self.media_type());
+        MediaCondition::Type(self.media_type())
     }
 
     /// ```txt
@@ -295,12 +316,12 @@ impl CSS3Parser {
     fn css_style_declaration_list(&mut self) -> Vec<CSSStyleDeclaration> {
         let mut list = Vec::new();
 
-        // todo: add condtion
+        // todo: add condition
         while self.is_next_token(TokenType::Identifier) {
             list.push(self.css_style_declaration());
         }
 
-        return list;
+        list
     }
 
     /// ```txt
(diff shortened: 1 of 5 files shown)


jaytaph:
@neuodev could you try and fix the clippy issues?

jaytaph:
how does it deal with styling errors? (i don't know how normal css parsers are doing this)?

src/css3_parser/tokenizer.rs
```diff
@@ -0,0 +1,158 @@
...
+        self.cursor == self.raw.len()
+    }
+
+    pub fn get_next_token(&mut self) -> Option<Token> {
+        if !self.has_more_tokens() {
+            return None;
+        }
+
+        let raw = &self.raw[self.cursor..];
+        for (regex, token_type) in TOKENS_REGEXS.iter() {
```
emwalker, line 43: I'm guessing the use of regular expressions in the inner loop parser to identify the tokens will need to be provisional, for performance reasons.  But that can be profiled later and addressed if needed.

emwalker (approved):
I like it!  I like that we get an AST back that we can make assertions about in tests.

I would not be too attached to this code, as it is might change as we run into challenges and figure new things out.  But it looks like a promising start.

src/css3_parser/parser.rs
```diff
@@ -0,0 +1,500 @@
...
+    path: Vec<String>,
+}
+
+impl Default for CSSSelector {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl CSSSelector {
```
emwalker, line 45: I'm thinking it would eventually be good to drop the "CSS" prefix, and do something like
```rust
use crate::gosub::css3::{Selector, StyleRule};
// ...
```

But I wouldn't attempt that change in this PR, since it implies larger changes as well.

src/css3_parser/parser.rs
```diff
@@ -0,0 +1,500 @@
...
+                    max-height: 100vh;
+                }
+            }
+
+
+            @media all {}
+        "#,
+        );
+
+        assert_eq!(
```
emwalker, line 441: I love this test.  It would be nice if we had things up so that the testing of the HTML parsing was like this as well.

src/css3_parser/parser.rs
```diff
@@ -0,0 +1,500 @@
...
+        None
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn should_parse_css() {
```
emwalker, line 411:
```suggestion
    fn parses_css() {
```

Kiyoshika (approved):
I think this is an awesome start and will be iterated a lot in the future. We'll figure out how to deal with and how to handle the errors later so I'm not bothered about the panics yet.

I do agree with @emwalker that we should drop the `CSS` prefix in the struct names at a later point, but overall great!

src/css3_parser/parser.rs
```diff
@@ -0,0 +1,500 @@
...
+}
+
+#[derive(Debug, PartialEq)]
+pub struct CSSStyleRule {
+    selector: CSSSelector,
+    style_declarations: Vec<CSSStyleDeclaration>,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct CSSMeidaRule {
```
Kiyoshika, line 70: I just noticed this now, is this supposed to me `Meida` or `Media`?

jaytaph: Do we need to fix this before merge? Doesn't seem a blocker

jaytaph (approved):

src/css3_parser/parser.rs
emwalker: This is easy to fix later on.  But eventually we'll want to block on this kind of thing, once the rate of PRs slows down or there's more reviewers.

Kiyoshika: Agree, and since the CSS parser won’t even be used for a bit/in hot development, I don’t think it’s a big deal to fix right away, maybe just create an issue so we don’t forget

jaytaph:
Thanks @neuodev for the pr!

PR #135, opened 2023-10-14
Feat/nodes with document handle

jaytaph:
I think this current setup sounds like a reasonable "stable" state for now which fixes some problems concerning document handling through the parser.

It does not fix anything in the parser-tests, since these changes are needed for "template" elements which aren't tested yet.

From this point, we can fix the parser adjust_insert_location() function which IS needed to make the template elements (among others), so it will pass more test.

### DIFF
--- a/src/html5_parser/parser/adoption_agency.rs
+++ b/src/html5_parser/parser/adoption_agency.rs
@@ -179,7 +179,7 @@ impl<'stream> Html5Parser<'stream> {
                 };
 
                 let replacement_node =
-                    Node::new_element(node.name.as_str(), node_attributes, HTML_NAMESPACE);
+                    Node::new_element(&self.document, node.name.as_str(), node_attributes, HTML_NAMESPACE);
                 let replacement_node_id = self
                     .document
                     .borrow_mut()
@@ -222,7 +222,7 @@ impl<'stream> Html5Parser<'stream> {
                 NodeData::Element(ElementData {
                     name, attributes, ..
                 }) => {
-                    Node::new_element(name.as_str(), attributes.attributes.clone(), HTML_NAMESPACE)
+                    Node::new_element(&self.document, name.as_str(), attributes.attributes.clone(), HTML_NAMESPACE)
                 }
                 _ => panic!("formatting element is not an element"),
             };
@@ -312,7 +312,7 @@ mod test {
 
     macro_rules! node_create {
         ($self:expr, $name:expr) => {{
-            let node = Node::new_element($name, HashMap::new(), HTML_NAMESPACE);
+            let node = Node::new_element(&$self.document, $name, HashMap::new(), HTML_NAMESPACE);
             let node_id = $self.document.borrow_mut().add_node(node, NodeId::root());
             $self.open_elements.push(node_id);
         }};
--- a/src/html5_parser/parser/adoption_agency.rs
+++ b/src/html5_parser/parser/adoption_agency.rs
@@ -273,7 +273,8 @@ impl<'stream> Html5Parser<'stream> {
         };
 
         // Iterate
-        ((element_idx_oe + 1)..self.open_elements.len()).find(|&idx| open_elements_get!(self, idx).is_special())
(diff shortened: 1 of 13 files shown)


src/html5_parser/parser/adoption_agency.rs
```diff
@@ -12,13 +12,15 @@ pub enum AdoptionResult {
...
+impl<'stream> Html5Parser<'stream> {
+    /**
+     * When we talk about nodes, there are 3 contexts to consider:
+     *
+     * - The actual node data. This is called "node" in the code.
+     * - The node id. This is called "node_id" in the code.
+     * - The node index. This is called "node_idx" in the code. This is the index of the node in
+     *   either the open_elements or active_formatting_elements stack.
+     */
     pub fn run_adoption_agency(&mut self, token: &Token) -> AdoptionResult {
```
emwalker, line 24: We should probably do a second pass soon and reintroduce some of the good changes, like this one, that are being reverted here (not necessarily in this PR).

src/html5_parser/parser/document.rs
```diff
@@ -50,6 +85,29 @@ impl Document {
         }
     }
 
+    pub fn shared() -> DocumentHandle {
+        DocumentHandle(Rc::new(RefCell::new(Self::new())))
+    }
```
emwalker, line 90: The name `Document::shared()` is a little weird.  But the interface can be dialed in in a later PR.

emwalker (approved):
Most of the credit goes to @Kiyoshika for figuring out how to incorporate the ref-counted pointer into the existing code (most of this work is his).

I'm hoping the interface can be dialed in and revisited as we gain more experience using these smart pointers.  And it would be nice to reintroduce some of the changes that have been rolled back, such as the dropping of the macros and the touching up of the comments.  But this PR is a net improvement, and we can follow up with those things in later PRs.

emwalker:
@jaytaph I suggest configuring the merge button to only allow squash and merge, so that we can preserve a linear commit history of atomic commits on `main`.  That will make using `git bisect` much easier in the future.

jaytaph:
@emwalker my opinion about squash and merge is that it foes a rewrite which should be the author's responsibility. I've already rebased my tmp commits in this pr, and i think we could rebase again into 1 commit again. However, this is not always the case: there are many valid reasons why we want to have multiple commits inside a single pr (wspecially when it comes to bisecting), so i would suggest to do merges, but manually check if the commits actually make sense (this will work better at a later stage, as there are a lot of changes done presently.

jaytaph:
Looking at it from mobile, i would rebase it to 3 commits in this case

src/html5_parser/parser/adoption_agency.rs
jaytaph: Agreed. I think that some of the reverts are due to a lot of manual fixups during rebasing of this branch on main, as there were a lot of changes done in both branches. This should (hopefully) not happen very often as codebase matures

emwalker:
I think having authors rebase is sensible if we can maintain discipline.  As king commits are atomic and the tests pass, it will be straightforward to do a `git bisect` when problems come up.  If commits are merged that are just broken checkpoint commits, a `git bisect` will be laborious and challenging.

Some of this can be mitigated by keeping PRs small and focused (when possible).

jaytaph:
@Kiyoshika any input?  I wouldn't mind to change it to squash/merge, and i won't be standing in the way if we need to do this, but i'm not directly for it :-)

Kiyoshika:
> @Kiyoshika any input? I wouldn't mind to change it to squash/merge, and i won't be standing in the way if we need to do this, but i'm not directly for it :-)

I don't have a strong opinion on squash and merge, this is actually my first ever open source project I've contributed to so I have only dealt with my own simple commits in my own projects; I'm not aware of the benefits/drawbacks

Kiyoshika:
Also just noticed that I forgot to sign one of my commits, I really need to update my default

Kiyoshika (approved):

jaytaph:
I'll just squash commit this for now

PR #136, opened 2023-10-14
Implemented adjusted_insert_location function

jaytaph:
working version of the adjusted-insert-location.

This PR should implement a working algorithm as defined by https://html.spec.whatwg.org/multipage/parsing.html#creating-and-inserting-nodes

The main idea is to have the adjusted-insert-location function return not only a node-id which is the node that needs to be returned, but it's possible that this node is not in the main document, but instead in a template's document(fragment).  Besides this, it's possible to have a node inserted between children of a certain node. 

This means we have a new `NodeInsertLocation` structure that contains a document handle, a node_id, and a position.

  - handle defines which document(fragment) to insert into
  - node_id is the node in which to insert the new node
  - position is an optional index where the node should be inserted in the current children (0 = at the front, 1 = second place etc). If the position is larger than the length, it will be added to the end. If there is no position given (`None`), it will be added to the end of the children as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,7 @@ colored = "2.0.4"
 [dev-dependencies]
 criterion = { version = "0.5", features = ["html_reports"] }
 test-case = "3.2.1"
+
+[features]
+# Enables extended debugging information during parsing.
+debug_parser = []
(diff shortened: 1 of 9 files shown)


emwalker (approved):
LGTM!

I see checks are failing on one of the commits, so how about rebasing that and fixing the breaking checks or, alternatively, squashing?

PR #137, opened 2023-10-14
fear(css): add parser grammar rules (tokens & experssions)

neuodev:
This merge request adds: 
- [x]  Full AST struct types definitions
- [x] Definition of the CSS grammar rules in BNF format (will change in the future) 
- [x]  Basic parsing for the css "Rule"
- [x] Parse Id selector and class selector 
- [x] Parse the `!important` rule


jaytaph (approved):

jaytaph:
@neuodev it would be a good thing to rebase this PR into a single commit.

src/css/node.rs
```diff
@@ -0,0 +1,405 @@
...
+#[derive(Debug, PartialEq)]
+pub struct LayerList {
+    children: Vec<Layer>,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum MediaQueryChild {
+    Identifier(Identifier),
+    MediaFeature(MediaFeature),
+}
```
emwalker, line 252: @Kiyoshika this is the pattern I was thinking of, where you have an enum variant and then an inner struct with the same name.  We're almost there with [NodeData](https://github.com/gosub-browser/gosub-engine/blob/9fa2622869f5723d151ecb6e05aa454992008d76/src/html5_parser/node.rs#L30-L35), but not quite.

src/css/parser.rs
```diff
@@ -0,0 +1,326 @@
...
+                            important: true,
+                            property: "font-size".to_string(),
+                            value: ValueList::new(vec![Value::Dimension(Dimension::new(
+                                "1".to_string(),
+                                Some("rem".to_string())
+                            ))])
+                        }
+                    ]
+                ))])
+            ))])
```
emwalker, line 323: Another nice test.

src/css/tokens.rs
```diff
@@ -0,0 +1,66 @@
...
+static NAME: &str = r"^[_a-zA-Z0-9-]*";
+static HASH: &str = r"^#";
+static AT: &str = r"^@";
+static DOT: &str = r"^\.";
+static SEMICOLON: &str = r"^;";
+static COLON: &str = r"^:";
+static LCURLY: &str = r"^\{";
+static RCURLY: &str = r"^\}";
+static SPACE: &str = r"^[ \t\r\n\f]+";
+static IMPORTANT: &str = "^!important";
```
emwalker, line 52: As before, we'll want a real tokenizer at some point, I'm thinking, instead of using regexes.  But that can wait for a future day.

src/lib.rs
```diff
@@ -2,7 +2,7 @@ extern crate alloc;
 
 pub mod api;
 #[allow(dead_code)]
-pub mod css3_parser;
+pub mod css;
```
emwalker, line 5: I like this change.  Hopefully we can make similar changes elsewhere in the project.

emwalker (approved):
I like the idiomatic approach that you've followed, @neuodev, of using as many helper structs are are needed (in this case, to build an AST).  We should be quick to break up work and delegate it to smaller structs, instead of having huge structs with lots of fields and huge methods.

src/lib.rs
emwalker: Is your plan to pick up other versions of CSS?  Would `pub mod css3;` be better?

src/css/node.rs
Kiyoshika: @emwalker I almost followed this but thought it might be confusing having two structs named `Document` in the code base that served different purposes, which is why I settled on `...Data` suffix. Although my opinion is not that strong so I wouldn't mind refactoring that at a later point

Kiyoshika: (not to mention the `DocumentData` actually doesn't really do anything anyways)

Kiyoshika (approved):

src/css/node.rs
emwalker: Rust's module system allows the reuse of struct names in different contexts as many times as you need.  If you ever have to use them in the same file, you can do:
```rust
use crate::{
  dom::Element,
  node::data::element::Element as ElementData,
};
```

So there's less of a need to coordinate struct names globally like you would in some other languages.  (Now that I'm looking at this, I'm wondering why `node::` isn't under `dom::`.)

jaytaph: the reason is that the current "document" and nodes are an intermediate step before generating the DOM.   It made more sense in getting things up and running quickly.. but will eventually (and sooner rather than later), be moved to generating DOM as soon as possible, most likely removing the whole intermediate step..

src/lib.rs
jaytaph: Is css3 compatible with css2  (ie: can a css2 stylesheet be loaded by a css3 parser)?

emwalker: I suspect we would want separate `css3::` and `css2::` Rust modules at some point, possibly with a shared trait to serve as the interface.

jaytaph:
Thank for the pr @neuodev

PR #139, opened 2023-10-15
Add `DocumentHandle` to `ElementData` and `ElementAttributes`

Kiyoshika:
This is the first step in refactoring `ElementAttributes.insert()` to allow direct modification of the DOM which will remove `Document.set_named_id`. See [this discussion](https://github.com/gosub-browser/gosub-engine/discussions/123) for more details.

However, I didn't want to make too many changes at once to make it hard to follow, so this PR does the following things:
* Adds the `DocumentHandle` shared pointer to `ElementData` and `ElementAttributes`
* Moves `ElementData` to the heap with `Box<>` now that it has grown a little large
    * NOTE: we are not able to pattern match on attributes anymore with `Box<>` (at least, while the other items in the enum are stack-allocated) so any usage of `NodeData::Element` had to be slightly refactored
* Adds the `node_id` to the constructor for `ElementData` as we need to know the node ID when making changes to the DOM (later)
    * NOTE: `Document.add_node` was changed to update the node ID within the element data. In a follow up PR when `ElementAttributes.insert()` is refactored we will be able to clean up this method quite a bit; see `TODO:` comments
    * A new test was added to verify that the IDs in the underlying `ElementData` structs are properly updated when adding a node to the DOM (when node's are first created, they all use the default ID which is 0)
* Small cleanup in tests changing `NodeId(X)` to `NodeId::from(X)` to have some sort of consistency

### DIFF
--- a/src/html5_parser/parser/document.rs
+++ b/src/html5_parser/parser/document.rs
@@ -304,6 +304,12 @@ impl PartialEq for DocumentHandle {
     }
 }
 
+impl Clone for DocumentHandle {
+    fn clone(&self) -> DocumentHandle {
+        DocumentHandle(Rc::clone(&self.0))
+    }
+}
+
 impl Eq for DocumentHandle {}
 
 impl DocumentHandle {
--- a/src/html5_parser/parser/document.rs
+++ b/src/html5_parser/parser/document.rs
@@ -311,6 +311,10 @@ impl PartialEq for DocumentHandle {
     }
 }
 
+// NOTE: it is preferred to use Document::clone() when
+// copying a DocumentHandle reference. However, for
+// any structs using this handle that use #[derive(Clone)],
+// this implementation is required.
 impl Clone for DocumentHandle {
     fn clone(&self) -> DocumentHandle {
         DocumentHandle(Rc::clone(&self.0))
--- a/src/html5_parser/parser/document.rs
+++ b/src/html5_parser/parser/document.rs
@@ -181,6 +181,8 @@ impl Document {
 
     // Add to the document
     pub fn add_node(&mut self, node: Node, parent_id: NodeId) -> NodeId {
+        // TODO: this will be refactored and removed in an upcoming PR
+        // when we make modifications to ElementAttributes.insert()
         let mut node_named_id: Option<String> = None;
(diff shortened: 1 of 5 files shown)


src/html5_parser/parser/document.rs
```diff
@@ -304,6 +311,12 @@ impl PartialEq for DocumentHandle {
     }
 }
 
+impl Clone for DocumentHandle {
+    fn clone(&self) -> DocumentHandle {
+        DocumentHandle(Rc::clone(&self.0))
+    }
+}
```
emwalker, line 318: I like keeping `Document::clone(...)`, copying `Rc::clone(&reference)` instead of `reference.clone()`, as it communicates to the developer that we're only doing a fast, shallow clone of a reference.

src/html5_parser/parser/document.rs
```diff
@@ -478,23 +491,55 @@ mod tests {
             _ => panic!(),
         }
 
-        let _ = document.get_mut().add_node(node1, NodeId(0));
-        let _ = document.get_mut().add_node(node2, NodeId(1));
+        let _ = document.get_mut().add_node(node1, NodeId::from(0));
+        let _ = document.get_mut().add_node(node2, NodeId::from(1));
```
emwalker, line 495: `0.into()` and `1.into()` will also work here, btw.  Although I think `NodeId::from(0)`, etc., is good.

emwalker (approved):
LGTM with a question about keeping `Document::clone(...)` and not implementing `impl Clone` for `DocumentHandle`.

src/html5_parser/parser/document.rs
Kiyoshika: I also prefer that as well, but the `#[derive(Clone)]` tags (or whatever they're called) was complaining that it wasn't implemented

emwalker: Ah, gotcha.  Maybe a comment above the trait `impl` describing the preferred usage?

Kiyoshika: Can do!

Kiyoshika: Comment added

jaytaph (approved):

PR #141, opened 2023-10-15
Refactor `ElementAttributes` to directly modify DOM

Kiyoshika:
(DRAFT PR)

This is the progress on refactoring `ElementAttributes.insert` to directly modify the DOM in the special cases of `id` and `class` attribute. `Document.named_id_elements` will also eventually be renamed to `queryable_id_nodes` and `queryable_class_nodes` will be introduced at a later point when `Document.get_node_by_class_name(...)` is introduced.

Current state, some of the tests need to be refactored slightly to handle the API changes (some methods in `Node` and `Document` are removed) and the `class` special case needs to be implemented

### DIFF
--- a/src/html5_parser/node/data/element.rs
+++ b/src/html5_parser/node/data/element.rs
@@ -45,8 +45,44 @@ impl ElementAttributes {
         self.attributes.contains_key(name)
     }
 
+    /// according to HTML5 spec: 3.2.3.1
+    /// https://www.w3.org/TR/2011/WD-html5-20110405/elements.html#the-id-attribute
+    fn validate_named_id(&self, named_id: &str) -> bool {
+        if named_id.contains(char::is_whitespace) {
+            return false;
+        }
+
+        if named_id.is_empty() {
+            return false;
+        }
+
+        // must contain at least one character, but
+        // doesn't specify it should *start* with a character
+        if !named_id.contains(char::is_alphabetic) {
+            return false;
+        }
+
+        true
+    }
+
     /// Inserts a new attribute into the map.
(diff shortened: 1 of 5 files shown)


src/html5_parser/node/data/element.rs
```diff
@@ -45,8 +45,44 @@ impl ElementAttributes {
...
+        }
+
+        // must contain at least one character, but
+        // doesn't specify it should *start* with a character
+        if !named_id.contains(char::is_alphabetic) {
+            return false;
+        }
+
+        true
+    }
```
emwalker, line 66:
```suggestion
    fn validate_named_id(&self, named_id: &str) -> bool {
        if named_id.contains(char::is_whitespace) {
            return false;
        }

        if named_id.is_empty() {
            return false;
        }

        // must contain at least one character, but
        // doesn't specify it should *start* with a character
        named_id.contains(char::is_alphabetic)
    }
```

emwalker: (Sorry, remembering this PR is in draft right now.  Please ignore.)

Kiyoshika: No worries, I like the suggestion

jaytaph:
There are probably more conflict wuth regards to the node,/document. It should be easier to use now (at least more clear)

Kiyoshika:
> There are probably more conflict wuth regards to the node,/document. It should be easier to use now (at least more clear)

yeah I was just thinking of this recently. This branch is now pretty out of date and after Eric and I figure out how exactly we want to handle the new changes we will probably just recreate them ontop of the new main branch

Kiyoshika:
closing this as this is currently under a redesign

PR #145, opened 2023-10-16
Fixed adoption agency to pass failing test correctly

jaytaph:
- Fixed issues with the adoption agency which made a test (test1::31) not pass correctly.
 - also fixed some debug messages, as this resulted in a recursive loop and caused a stack overflow when outputting debug info.
 - generalized the insert_*_element functions so it can also accept created nodes instead of just tokens.

### DIFF
--- a/src/html5_parser/parser.rs
+++ b/src/html5_parser/parser.rs
@@ -3077,7 +3077,7 @@ impl<'stream> Html5Parser<'stream> {
         }
     }
 
-    /// Remove the given node_id from the active formatting elements list. Wil do nothing when the node is not found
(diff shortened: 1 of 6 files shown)


src/html5_parser/parser.rs
```diff
@@ -3154,7 +3077,7 @@ impl<'stream> Html5Parser<'stream> {
         }
     }
 
-    /// Remove the given node_id from the active formatting elements list
+    /// Remove the given node_id from the active formatting elements list. Wil do nothing when the node is not found
```
emwalker, line 3080:
```suggestion
    /// Remove the given node_id from the active formatting elements list. Will do nothing when the node is not found
```

emwalker (approved):

PR #147, opened 2023-10-17
Separate parser test validation logic from display logic

emwalker:
This PR separates the parser test display logic from the validation logic.  In a later PR, this will make it easier to integrate the validation logic into the CI run without interfering with display concerns.

### DIFF
--- a/src/css3/parser.rs
+++ b/src/css3/parser.rs
@@ -562,12 +562,12 @@ mod test {
 
         assert_eq!(
             parser.parse(
-                r##"
+                r#"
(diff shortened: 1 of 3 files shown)


src/css3/parser.rs
```diff
@@ -562,12 +562,12 @@ mod test {
 
         assert_eq!(
             parser.parse(
-                r##"
+                r#"
```
emwalker, line 565: The result of running `make fix`.

jaytaph (approved):

PR #149, opened 2023-10-17
Cleanup of doc/arena/node

jaytaph:
This is a cleanup of the document, arena and node structures.

The arena is very simple now: you just can get a (mut) node, and you can "register" a node. Registering a node means that you have created a node (somewhere else, it doesn't matter), and you register it into the arena. From this point on, it has a (valid) NodeID and the node is considered "registered".

The node structure has a "is_registered" boolean, which tells us if the node we are using is already registered or not. This boolean is probably not needed, but currently it can be useful to see when I'm trying to register an already registered node (shouldn't happen), or when a work with a unregistered node (also shouldn't happen).

The last structure changed is the "document" structure. A document knows about nodes and arena's and can communicate between them

There is a single `add_node` inside the `documentHandle` structure, which calls the `add_node` function in the document itself. This will add a node to the given parent (and optional child position). It will also take care of registering the node if not already done so.

Also, there are two functions to attach/detach nodes that are already registered:

  - `attach_node_to_parent`
  - `detach_node_from_parent`

They do what it says: adds or removes the given node from/to a parent. All add/remove/insert node functionality is removed from the node structure itself.

Furthermore it has a "has_cyclic_reference", which checks if a given node has another node as its parent.

Finally, there is the "relocate" function, which removed a (registered) node from its parent, and move it
to another parent. This is sometimes needed in the parsing algorithm, where nodes gets shifted around when
there are mismatched tags for instance.

It's a bit of a refactoring, but it should make the whole process much clearer.

### DIFF
--- a/src/html5_parser/parser/document.rs
+++ b/src/html5_parser/parser/document.rs
@@ -259,7 +259,7 @@ impl Document {
         if let Some(parent_node) = self.get_node_by_id_mut(parent_id) {
             // Make sure position can never be larger than the number of children in the parent
             if let Some(mut position) = position {
-                if position >= parent_node.children.len() {
+                if position > parent_node.children.len() {
                     position = parent_node.children.len();
                 }
                 parent_node.children.insert(position, node_id);
--- a/src/html5_parser/parser/document.rs
+++ b/src/html5_parser/parser/document.rs
@@ -1,8 +1,8 @@
 use crate::html5_parser::node::arena::NodeArena;
 use crate::html5_parser::node::data::{comment::CommentData, text::TextData};
+use crate::html5_parser::node::NodeTrait;
 use crate::html5_parser::node::NodeType;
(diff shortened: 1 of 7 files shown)


src/html5_parser/parser/document.rs
```diff
@@ -206,52 +221,131 @@ impl Document {
...
-            .unwrap()
-            .children
-            .push(node_id);
-        self.arena.get_node_mut(node_id).unwrap().parent = Some(parent_id);
+        if let Some(parent_node) = self.get_node_by_id_mut(parent_id) {
+            // Make sure position can never be larger than the number of children in the parent
+            if let Some(mut position) = position {
+                if position >= parent_node.children.len() {
+                    position = parent_node.children.len();
+                }
```
Kiyoshika, line 264: Minor: could probably change this to `>`

Kiyoshika (approved):

PR #152, opened 2023-10-18
feat(css): impl. css tokenizer (comments & numbers)

neuodev:


src/css3/new_tokenizer.rs
```diff
@@ -0,0 +1,161 @@
+// note: input_stream should come from a shared lib.
+use crate::html5_parser::input_stream::InputStream;
```
emwalker, line 2: I agree with the comment.  Although I can also see there being an html5 token stream and a css3 token stream, and perhaps they implement a shared interface (don't have any insights at this point).

src/css3/new_tokenizer.rs
```diff
@@ -0,0 +1,161 @@
...
+
+            if self.stream.current_char().utf8() == '-' || self.stream.current_char().utf8() == '+'
+            {
+                value.push(self.stream.read_char().utf8());
+            }
+        }
+
+        value.push_str(&self.consume_digits());
+
+        Number::new(kind, value.parse().unwrap())
```
emwalker, line 90: Have you put thought into how errors will percolate up to the caller?

src/css3/new_tokenizer.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    #[test]
+    fn parse_numbers() {
+        let mut is = InputStream::new();
+        is.read_from_str(
+            "12#+34#-56#7.8#-9.10#0.0001#1e+1#1e1#1e-1",
+            Some(Encoding::UTF8),
+        );
+
```
emwalker, line 139: I like these tests.

emwalker (approved):

src/css3/new_tokenizer.rs
emwalker: Ok, I understand now.

In cases where we know that the call will succeed, despite the `.unwrap()`, how about using `.expect()` instead, to communicate to developers that this call will always succeed?

src/css3/new_tokenizer.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub fn consume_chars(&mut self, mut len: usize) -> String {
+        let mut value = String::new();
+
+        while len > 0 {
+            println!("value_={:?}", self.stream.current_char().utf8());
+            value.push(self.stream.read_char().utf8());
+            len -= 1
+        }
+
+        println!("v={}", value);
```
emwalker, line 111: Did you intend to keep these `println!` statements?  Should we use `logger` instead?

emwalker (approved):
LGTM with question.

src/css3/new_tokenizer.rs
emwalker: Any error message that will allow the developer to find the code again, e.g., this (or anything):

```rust
let number = value.parse().expect("error parsing number");
Number::new(kind, number)
```

How about all lowercase, and no capital letters?

src/css3/new_tokenizer.rs
```diff
@@ -0,0 +1,165 @@
...
+            }
+        }
+
+        value.push_str(&self.consume_digits());
+
+        Number::new(
+            kind,
+            value
+                .parse()
+                .expect("Unexpected error, should never happen"),
```
emwalker, line 93:
```suggestion
                .expect("failed to parse number"),
```

jaytaph (approved):

PR #154, opened 2023-10-19
feat(css): consume ident and ident-like tokens (e.g. url, function)

neuodev:


jaytaph (approved):

src/css3/new_tokenizer.rs
```diff
@@ -89,6 +101,148 @@ impl<'stream> Tokenizer<'stream> {
...
+    ///
+    /// Returns either a `<url-token>` or a `<bad-url-token>`
+    pub fn consume_url(&mut self) -> Token {
+        let mut url = String::new();
+
+        self.consume_whitespace();
+
+        loop {
+            if self.stream.current_char().utf8() == ')' {
+                println!("here..");
```
Kiyoshika, line 139: Just adding a reminder to remove this `println` later, assuming it was being used for debugging

src/css3/new_tokenizer.rs
```diff
@@ -131,30 +335,128 @@ mod test {
...
+            assert_eq!(tokenizer.consume_ident_like_seq(), url_token);
+        }
+    }
+
+    #[test]
+    fn parse_function_tokens() {
+        let mut is = InputStream::new();
+
+        let functions = vec![
+            ("url(\"", Token::Function("url".into())),
```
Kiyoshika, line 447: Should the other cases mentioned in 4.3.4 be added here such as `url( \"`, `url(\'` and `url( \')`?

Kiyoshika (approved):

emwalker (approved):
I think we could all learn some Rust from you.  Always a pleasure to review your PRs.

I'm wondering whether this approach, with its small functions, would lend itself to the HTML5 parser code?  I don't know what that would look like.  I also don't know how the relative performance would compare.

PR #157, opened 2023-10-20
Get some parser tests to pass

emwalker:
The main point of this PR was to get some additional parser tests to pass:

This PR:
* Gets some parser tests to pass
* Uses an index map to preserve order for element attributes, but hides it behind an `AttributeMap` type alias
* Adds the ability to render a document in the format used in the tree construction tests

I assume order is not important when interpreting attributes in an element.  But in order to reliably reproduce the printed tree that is used in the tree construction tests, we can't use `HashMap`, which randomizes the order of the keys.  Instead we use `IndexMap`, which preserves insertion order.  In order not to create too much of a dependency on the `indexmap` library in case we want to replace it in the future, an `AttributeMap` type alias has been used in the code in place of `IndexMap`.

### DIFF
--- a/src/html5/node/data/element.rs
+++ b/src/html5/node/data/element.rs
@@ -1,9 +1,9 @@
 use crate::html5::element_class::ElementClass;
 use crate::html5::node::NodeId;
 use crate::html5::parser::document::{Document, DocumentFragment, DocumentHandle};
+use crate::types::AttributeMap;
 use core::fmt::{Debug, Formatter};
-use std::collections::hash_map::Iter;
-use std::collections::HashMap;
+
 use std::fmt;
 
 #[derive(Debug, PartialEq, Clone)]
@@ -14,7 +14,7 @@ pub(crate) struct ElementAttributes {
     /// Pointer to the document that the node associated with these attributes are tied to
     pub(crate) document: DocumentHandle,
     /// Key-value pair of all attributes
-    attributes: HashMap<String, String>,
+    attributes: AttributeMap,
 }
 
 /// This is a very thin wrapper around a HashMap.
@@ -26,14 +26,14 @@ impl ElementAttributes {
         Self {
             node_id,
             document,
-            attributes: HashMap::new(),
+            attributes: AttributeMap::new(),
         }
     }
 
     pub(crate) fn with_attributes(
         node_id: NodeId,
         document: DocumentHandle,
-        attributes: HashMap<String, String>,
+        attributes: AttributeMap,
     ) -> Self {
         Self {
             node_id,
@@ -78,19 +78,19 @@ impl ElementAttributes {
     }
 
     /// Returns an iterator over the attribute map.
-    pub(crate) fn iter(&self) -> Iter<'_, String, String> {
+    pub(crate) fn iter(&self) -> impl Iterator<Item = (&String, &String)> {
         self.attributes.iter()
     }
 
     /// Adds the given attributes to the attribute map.
-    pub(crate) fn copy_from(&mut self, attribute_map: HashMap<String, String>) {
+    pub(crate) fn copy_from(&mut self, attribute_map: AttributeMap) {
         for (key, value) in attribute_map.iter() {
             self.insert(key, value);
         }
     }
 
     /// Clones the internal map of attributes (NOT the attributes object itself)
-    pub(crate) fn clone_map(&self) -> HashMap<String, String> {
+    pub(crate) fn clone_map(&self) -> AttributeMap {
         self.attributes.clone()
     }
 }
@@ -138,7 +138,7 @@ impl ElementData {
         node_id: NodeId,
         document: DocumentHandle,
         name: &str,
-        attributes: HashMap<String, String>,
+        attributes: AttributeMap,
     ) -> Self {
         Self {
             node_id,
--- a/src/html5/parser.rs
+++ b/src/html5/parser.rs
@@ -5,7 +5,6 @@ mod quirks;
 // ------------------------------------------------------------
 
 use self::document::DocumentHandle;
-
 use super::node::NodeId;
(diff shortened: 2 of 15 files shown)


src/html5/node/data/element.rs
```diff
@@ -78,19 +78,19 @@ impl ElementAttributes {
     }
 
     /// Returns an iterator over the attribute map.
-    pub(crate) fn iter(&self) -> Iter<'_, String, String> {
+    pub(crate) fn iter(&self) -> impl Iterator<Item = (&String, &String)> {
```
emwalker, line 81: `impl Iterator` is generic and can be used almost anywhere, while `Iter` is a concrete type.

tests/tree_construction.rs
```diff
@@ -25,8 +25,7 @@ const DISABLED_CASES: &[&str] = &[
     "<!DOCTYPE html><script></script>  <title>x</title>  </head>",
     "<!DOCTYPE html><html><body><html id=x>",
     r#"<!DOCTYPE html>X</body><html id="x">"#,
-    "<!DOCTYPE html>X</html>",
-    "<!DOCTYPE html>X<p/x/y/z>",
+    "<!DOCTYPE html>X</html>", // Test formatting issue?
```
emwalker, line 28: Dropping the override for the `<!DOCTYPE html>X<p/x/y/z>` test case (line 29) and getting 28 to pass in a unit test was what most of this PR is about.

src/html5/parser.rs
```diff
@@ -3905,4 +3894,56 @@ mod test {
...
+        assert_same(
+            &document,
+            r#"
+            | <!DOCTYPE html>
+            | <html>
+            |   <head>
+            |   <body>
+            |     "X "
+            "#,
+        );
```
emwalker, line 3927: The fact that this test passes suggests that we're not handling it right in the other test code, making it a false negative.

src/html5/parser.rs
```diff
@@ -2103,10 +2103,9 @@ impl<'stream> Html5Parser<'stream> {
                     return;
                 }
 
+                // Ignore for now.
                 // @TODO: Other stuff
-
-                self.insertion_mode = InsertionMode::AfterBody;
-                self.reprocess_token = true;
+                return;
```
emwalker, line 2108: This change helps to get one of the unit tests below to pass.

src/html5/parser.rs
```diff
@@ -1975,11 +1979,7 @@ impl<'stream> Html5Parser<'stream> {
                 // ignore token
             }
             Token::TextToken { .. } if self.current_token.is_empty_or_white() => {
-                self.reconstruct_formatting();
-
-                let node = self.create_node(&self.current_token, HTML_NAMESPACE);
-                let parent_node = current_node!(self);
-                self.document.get_mut().add_node(node, parent_node.id, None);
+                self.create_or_merge_text(self.current_token.clone());
```
emwalker, line 1982: This change helps to get one of the unit tests below to pass.

src/html5/parser.rs
```diff
@@ -3905,4 +3894,56 @@ mod test {
...
+            &document,
+            r#"
+            | <!DOCTYPE html>
+            | <html>
+            |   <head>
+            |   <body>
+            |     "XX"
+            "#,
+        );
+    }
```
emwalker, line 3912: Test `tests2_dat_720` was passing before this PR, and I just used it to debug the tree display format.

src/html5/parser.rs
```diff
@@ -3905,4 +3894,56 @@ mod test {
...
+            | <!DOCTYPE html>
+            | <html>
+            |   <head>
+            |   <body>
+            |     "X"
+            |     <p>
+            |       x=""
+            |       y=""
+            |       z=""            
+            "#,
```
emwalker, line 3946: We need a stable key ordering in order to pass this and similar tests.

src/html5/parser.rs
jaytaph: This should not happen this way. Reprocess token means the current token is well, reprocessed so it should stay in the loop. I think there is something else not quite ok in the parser (which might take a while to find,.. there sometimes really nasty issues that are very easily overlooked)

src/html5/parser.rs
jaytaph: This makes sort of sense, but is not right. The add_node uses parent_id, which may or may not be the current_token.  The "Create_or_merge" basically checks if the last token is a text token, and merges any text into it.

So it's an improvement on the merging text part.. but it needs to have parent_node.id as an override of some sorts

src/html5/parser.rs
jaytaph: i think we can order attributes alphabetically.. we might need to have a separate Vec<usize> for ordering besides the attribute hashmap (or have some kind of "ordered" hashmap that does this for us)

src/html5/parser.rs
emwalker: Ok, no worries.  I can back out this change and add `#[ignore]` to the unit test.

src/html5/parser.rs
emwalker: I like not preprocessing the test input.  Perhaps we can parameterize this type for use in tests.  Ordering alphabetically incurs `O(n log(n))` overhead (for small n).

src/html5/parser.rs
emwalker: I've backed that change out.  Apparently it was an experiment that I tried as I was poking around that wasn't needed.

I'd argue the larger issue is that there's no test (of the `make test` variety) currently protecting it.  The code that was dropped was presumably important, so there should be a test that fails the checks when it's dropped.

emwalker:
I think maybe this PR might not be opportune at this time.  I'll close it, and maybe we can revisit some of the topics at a later point.

PR #160, opened 2023-10-20
fixed <li> tests

jaytaph:
Added parser code for <li> tags, which was not implemented yet.

### DIFF
--- a/src/html5/parser.rs
+++ b/src/html5/parser.rs
@@ -1723,7 +1723,9 @@ impl<'stream> Html5Parser<'stream> {
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/html5/parser.rs
```diff
@@ -1723,7 +1723,9 @@ impl<'stream> Html5Parser<'stream> {
         }
     }
 
-    fn flush_pending_table_character_tokens(&mut self) {}
+    fn flush_pending_table_character_tokens(&mut self) {
+        todo!()
```
jaytaph, line 1727: This needs a later implementation

PR #166, opened 2023-10-22
fix some test

CharlesChen0823:
found some test case failed in tree_constructions for tests3.dat. this is correctly behavior, because current for tag `<pre>` and `<textarea>` not implement ignore next `LF` token.


CharlesChen0823:
current close this PR, send another PR.

PR #173, opened 2023-10-23
fixed

CharlesChen0823:
because my current person branch has many different, so specify current modify for which test is difficult.

### DIFF
--- a/src/html5/parser.rs
+++ b/src/html5/parser.rs
@@ -2009,7 +2009,7 @@ impl<'stream> Html5Parser<'stream> {
                 }
 
                 // Add attributes to html element
-                let first_node_id = self.open_elements.first().unwrap().clone();
+                let first_node_id = *self.open_elements.first().unwrap();
                 let mut doc = self.document.get_mut();
                 let first_node = doc
                     .get_node_by_id_mut(first_node_id)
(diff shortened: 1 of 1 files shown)


src/html5/parser.rs
```diff
@@ -112,17 +112,6 @@ macro_rules! current_node {
     }};
 }
 
-macro_rules! current_node_mut {
```
jaytaph, line 115: ah, this is nice.. there was only one place that still uses this.. and we cannot get rid of ALL the macro's (yet).. but the more we can remove, the better

src/html5/parser.rs
```diff
@@ -2817,6 +2805,9 @@ impl<'stream> Html5Parser<'stream> {
                 self.parse_error("doctype not allowed in before head insertion mode");
                 // ignore token
             }
+            Token::StartTagToken { name, .. } if name == "html" => {
```
jaytaph, line 2808: does this fix a test or did you just happen to find some unimplemented start-tag?

CharlesChen0823: in testing find this issue, and then reading docs, found `html` start-tag not implemented.

src/html5/parser.rs
CharlesChen0823: because found an issue in `parser.rs`, which is using the macro, then fixed that issue, the macro is not used, so delete.

jaytaph:
@CharlesChen0823 thanks for the pr

PR #179, opened 2023-10-24
pass more tests

CharlesChen0823:
add more test, but found the method  `run_adoption_agency` might have some problem.

### DIFF
--- a/src/html5/parser.rs
+++ b/src/html5/parser.rs
@@ -3032,7 +3032,6 @@ impl<'stream> Html5Parser<'stream> {
                 } else {
                     self.parse_error("optgroup end tag not allowed in in select insertion mode");
                     // ignore token
-                    return;
                 }
             }
             Token::EndTagToken { name, .. } if name == "option" => {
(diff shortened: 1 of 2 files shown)


jaytaph:
@CharlesChen0823  Thanks for the PR..  The adoption_agency is a nasty algorithm to get right. I assume there are still a few things not quite right yet but fortunately we have good tests to check this.

jaytaph:
@CharlesChen0823 are you also on slack, maybe we can coordinate parser fixes so we don't do the same tests? :)

CharlesChen0823:
> @CharlesChen0823 are you also on slack, maybe we can coordinate parser fixes so we don't do the same tests? :)

Because network problem, I cannot directly connect slack. 
I had rewrite the adoption_agency_algorithm and find the appropriate insert place method,  and current the implement pass more test, except template case and fragment parser case. because for simple, current not support template case in my implement.

I will send an new PR soon.

jaytaph:
we can move to discussions otherwise: https://github.com/gosub-browser/gosub-engine/discussions/182

PR #183, opened 2023-10-25
Fixed foreign elements (svg, mathml)

jaytaph:
This PR fixes the situation where we are parsing foreign elements (svg and mathml). In this case, the parser should use a different processing than for html elements. 

I've moved most of the big if-statement for html elements into a separate function, implemented the foreign element statements into a separate function, and have a "dispatcher" that checks which one to call.

There is some small issues with EOF during the testing of tests16.dat, so for now I've deactivated this for now, but it's only one test that fails as regression, so that will be fixed quickly.


jaytaph:
@CharlesChen0823  Would you mind to take a look at this PR as well, as it affects your work.

CharlesChen0823:
@jaytaph I send an PR, this can merge. but I move the template content.

jaytaph:
@CharlesChen0823 Since your work is smaller, I'll wait until you are finished with your PR..  I will merge your PR first.. and then i will rebase my work on top of that and merge that.

tests/tree_construction.rs
```diff
@@ -46,9 +63,11 @@ lazy_static! {
 #[test_case("tests1.dat")]
 #[test_case("tests2.dat")]
 #[test_case("tests3.dat")]
+#[test_case("tests4.dat")]
 #[test_case("tests5.dat")]
 #[test_case("tests6.dat")]
 #[test_case("tests8.dat")]
+#[test_case("tests10.dat")]
```
emwalker, line 70: Nice.

src/bin/parser_test.rs
```diff
@@ -31,34 +31,35 @@ fn main() -> Result<()> {
...
 
-    let filenames = Some(&["tests3.dat"][..]);
+    let filenames = Some(&["tests16.dat"][..]);
     let fixtures = testing::tree_construction::fixtures(filenames).expect("fixtures");
 
     for fixture_file in fixtures {
         println!(
-            "🏃‍♂️ Running {} tests from 🗄️ {:?}\n",
+            "\
+            🏃‍♂️ Running {} tests from 🗄️ {:?}",
```
emwalker, line 40: Not sure if this is the intended outcome with this formatting:



Might look different on Windows versus Linux.

src/bin/parser_test.rs
```diff
@@ -31,34 +31,35 @@ fn main() -> Result<()> {
...
-    );
+        println!(
+            "\
+    🏁 Tests completed: Ran {} tests, {} assertions, {} succeeded, {} failed ({} position failures)\n",
+            results.tests,
+            results.assertions,
+            results.succeeded,
+            results.failed,
+            results.failed_position
+        );
```
emwalker, line 61: Your statement is indented now, but I'm not seeing it indented.  Something I don't fully understand is going on (not that it's important).

src/html5/parser/attr_replacements.rs
```diff
@@ -2,6 +2,46 @@ pub static MATHML_ADJUSTMENTS: phf::Map<&str, &str> = phf::phf_map! {
     "definitionurl" => "definitionURL",
 };
 
+pub static SVG_TAGNAMES: phf::Map<&str, &str> = phf::phf_map! {
```
emwalker, line 5: I learned about the `phf` crate from this PR.

emwalker (approved):
LGTM.

src/bin/parser_test.rs
jaytaph: unicode emoticons seems to be displayed a bit off on my terminal. It won't matter much anyway so i'll revert this back

jaytaph:
Closed in favor of #185

PR #188, opened 2023-10-27
Add a nom parser for the tree construction tests

emwalker:
The purpose of this PR is to parse the `html5lib-tests` tree construction tests with a [nom parser](https://docs.rs/nom/latest/nom/), and to provide unit tests for the parser as issues are identified in the tree construction test parsing later on.  While we're here, let's replace the boolean `scripting_enabled` flag with a three-state `ScriptMode` enum.

### DIFF
--- a/src/testing/tree_construction/parser.rs
+++ b/src/testing/tree_construction/parser.rs
@@ -1,0 +1,784 @@
+// See https://github.com/html5lib/html5lib-tests/tree/master/tree-construction
+use crate::types::{Error, Result};
+use nom::{
+    branch::alt,
+    bytes::complete::{tag, take_until, take_until1},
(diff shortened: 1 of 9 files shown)


src/testing/tree_construction/parser.rs
```diff
@@ -0,0 +1,784 @@
...
+
+    let (_, tests) = all_consuming(files)(Span::new(&input))
+        .finish()
+        .map_err(|err| Error::Test(format!("{}", err)))?;
+
+    Ok(tests)
+}
+
+#[cfg(test)]
+mod tests {
```
emwalker, line 324: These unit tests and the parser they test are the main purpose of this PR.

jaytaph (approved):

PR #192, opened 2023-10-28
Bugfix: deal with some bugs in parsing some of the other .dat files

emwalker:
When reading through Slack, I noticed that there is a new `html5-parser-test` binary under `src/bin`. When I tried to run it after merging my latest PR, I saw that it exits early because the new tree construction test parsing code is unable to handle several corner cases.  This PR fixes that handling.

While we're here, let's add `--features debug_logger` to the test run, so that we can catch any compilation issues when this feature is enabled.  Otherwise Rust will ignore the code behind the feature, and you might have broken the feature without knowing it.

```sh
$ cargo run --color=always --package gosub-engine --bin html5-parser-test
```

Before this PR:

After this PR:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ SHELL=/usr/bin/env bash -O globstar
 
 all: help
 
-test: test_unit test_clippy test_fmt ## Runs tests
+test: test_commands test_unit test_clippy test_fmt ## Runs tests
 
 bench:
 	cargo bench
@@ -21,7 +21,7 @@ fix:
 test_unit:
 	source test-utils.sh ;\
 	section "Cargo test" ;\
-	cargo test
+	cargo test --features debug_parser
 
 test_clippy:
 	source test-utils.sh ;\
@@ -32,6 +32,10 @@ test_fmt:
 	source test-utils.sh ;\
 	section "Cargo fmt" ;\
 	cargo fmt -- --check
+
+test_commands:
+	cargo run --bin html5-parser-test >/dev/null
+	cargo run --bin parser_test >/dev/null
 
 help: ## Display available commands
 	echo "Available make commands:"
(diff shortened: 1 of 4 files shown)


src/testing/tree_construction.rs
```diff
@@ -445,7 +447,11 @@ fn document(s: &str) -> Vec<String> {
                 Some(format!("|{}", l.trim_end()))
             }
         })
-        .collect::<Vec<_>>()
+        .collect::<Vec<_>>();
+
+    // TODO: drop the following line
+    document.push("".into());
+    document
```
emwalker, line 454: It looks like some of the newer tests to be parsed are relying (perhaps incorrectly) on there being an extra blank string at the end of the document array.

src/testing/tree_construction/parser.rs
```diff
@@ -306,7 +308,8 @@ fn test(i: Span) -> IResult<Span, TestSpec> {
 }
 
 pub fn parse_str(i: &str) -> Result<Vec<TestSpec>> {
-    let input = i.to_owned() + "\n";
+    // Deal with a corner case that makes it hard to parse tricky01.dat.
+    let input = i.replace("\"\n\n\"", QUOTED_DOUBLE_NEWLINE).to_owned() + "\n";
```
emwalker, line 312: The challenge with this one is that every test but this one is delimited by `\n\n`, which is what is currently being used as the delimiter between tests.

Makefile
```diff
@@ -33,6 +33,10 @@ test_fmt:
 	section "Cargo fmt" ;\
 	cargo fmt -- --check
 
+test_commands:
+	cargo run --bin html5-parser-test >/dev/null
+	cargo run --bin parser_test >/dev/null
+
```
emwalker, line 39: While we're here, let's run these commands during the test run so that any errors that are introduced fail the CI run.  I was going to continue the pattern of the others, of including a `section "Commands"`, but for some reason it was swallowing nonzero exit codes for these commands.

jaytaph (approved):

PR #193, opened 2023-10-28
Added documentBuilder struct

jaytaph:
(this PR is an initial step in supporting fragment cases)

Currently, creating a document must be done in 2 steps:

1. create the document
2. create the root node in the document

This is a 2-step process, as we cannot create a root node inside the constructor of the document.

We moved this to a factory/builder pattern where we can now simply call:

```
DocumentBuilder::new_document()
```

and which will correctly create the document AND set the root node.

Also, we have moved the document to the new() constructor instead of the parse() function. Basically, we should not even need to initialize the parser, as it's a one-time function (we cannot call the parser twice, since it will be in a non-init state).


Kiyoshika (approved):

src/html5/node/arena.rs
```diff
@@ -152,239 +152,4 @@ mod tests {
...
-    //
-    //     let parent = arena.get_node(parent_id);
-    //     assert!(parent.is_none());
-    //     let child = arena.get_node(child_id);
-    //     assert!(child.is_none());
-    //     let grandchild = arena.get_node(grandchild_id);
-    //     assert!(grandchild.is_none());
-    //     let sibling = arena.get_node(sibling_id);
-    //     assert!(sibling.is_none());
-    // }
```
emwalker, line 389: I'm assuming we won't need these tests, because we already have coverage elsewhere.

src/html5/parser.rs
```diff
@@ -187,12 +187,14 @@ pub struct Html5Parser<'stream> {
     parser_pause_flag: bool,
     /// Keeps the position of where any document.write() should be inserted when running a script
     insertion_point: Option<usize>,
-    // ignore when next token is LF
+    /// ignore when next token is LF
```
emwalker, line 190:
```suggestion
    /// Ignore when next token is LF
```

src/html5/parser.rs
```diff
@@ -187,12 +187,14 @@ pub struct Html5Parser<'stream> {
...
     ignore_lf: bool,
-    // Sometimes tokens needs to be split up (and it seems the tokenizer cannot do this?)
+    /// Sometimes tokens needs to be split up (and it seems the tokenizer cannot do this?)
     token_queue: Vec<Token>,
-    // When true, the parser is finished and should not consume more tokens (there aren't any)
+    /// When true, the parser is finished and should not consume more tokens (there aren't any)
     parser_finished: bool,
+    /// Context node for fragment parsing
+    context: Option<NodeId>,
```
emwalker, line 197: Is there a way to make this non-option?
1. In the fragment case, you have some random NodeId
2. In the non-fragment case, you have the root id

src/html5/parser.rs
```diff
@@ -211,17 +213,11 @@ enum DispatcherMode {
 }
 
 impl<'stream> Html5Parser<'stream> {
-    // Creates a new parser object with the given input stream
-    pub fn new(stream: &'stream mut InputStream) -> Self {
+    // Initializes the parser for whole document parsing
+    pub fn new(stream: &'stream mut InputStream, document_handle: DocumentHandle) -> Self {
```
emwalker, line 217:
```suggestion
    pub fn new(stream: &'stream mut InputStream, document: DocumentHandle) -> Self {
```

src/html5/parser.rs
```diff
@@ -243,23 +239,79 @@ impl<'stream> Html5Parser<'stream> {
...
+        }
+    }
+
+    // Initializes the parser for document fragment parsing
+    pub fn new_fragment(
+        stream: &'stream mut InputStream,
+        document_handle: DocumentHandle,
+        context: NodeId,
+        form_element: Option<NodeId>,
+    ) -> Self {
```
emwalker, line 260:
```suggestion
    pub fn new_fragment(
        stream: &'stream mut InputStream,
        document: DocumentHandle,
        context: NodeId,
        form_element: Option<NodeId>,
    ) -> Self {
```

emwalker: Etc.

src/html5/node/arena.rs
jaytaph: yes. This is not the place for these tests anymore anyway.

src/html5/parser.rs
```diff
@@ -3878,8 +3943,9 @@ mod test {
 
     #[test]
     fn is_in_scope_7() {
+        let document = Document::shared();
         let mut stream = InputStream::new();
-        let mut parser = Html5Parser::new(&mut stream);
+        let mut parser = Html5Parser::new(&mut stream, document.clone());
```
emwalker, line 3948:
```suggestion
        let mut parser = Html5Parser::new(&mut stream, Document::clone(&document));
```

This usage makes people think of `Rc::clone(&reference)`, and communicates to the developer that it's a shallow copy of a reference rather than a deep copy of the contents.

emwalker: Etc.

src/html5/parser/document.rs
```diff
@@ -485,17 +475,38 @@ impl DocumentHandle {
     }
 }
 
+/// This struct will be used to create a fully initialized document or document fragment
+pub struct DocumentBuilder {}
```
emwalker, line 479:
```suggestion
pub struct DocumentBuilder;
```

(The [unit struct](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#unit-like-structs-without-any-fields).) This can be updated when there are actual fields.

emwalker (approved):
LGTM with thoughts / suggestions.

See also my [thoughts here](https://github.com/gosub-browser/gosub-engine/discussions/191#discussioncomment-7408103).  For this reason, it would be nice to have fragment parsing be the default case, and anything beyond that a special case intended for testing or something.  For that reason it would be nice to use `context: NodeId` instead of `context: Option<NodeId>`.

src/html5/parser.rs
jaytaph: in theory yes.. The `context` node is only set when `is_fragment_case` is `true`.  This can cause an issue when `is_fragment_case` is set, but the `context` node isn't (we use `expect()` to fetch the context in fragment cases). 

However, the only place the `context` and `is_fragment_case` are set, is set is in the `new_*` constructors.  We could remove `is_fragment_case` and have a function `is_fragment_case()` that checkx if `context` is `Some()`, but we couldn't set the `context` to root in non-fragment cases.

However, next iteration already changes this context node so I'll see what we can do there to make it less error-prone

jaytaph:
Closed in favor of #199

PR #195, opened 2023-10-28
Handle all script modes in the tree construction tests

emwalker:
This PR adds support for testing both `#script-on` and `#script-off` when these flags are present or absent in a tree construction test spec.

In this case, neither `#script-on` nor `#script-off` is present, so [both modes](https://github.com/html5lib/html5lib-tests/tree/master/tree-construction#tree-construction-tests) should be used for the test:

```
#data
FOO<!-- BAR --!>BAZ
#errors
(1,3): expected-doctype-but-got-chars
(1,15): unexpected-bang-after-double-dash-in-comment
#new-errors
(1:16) incorrectly-closed-comment
#document
| <html>
|   <head>
|   <body>
|     "FOO"
|     <!--  BAR  -->
|     "BAZ"
```

### DIFF
--- a/src/testing/tree_construction.rs
+++ b/src/testing/tree_construction.rs
@@ -137,20 +137,25 @@ impl Test {
     }
 
     /// Runs the test and returns the result
-    pub fn run(&self) -> Result<TestResult> {
-        let (actual_document, actual_errors) = self.parse()?;
-        let root = self.match_document_tree(&actual_document.get());
+    pub fn run(&self) -> Result<Vec<TestResult>> {
+        let mut results = vec![];
 
-        Ok(TestResult {
-            root,
-            actual_document,
-            actual_errors,
-        })
+        for &scripting_enabled in self.script_modes() {
+            let (actual_document, actual_errors) = self.parse(scripting_enabled)?;
+            let root = self.match_document_tree(&actual_document.get());
+            results.push(TestResult {
+                root,
+                actual_document,
+                actual_errors,
+            });
+        }
+
+        Ok(results)
     }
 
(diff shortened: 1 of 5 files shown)


src/testing/tree_construction.rs
```diff
@@ -137,20 +137,25 @@ impl Test {
...
+
+        for &scripting_enabled in self.script_modes() {
+            let (actual_document, actual_errors) = self.parse(scripting_enabled)?;
+            let root = self.match_document_tree(&actual_document.get());
+            results.push(TestResult {
+                root,
+                actual_document,
+                actual_errors,
+            });
+        }
```
emwalker, line 151: This is the heart of this PR.

Kiyoshika (approved):

src/testing/tree_construction.rs
jaytaph: Didnt i implement this already? Or did i miss a case somewhere?

emwalker: [This is what was there up until recently](https://github.com/gosub-browser/gosub-engine/blob/581dd3501c7b760a035eaac306f34c75356f40fa/src/testing/tree_construction.rs#L414-L419).  It implies a single run for a test — either scripting is enabled, or it's not enabled.  By contrast, the spec implies 1-2 runs: [on, off, or both](https://github.com/html5lib/html5lib-tests/tree/master/tree-construction#tree-construction-tests).

This PR picks up the "both" case (2 iterations of the test), which we weren't doing before.

jaytaph (approved):

PR #196, opened 2023-10-29
Begin to revisit tokenization

emwalker:
This PR begins to revisit tokenization:
* Further implement deserialization of the tokenization tests, and add unit tests
* Drop the `TokenTrait` trait, `TokenType` enum, and `type_of` method, as `matches!` and `match` can be used here instead
* Simplify assertions in pass/fail CI tests by using partial equality to compare tokens
* Remove stutter from the names of token variants

### DIFF
--- a/src/html5/tokenizer/token.rs
+++ b/src/html5/tokenizer/token.rs
@@ -19,37 +19,32 @@ pub struct Attribute {
 }
 
 /// The different token structures that can be emitted by the tokenizer
-#[derive(Clone, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum Token {
-    DocTypeToken {
+    DocType {
         name: Option<String>,
         force_quirks: bool,
         pub_identifier: Option<String>,
         sys_identifier: Option<String>,
     },
-    StartTagToken {
+    StartTag {
         name: String,
         is_self_closing: bool,
         attributes: HashMap<String, String>,
     },
-    EndTagToken {
+    EndTag {
         name: String,
         is_self_closing: bool,
-        attributes: HashMap<String, String>,
     },
-    CommentToken {
-        value: String,
-    },
-    TextToken {
-        value: String,
-    },
-    EofToken,
+    Comment(String),
+    Text(String),
+    Eof,
 }
 
 impl Token {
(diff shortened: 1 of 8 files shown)


src/html5/tokenizer/token.rs
```diff
@@ -19,37 +19,32 @@ pub struct Attribute {
...
+    StartTag {
         name: String,
         is_self_closing: bool,
         attributes: HashMap<String, String>,
     },
-    EndTagToken {
+    EndTag {
         name: String,
         is_self_closing: bool,
-        attributes: HashMap<String, String>,
```
emwalker, line 38: I don't think `attributes` is needed.  I suspect `is_self_closing` is also not needed, but I did not look into this question.

src/html5/tokenizer/token.rs
```diff
@@ -124,84 +119,49 @@ impl std::fmt::Display for Token {
...
-    #[test]
-    fn test_token_type() {
-        let token = Token::DocTypeToken {
-            name: None,
-            force_quirks: false,
-            pub_identifier: None,
-            sys_identifier: None,
-        };
-        assert_eq!(token.type_of(), TokenType::DocTypeToken);
-    }
```
emwalker, line 170: The trait, the enum and the `type_of` method have all been dropped in this PR.

src/testing/tokenizer.rs
```diff
@@ -387,3 +358,143 @@ pub fn fixtures() -> impl Iterator<Item = FixtureFile> {
         fixture_from_path(&path).ok()
     })
 }
+
+#[cfg(test)]
+mod tests {
```
emwalker, line 363: In addition to simplifying the names of the tokens, these unit tests, and the deserialization needed for them to work, are one of the main purposes of this PR.

src/html5/tokenizer/token.rs
jaytaph: in theory you could do:

```
</foo bar="baz" />
```

although i doubt this would actually result in anything.

jaytaph:
@emwalker Thanks for the PR.. I like these small PR's that does a single thing. Makes it a lot easier to check :-=)

PR #201, opened 2023-10-29
Added AUTHORS file and change copyright statements

jaytaph:
Added an AUTHORS file, which can be generated with `make authors` (we can add this to CI later to check if we are missing contributors).

Also updated the copyright in the license to "gosub contributors"


AUTHORS
```diff
@@ -0,0 +1,20 @@
+Ahmed Ibrahim <me@ahmedibrahim.dev>
+blaumeise20 <62756994+blaumeise20@users.noreply.github.com>
+CharlesChen0823 <yongchen0823@gmail.com>
+Dibash Thapa <dibashthapa55@gmail.com>
+Echo <echo-the-coder@tuta.io>
+Eric Walker <emwalker@users.noreply.github.com>
+Eric Walker <eric.walker@gmail.com>
```
emwalker, line 7: Can we drop the one with "eric.walker@...", since both of these people are me?

Cargo.toml
```diff
@@ -4,11 +4,11 @@ version = "0.1.0"
 edition = "2021"
 rust-version = "1.73"
 authors = ["Joshua Thijssen <jaytaph@deadcode.nl>"]
-description = "html5 browser engine"
+description = "A modern HTML5 browser engine written in Rust."
```
emwalker, line 7: I kind of want to avoid selling this project as being bigger than it is at this point.  Even if we strive for a "modern" brower, I think we're years away from that.  So I'd pitch for going with understatement here.

```suggestion
description = "An HTML5 browser engine written in Rust."
```

Makefile
```diff
@@ -37,6 +37,9 @@ test_commands:
 	cargo run --bin html5-parser-test >/dev/null
 	cargo run --bin parser_test >/dev/null
 
+authors: ## Update the authors file
+	git log --format='%aN <%aE>' | LC_ALL=C.UTF-8 sort -uf  | grep -v @localhost | grep -v 'dependabot\[bot\]' > AUTHORS
```
emwalker, line 41: I figured this was automated.  I wonder whether we can keep the updating of the file manual for now?  Just a thought.  People might not want their email addresses in such an easy place for bots to crawl for sending out spam.

Don't block on this question.  What you have is fine.  Just wanted to ask it.

emwalker (approved):
LGTM with questions

Cargo.toml
```diff
@@ -4,11 +4,11 @@ version = "0.1.0"
 edition = "2021"
 rust-version = "1.73"
 authors = ["Joshua Thijssen <jaytaph@deadcode.nl>"]
```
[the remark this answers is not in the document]
jaytaph: I was thinking about this one as well.. I don't think it is wise to keep a list of ALL contributors here, but if we only have a select few, then we have to think of a cut-off point which can be difficult. 

I see that servo does it a bit different:

https://github.com/servo/html5ever/blob/master/html5ever/Cargo.toml#L5C1-L5C49

So we can use this for now (maybe use info@gosub.io as a generic contact address)

AUTHORS
jaytaph: The list is generated from the commits in the git repo. The problem with removing is that it will be added again with a new "make authors". Unless you really have an issue with it, i would like to keep it

Cargo.toml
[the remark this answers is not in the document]
jaytaph: this was chatgpt generated :-)

Makefile
jaytaph: I wouldn't mind to change it to manual. It's a fair point not to have their email directly in the open.

I would suggest to manually create the AUTHORS file, and have some contribution page where people can add their email if they want to the AUTHORS list manually..

jaytaph: I've removed the automatic process

emwalker (approved):

PR #204, opened 2023-10-29
clean up document and node legacy code

Kiyoshika:
With the introduction of the new `TreeBuilder` interface this allows me to clean up some now "legacy" code that tried to keep nodes in sync with the DOM when inserting special attributes (such as ID)

This PR mainly focuses on removing the old methods for setting/getting nodes by named ID and replacing `ElementAttribute` struct with a regular hashmap. Originally this was intended to deal with the special id/class attributes to manipulate the DOM via handle, but this is no longer required with `TreeBuilder`. I don't know if we still need a pointer to document handle in the nodes but I will keep it there for now just in case.

I've also removed some unneeded/redundant tests related to named IDs as they are now mostly included with the `TreeBuilder` tests in the document.

The logic for `Document.add_new_node` has been updated slightly to properly handle updating `named_id_elements` when a new node contains attributes. Previously I wasn't actually tying a named ID to a new node properly, so when parsing HTML we wouldn't actually be able to fetch with `document.get_node_by_named_id`. This was fixed and test was updated in `parser.rs`

TLDR; cleaning up older code now that we have a better interface to work with DOM mutations


Kiyoshika:
Hm, actually I realized there is one slight flaw. I need to remove the old named ID if one is updated. Converting to draft for now until that's fixed. That's what the second assertion in the `div/p` test was..

Kiyoshika:
Nevermind to my previous comment, that doesn't matter when it's a NEW element, but I did notice this functionality is currently missing in the `TreeBuilder.insert_attribute` and also the validation is happening in the wrong spot... I think this PR for the sake of cleanup is good to go but will create an issue for myself to fix up some issues with `insert_attribute`

jaytaph (approved):

jaytaph:
It seems that I can't rebase manually from github, as my commits will not be signed in that case.. I'll rebase this manually and force-push and merge it.

jaytaph:
@Kiyoshika on the other hand, it might be better if you do this, since otherwise i'm not sure if we loose contribution credit this way.

PR #206, opened 2023-10-29
Continue revisiting tokenization stage

emwalker:
This PR continues the revisiting of the HTML5 tokenization:

* Deserialize the initial states directly into the `TestSpec` struct instead of capturing them after the fact
* Remove stutter from the names of the states, sort them alphabetically, and use the [all-caps name](https://www.w3.org/TR/2011/WD-html5-20110113/tokenization.html)  when applicable


src/testing/tokenizer.rs
```diff
@@ -61,12 +61,34 @@ pub struct TestSpec {
     pub errors: Vec<TokenError>,
     #[serde(default)]
     pub double_escaped: bool,
-    #[serde(default)]
-    pub initial_states: Vec<String>,
+    #[serde(default, deserialize_with = "deserialize_states")]
+    pub initial_states: Vec<TokenState>,
```
emwalker, line 65: This is the main change in this PR.

Sharktheone (approved):

src/html5/tokenizer/state.rs
```diff
@@ -1,78 +1,78 @@
...
-    AfterDocTypeSystemKeywordState,
-    BeforeDocTypeSystemIdentifierState,
-    DocTypeSystemIdentifierDoubleQuotedState,
-    DocTypeSystemIdentifierSingleQuotedState,
-    AfterDocTypeSystemIdentifierState,
-    BogusDocTypeState,
-    CDataSectionState,
-    CDataSectionBracketState,
-    CDataSectionEndState,
+    AfterAttributeName,
```
jaytaph, line 4: The reason this was not alphabetical is that it matches the states as defined in the specification (https://html.spec.whatwg.org).

I don't mind changing this to alphabetical though.

jaytaph (approved):

src/html5/tokenizer/state.rs
emwalker: Might just be me personally, but I think it will be easier to find the states if they're alphabetized.  (I was thinking of adding comments for the sections in a later PR.)

PR #209, opened 2023-10-30
Continue revisiting tokenization stage

emwalker:
This PR continues the revisiting of the tokenization code:
* Include stronger type checking in the matches instead of losing type info by casting to a `char` or `u32`

### DIFF
--- a/src/html5/input_stream.rs
+++ b/src/html5/input_stream.rs
@@ -50,58 +50,46 @@ impl fmt::Display for Position {
 /// a surrogate characters since these cannot be stored in a single char.
 /// Eof is denoted as a separate element.
 #[derive(PartialEq, Debug, Copy, Clone)]
-pub enum Element {
+pub enum Bytes {
     /// Standard UTF character
-    Utf8(char),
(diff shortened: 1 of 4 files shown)


src/html5/input_stream.rs
```diff
@@ -50,58 +50,46 @@ impl fmt::Display for Position {
...
-pub enum Element {
+pub enum Bytes {
     /// Standard UTF character
-    Utf8(char),
+    Ch(char),
     /// Surrogate character (since they cannot be stored in char)
     Surrogate(u16),
     /// End of stream
     Eof,
 }
```
emwalker, line 60: The name of this struct will change from `Bytes` to something else in a later PR, once we deal with surrogate characters at an earlier stage of the input stream.

emwalker: The name `Ch` was chosen (1) because it's short and (2) because the Rust name for a UTF-8 character is `char`.

emwalker:
Merging now, since it touches so many places and so is likely to conflict with other work in flight.

PR #211, opened 2023-10-31
Continue to revisit the tokenization code

emwalker:
This PR
* Adds comments to the `State` enum variants pointing to sections of the [HTML5 tokenization spec](https://www.w3.org/TR/2011/WD-html5-20110113/tokenization.html) when a section is available
* Brings some of the state names closer in line with what is seen in the spec

### DIFF
--- a/src/html5/tokenizer/state.rs
+++ b/src/html5/tokenizer/state.rs
@@ -1,78 +1,214 @@
 /// These are the states in which the tokenizer can be in.
 #[derive(Debug, Copy, Clone, PartialEq)]
 pub enum State {
+    /// 8.2.4.36 After attribute name state
     AfterAttributeName,
+
+    /// 8.2.4.42 After attribute value (quoted) state
     AfterAttributeValueQuoted,
-    AfterDocTypeName,
-    AfterDoctypePublicIdentifier,
-    AfterDocTypePublicKeyword,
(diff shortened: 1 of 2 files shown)


src/html5/tokenizer/state.rs
```diff
@@ -1,78 +1,212 @@
...
-    BogusDocType,
+
+    /// 8.2.4.67 Bogus DOCTYPE state
+    BogusDOCTYPE,
+
+    /// 8.2.4.68 CDATA section state
     CDATASection,
+
     CDATASectionBracket,
     CDATASectionEnd,
```
emwalker, line 63: I did not find sections for these states.

src/html5/tokenizer/state.rs
```diff
@@ -1,78 +1,212 @@
...
+    /// 8.2.4.51 Comment end bang state
     CommentEndBang,
+
+    /// 8.2.4.49 Comment end dash state
     CommentEndDash,
+
     CommentLessThanSign,
     CommentLessThanSignBang,
     CommentLessThanSignBangDash,
     CommentLessThanSignBangDashDash,
```
emwalker, line 89: I did not find sections for these states.

jaytaph (approved):

PR #213, opened 2023-10-31
Added documentBuilder struct and implemented fragment cases

jaytaph:
This PR adds the document builder (already merged into main), and the handling of fragment cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # GoSub: Gateway to Optimized Searching and Unlimited Browsing
 
-This repository holds the GoSub HTML5 parser/tokenizer. It is a standalone library that can be used by other projects but will ultimately be used by the GoSub browser. See the [About](#about) section for more information.
+This repository holds the GoSub HTML5 parser/tokenizer. It is a standalone library that can be used by other projects
+but will ultimately be used by the GoSub browser. See the [About](#about) section for more information.
 
 ```
                        _     
@@ -14,29 +15,32 @@ This repository holds the GoSub HTML5 parser/tokenizer. It is a standalone libra
          Unlimited Browsing                    
 ```
 
-
 ## About
 
-This repository is part of the GoSub browser project. Currently, there is only a single component/repository (this one), but the idea will be that there are many other components that, as a whole, make up a full-fledged browser. Each of the components can probably function as something standalone (ie, html5 parser, CSS parser, etc.).
+This repository is part of the GoSub browser project. Currently, there is only a single component/repository (this one),
+but the idea will be that there are many other components that, as a whole, make up a full-fledged browser. Each of the
+components can probably function as something standalone (ie, html5 parser, CSS parser, etc.).
 
-In the future, this component (HTML5 parser) will receive a stream of bytes through an API and output a stream of events. The next component will consume the events, and so on, until we can display something in a window/user agent. This could be a text-mode browser, but the idea is to have a graphical browser.
-
+In the future, this component (HTML5 parser) will receive a stream of bytes through an API and output a stream of
+events. The next component will consume the events, and so on, until we can display something in a window/user agent.
+This could be a text-mode browser, but the idea is to have a graphical browser.
 
 ## Status
 
 > This project is in its infancy. There is no browser you can use yet.
 
 This is a work in progress. The current status is that the parser can parse a few HTML5 documents, but it is far from
-ready. The main goal is to be able to parse correctly all the tests in the html5lib-tests repository 
-(https://github.com/html5lib/html5lib-tests). 
+ready. The main goal is to be able to parse correctly all the tests in the html5lib-tests repository
+(https://github.com/html5lib/html5lib-tests).
 
-Our goal at the moment is to research as much as possible and to setup proof-of-concepts in order to gain more 
+Our goal at the moment is to research as much as possible and to setup proof-of-concepts in order to gain more
 understanding in the field of browsers. We are not trying to create a full-fledged browser at the moment, but it will
-be our ultimate goal. 
+be our ultimate goal.
 
 ## How to build
 
-This project uses [cargo](https://doc.rust-lang.org/cargo/) and [rustup](https://www.rust-lang.org/tools/install).  First you must install `rustup` at the link provided.  After installing `rustup`, run:
+This project uses [cargo](https://doc.rust-lang.org/cargo/) and [rustup](https://www.rust-lang.org/tools/install). First
+you must install `rustup` at the link provided. After installing `rustup`, run:
 
 ```bash
 $ rustup toolchain install 1.73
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -1,6 +1,7 @@
 # GoSub: Gateway to Optimized Searching and Unlimited Browsing
 
-This repository holds the GoSub HTML5 parser/tokenizer. It is a standalone library that can be used by other projects but will ultimately be used by the GoSub browser. See the [About](#about) section for more information.
+This repository holds the GoSub HTML5 parser/tokenizer. It is a standalone library that can be used by other projects
```
jaytaph, line 3: Seems like my editor wraps with newlines.

src/bin/gosub-parser.rs
```diff
@@ -42,8 +42,7 @@ fn main() -> Result<()> {
     }
 
     let document = DocumentBuilder::new_document();
-    let mut parser = Html5Parser::new(&mut stream, document.clone());
-    let parse_errors = parser.parse()?;
+    let parse_errors = Html5Parser::parse_document(&mut stream, Document::clone(&document), None)?;
```
jaytaph, line 45: We don't call a "new()" method first to initialize a parser, but directly call the `parse_document()` or `parse_document_fragment()` instead.  This is due the fact that we need to initialize the parser, but we cannot do this until we actually call the parser (otherwise we have to specify things with `option<>`, which doesn't sound right either.

src/html5/parser/document.rs
```diff
@@ -695,9 +696,23 @@ impl DocumentBuilder {
     }
 
     /// Creates a new document fragment with the context as the root node
-    pub fn new_document_fragment(_doc: DocumentHandle, _context: NodeId) -> DocumentHandle {
-        // @TODO: Create a document fragment and set doc/context as the root node
-        todo!();
+    pub fn new_document_fragment(context: Node) -> DocumentHandle {
```
jaytaph, line 699: The document builder can either return a `document` or a `document_fragment`. A document is just a simple document with a document root element. A fragment uses the `context` as a node and uses `html` as a root element.

src/html5/parser/quirks.rs
```diff
@@ -156,15 +156,13 @@ static LIMITED_QUIRKS_PUB_IDENTIFIER_PREFIX_NOT_MISSING_SYS: &[&str] = &[
...
-    use crate::html5::parser::document::DocumentBuilder;
     use crate::html5::parser::Html5Parser;
     use crate::html5::parser::QuirksMode;
 
     #[test]
     fn test_quirks_mode() {
-        let doc = DocumentBuilder::new_document();
-        let mut stream = InputStream::new();
-        let parser = Html5Parser::new(&mut stream, doc.clone());
+        let stream = &mut InputStream::new();
```
jaytaph, line 164: During testing, we need to have a parser instance that doesn't care about the stream and document it contains. In these cases we can use `Html5Parser::new_parser()`. I doubt this function is called anywhere but in unit tests.

src/html5/tokenizer.rs
```diff
@@ -109,6 +118,11 @@ impl<'stream> Tokenizer<'stream> {
         self.error_logger.borrow()
     }
 
+    /// Sets the tokenizer state to a new state
```
jaytaph, line 121: When parsing a fragment, the tokenizer must be set to a specific state depending on the (context) node the fragment. We use a public `set_state` for this.

src/testing/tree_construction.rs
```diff
@@ -206,22 +208,67 @@ impl Test {
...
     /// Run the parser and return the document and errors
     pub fn parse(&self, scripting_enabled: bool) -> Result<(DocumentHandle, Vec<ParseError>)> {
-        let document = DocumentBuilder::new_document();
+        // let mut is_fragment = false;
+        let mut context_node = None;
+        let document;
+
+        let is_fragment;
+
+        if let Some(fragment) = self.spec.document_fragment.clone() {
```
jaytaph, line 217: This function now decides whether or not we are parsing a document or document-fragment based on the test specification. It will create a dummy main document with the context node, and then create the documen-fragment which gets filled by the parser.

src/testing/tree_construction.rs
```diff
@@ -206,22 +208,67 @@ impl Test {
...
+            Html5Parser::parse_document(&mut is, Document::clone(&document), Some(options))?
+        };
 
         Ok((document, parse_errors))
     }
 
     /// Returns true if the whole document tree matches the expected result
     pub fn match_document_tree(&self, document: &Document) -> SubtreeResult {
-        self.match_node(NodeId::root(), 0, -1, document)
+        if self.spec.document_fragment.is_some() {
```
jaytaph, line 267: If we have a fragment, the first element is the "html" node. This node is not present in the test output, so we skip this node by matching the tree from the second node instead of the root node.

src/testing/tree_construction/parser.rs
```diff
@@ -262,7 +262,7 @@ fn document(i: Span) -> IResult<Span, Span> {
 }
 
 fn document_fragment(i: Span) -> IResult<Span, Span> {
-    preceded(tag("#document-fragment\n"), take_until1("#"))(i)
+    preceded(tag("#document-fragment\n"), take_until1("\n"))(i)
```
jaytaph, line 265: It seems that we have to capture until the newline instead of the next line's `#`.

I thought this would mess up tests, but it turns out this works fine.

src/testing/tree_construction/parser.rs
```diff
@@ -635,7 +635,7 @@ template
 "#,
         );
 
-        assert_eq!(test.document_fragment, Some("template\n".into()));
+        assert_eq!(test.document_fragment, Some("template".into()));
```
jaytaph, line 638: Small fix for the \n that we do not capture (see above)

README.md
emwalker: GitHub doesn't render markdown properly when newlines are introduced.  For Markdown (only), you want lines to wrap.

src/bin/gosub-parser.rs
emwalker: I'd ultimately switch to something that looks a little like this:
```rust
// Parse a whole document
let result = gosub::html5::parse(&mut stream, Document::clone(&document), None)?;

// Parse a fragment
let result = gosub::html5::parse_fragment(&mut stream, Document::clone(&document) /*, ... */)?;
```

Since we already have the `html5` module, `Html5Parser` adds stutter, and `html::Parser` would be simpler.  But in this case, we can just expose two functions, `html5::parse` and `html::parse_fragment`, that delegate to a `Parser` without making it part of the public interface except in more advanced use cases.

src/html5/parser.rs
```diff
@@ -140,6 +140,18 @@ impl ActiveElement {
     }
 }
 
+pub struct Html5ParserOptions {
+    pub scripting_enabled: bool,
+}
```
emwalker, line 145: See my comments above about stutter — this might look better as `html5::ParserOptions`.

src/html5/parser/document.rs
```diff
@@ -695,9 +696,23 @@ impl DocumentBuilder {
...
+        let mut doc = Document::shared();
+        doc.get_mut().doctype = DocumentType::HTML;
```
emwalker, line 701: I like the shorter `doc` variable name for cases like this one, where we're not referencing a field on a struct.

src/html5/parser/quirks.rs
```diff
@@ -156,15 +156,13 @@ static LIMITED_QUIRKS_PUB_IDENTIFIER_PREFIX_NOT_MISSING_SYS: &[&str] = &[
...
+        let parser = Html5Parser::new_parser(stream);
```
emwalker, line 165: Since it's returning a parser, how about just `::new(stream)`, or `::from_stream(stream)`?

src/html5/tokenizer.rs
emwalker: I can see this changing when I start to fiddle with the state machine.  Hopefully this detail is not set in stone.

emwalker (approved):
Most of this PR looks fine.  The only thing I'd want to fix for sure is the newlines in the markdown, which GitHub won't display properly.

src/html5/tokenizer.rs
jaytaph: nothing is.. but the parser NEEDS a way to set the tokenizer state. This is important, as the document fragment tokenizing does not start in the default (data) state.

src/html5/parser/quirks.rs
jaytaph: i think this makes sense  (actually, go formatting will complain about stuttering in package  / module names).

However, i want to keep this for now away from this PR, and do a next run when cleaning up things like this

README.md
jaytaph: I'll update the pr to wrap them.

PR #214, opened 2023-10-31
test(html): add html5 tests as git subtree

### LINKED ISSUE #187: Use git-subtree for vendoring html5lib-tests
Instead of copying the html5lib-tests inside our repository, it might be better to use a git subtree.

shubhexists:
Fixes #187


jaytaph:
@shubhexists ah.. still a small issue of signing your commits :-) Check out the link [Learn more about signing commits.](https://docs.github.com/articles/about-gpg/)

emwalker (approved):

shubhexists:
So is there a way to sign already made commits?🙂

blaumeise20:
I don't think so, no. You will have to rebase your branch.

jaytaph:
@shubhexists there is a way through rebasing

- make sure git does verified commits.
- do a `git rebase -i fd51b80~1`
- you get a list of commits you want to rebase. They all start with the word "pick"
- save the file and exit. That will renegerate your commits
- do a `git push origin main -f`.  That will push your branch to your github and "overwrites" the current main branch that is there (since you just have rewritten it, you must use the -f flag).
- that should be it :-)

emwalker:
@shubhexists almost there!  It looks like the next hurdle is that all of your commits will need to be signed.



This can be done by rebasing.  Let us know if this sounds hard, and we might be able to figure something else out.

shubhexists:
I am sorry, I forgot to add a comment. 
Actually there's an issue. 
According to link , hat @jaytaph , I have confirgured my system to automatically sign every commit. I tried making commits in other repositories and I am getting that "Verified" mark.
But the issue is, when I run the `git subtree add` command,  There isn't any commit. I just direclty need to push the after the command.  So there is no chance to sign that : /

Now, I tried git rebase in editing mode as mentioned above, the issue is that after running the git rebase command, all the files of the subtree are added to the root directory of gosub-engine ( which shouldn't be the case. It should be added to data/ directory).
I'll give it another try after some time and update you guys on Slack.

jaytaph:
@shubhexists @emwalker I've got the same issue when I tried my test project:

image

It indeed seems that subtree commits are not signed.. I have to check why this is

jaytaph:
@shubhexists @emwalker After some discussion we have decided we leave the subtrees out of the system for now, until we come up with a better way to deal with them.

PR #215, opened 2023-11-01
Added initial example

jaytaph:
Added an example which can be automatically ran by using:

```
$ cargo run --example html5-parser
```

### DIFF
--- a/examples/html5-parser.rs
+++ b/examples/html5-parser.rs
@@ -1,0 +1,18 @@
+use gosub_engine::html5::parser::document::{Document, DocumentBuilder};
+use gosub_engine::html5::{
+    input_stream::{Encoding, InputStream},
+    parser::Html5Parser,
+};
+
+fn main() {
+    // Creates an input stream
+    let mut stream = InputStream::new();
+    stream.read_from_str("<p>Hello<b>world</b></p>", Some(Encoding::UTF8));
+
+    // Initialize a document and feed it together with the stream to the html5 parser
+    let document = DocumentBuilder::new_document();
+    let _ = Html5Parser::parse_document(&mut stream, Document::clone(&document), None);
+
+    // document now contains the html5 node tree
+    println!("Generated tree: \n\n {}", document);
+}


examples/html5-parser.rs
```diff
@@ -0,0 +1,18 @@
...
+};
+
+fn main() {
+    // Creates an input stream
+    let mut stream = InputStream::new();
+    stream.read_from_str("<p>Hello<b>world</b></p>", Some(Encoding::UTF8));
+
+    // Initialize a document and feed it together with the stream to the html5 parser
+    let document = DocumentBuilder::new_document();
+    let _ = Html5Parser::parse_document(&mut stream, Document::clone(&document), None);
```
blaumeise20, line 14: Can't you say `document.clone()` here?
```suggestion
    let _ = Html5Parser::parse_document(&mut stream, document.clone(), None);
```

jaytaph: I think you can.. but @emwalker had something to say about that.. (i forgot what :p)

Sharktheone: It is because it is a shallow copy and not a deep copy, which means, the heap memory doesn't get copied but reused.

jaytaph: would it make sense to somehow change the naming to reflect this? Or is this the common idiomatic approach?

Sharktheone: Yes, i think this initially comes from the Rc or Arc smart pointers. For those smart pointers it is common to use ::clone() and not .clone() but not for the String smart pointer somehow🤷. I'm not completely sure, how the Document type is defined in the code, but from the suggestion from Eric, it is probably some kind of smart pointer.

emwalker: The convention with `Rc` is to use `Rc::clone(&reference)` instead of `reference.clone()`.  It's purely a developer signal.  As @Sharktheone mentioned, it's to let the developer know that a shallow copy is being made rather than a deep copy.

That said, @blaumeise20 likely knows more about Rust than I do, so if he reads this and still thinks `document.clone()` is preferable, we should go with what he says.

examples/html5-parser.rs
```diff
@@ -0,0 +1,18 @@
...
+
+    // document now contains the html5 node tree
+    println!("Generated tree: \n\n {}", document);
+}
```
emwalker, line 18: Might be nice to add simple error handling:

```rust
use gosub_engine::html5::{
    input_stream::{Encoding, InputStream},
    parser::document::{Document, DocumentBuilder},
    parser::Html5Parser,
};
use gosub_engine::types::Result;

fn main() -> Result<()> {
    // Creates an input stream
    let mut stream = InputStream::new();
    stream.read_from_str("<p>Hello<b>world</b></p>", Some(Encoding::UTF8));

    // Initialize a document and feed it together with the stream to the html5 parser
    let document = DocumentBuilder::new_document();
    let _ = Html5Parser::parse_document(&mut stream, Document::clone(&document), None)?;

    // document now contains the html5 node tree
    println!("Generated tree: \n\n {}", document);
    Ok(())
}
```

emwalker (approved):
LGTM with suggestion.

Kiyoshika (approved):

examples/html5-parser.rs
blaumeise20: Oh I understand now why you did this, although the reason for this is that is that you often have something *inside* the `Rc` which is handled like there was no `Rc` around it, and thus it makes it clearer that the `Rc` itself is cloned, not the content.

From the docs:
> `Rc<T>`'s implementations of traits like `Clone` may also be called using fully qualified syntax. Some people prefer to use fully qualified syntax, while others prefer using method-call syntax.

In this case, since `DocumentHandle` is not a wrapper type but explicitly a handle to `Document`, it make more sense to me to use the function call. But I can totally understand your point of the shallow copy. Do whatever you want ;-)

examples/html5-parser.rs
jaytaph: Agreed, however this example should not fail as it's not bound to any runtime content (like files etc)

examples/html5-parser.rs
emwalker: I'm just a Rust newbie, still learning the conventions.

PR #216, opened 2023-11-01
Implemented doctype nodes

jaytaph:
This PR implements a new node type named "doctype". It is used for the `<!doctype>`  tags which works a bit differently than regular tags.

The common tree structure would be:

```
|- <document node (0)>
   |- <doctype node (1)>
   |- <element node "html" (2)>
   |  |- <element node "head" (3)>
   |  |   |- <element node "title" (4)>
   |- <element node "body" (5)>
```

Note that the document node (0) is the root node. We use a separate document node type for this, but this isn't really used anywhere except for holding a html tree.

The "document" which we use and create with `DocumentBuilder::new_document()` is a the structure that holds the node arena and other things. This may be a bit confusing in naming, especially when we will get to the "DOM" document (which is an interface that interacts with our document-tree).

Anyway, for now.. the doctype works, and we are at 92% pass rate.

### DIFF
--- a/src/html5/parser.rs
+++ b/src/html5/parser.rs
@@ -1751,11 +1751,17 @@ impl<'stream> Html5Parser<'stream> {
     /// Create a new node that is not connected or attached to the document arena
     fn create_node(&self, token: &Token, namespace: &str) -> Node {
         match token {
-            Token::DocType { name, .. } => {
-                let val = format!("!DOCTYPE {}", name.as_deref().unwrap_or(""),);
-
-                return Node::new_element(&self.document, val.as_str(), HashMap::new(), namespace);
-            }
+            Token::DocType {
+                name,
+                pub_identifier,
+                sys_identifier,
+                ..
+            } => Node::new_doctype(
+                &self.document,
+                &name.clone().unwrap_or_default(),
+                &pub_identifier.clone().unwrap_or_default(),
+                &sys_identifier.clone().unwrap_or_default(),
+            ),
             Token::StartTag {
                 name, attributes, ..
             } => Node::new_element(&self.document, name, attributes.clone(), namespace),
--- a/src/html5/parser/document.rs
+++ b/src/html5/parser/document.rs
@@ -1,4 +1,5 @@
 use crate::html5::node::arena::NodeArena;
+use crate::html5::node::data::doctype::DocTypeData;
 use crate::html5::node::data::{comment::CommentData, text::TextData};
 use crate::html5::node::HTML_NAMESPACE;
 use crate::html5::node::{Node, NodeData, NodeId};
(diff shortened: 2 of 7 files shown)


src/html5/parser/document.rs
```diff
@@ -494,6 +495,17 @@ impl Document {
             NodeData::Document(_) => {
                 _ = writeln!(f, "{}Document", buffer);
             }
+            NodeData::DocType(DocTypeData {
```
jaytaph, line 498: I want to get rid of this... here we generate the nodes as strings, but we also do this when generating the trees during testing, which does a lot of the same work

src/html5/parser.rs
```diff
@@ -1751,11 +1751,17 @@ impl<'stream> Html5Parser<'stream> {
...
+            Token::DocType {
+                name,
+                pub_identifier,
+                sys_identifier,
+                ..
+            } => Node::new_doctype(
+                &self.document,
+                &name.clone().unwrap_or("".into()),
+                &pub_identifier.clone().unwrap_or("".into()),
+                &sys_identifier.clone().unwrap_or("".into()),
```
emwalker, line 1763:
```suggestion
                &name.clone().unwrap_or_default(),
                &pub_identifier.clone().unwrap_or_default(),
                &sys_identifier.clone().unwrap_or_default(),
```

src/html5/parser/document.rs
```diff
@@ -494,6 +495,17 @@ impl Document {
...
+                name,
+                pub_identifier,
+                sys_identifier,
+            }) => {
+                _ = writeln!(
+                    f,
+                    "{}<!DOCTYPE {} \"{}\" \"{}\">",
+                    buffer, name, pub_identifier, sys_identifier
```
emwalker, line 506:
```suggestion
                    r#"{buffer}<!DOCTYPE {name} "{pub_identifier}" "{sys_identifier}">"#,
```

src/testing/tree_construction.rs
```diff
@@ -283,6 +299,43 @@ impl Test {
...
+                    if doctype.pub_identifier.is_empty() && doctype.sys_identifier.is_empty() {
+                        // <!DOCTYPE html>
+                        doctype.name.to_string()
+                    } else {
+                        // <!DOCTYPE html "pubid" "sysid">
+                        format!(
+                            "{} \"{}\" \"{}\"",
+                            doctype.name, doctype.pub_identifier, doctype.sys_identifier
+                        )
+                    };
```
emwalker, line 313:
```suggestion
            NodeData::DocType(DocTypeData {
                name,
                pub_identifier,
                sys_identifier,
            }) => {
                let doctype_text = if pub_identifier.is_empty() && sys_identifier.is_empty() {
                    // <!DOCTYPE html>
                    name.to_string()
                } else {
                    // <!DOCTYPE html "pubid" "sysid">
                    format!(r#"{name} "{pub_identifier}" "{sys_identifier}""#,)
                };
```

emwalker (approved):
LGTM with suggestions

Kiyoshika:
Would it make sense to repurpose the existing `Document` node type for this? That's actually what I originally thought it was, I didn't realize it was just a placeholder.

Maybe when creating a new document we have an empty root until we find a `doctype` token (or force insert a default if not found)

If it's more work than what it's worth, I don't really mind going for this, but just seems a little odd to have both a `Document` and `DocType` type

jaytaph:
The initial setup was indeed to have a document as a first class node but it turns out we dont really use it. However, there will be some settings (like the quirks mode etc,) that we want to keep in the document because the DOM document interface exposes these properties. I can image that at that point, the document node we have becomes more important

PR #217, opened 2023-11-02
Continue revisiting tokenization

emwalker:
Soon we'll need to think clearly about the different stages of a tokenization pipeline.  This PR begins to differentiate the stages that will be needed by
* Renaming `InputStream` to `CharIterator`
* Using "chars" in place of the more generic "stream", which will soon be too generic
* Moving `src/html5/input_stream.rs` to `src/bytes.rs`, which will become its own module, since it's shared between HTML5 and CSS3 (and probably other formats)

While we're here, let's drop the unused `TokenType` enum.

### DIFF
--- a/src/bin/gosub-parser.rs
+++ b/src/bin/gosub-parser.rs
@@ -1,8 +1,8 @@
 use anyhow::Result;
 use gosub_engine::html5::parser::document::{Document, DocumentBuilder};
-use gosub_engine::html5::{
-    input_stream::{Confidence, Encoding, InputStream},
-    parser::Html5Parser,
+use gosub_engine::{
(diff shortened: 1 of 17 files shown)


jaytaph (approved):

jaytaph:
@emwalker I assume it would be fairly trivial to add something like  `io::stdin()` into the charIterator?

emwalker:
> @emwalker I assume it would be fairly trivial to add something like io::stdin() into the charIterator?

`CharIterator` is transitional.  I expect it to change (including the name) at some point.  But I'm assuming once things are in place, it will be straightforward to have `io::stdin()`, a socket, a file, etc., as the input to `CharIterator`.

Issue #205, opened 2023-10-29
`TreeBuilder.insert_attribute` issues


Kiyoshika:
While going through some code cleanup, I noticed a couple issues with the current `insert_attribute` implementation..

1. The attribute value validation is happening in the wrong spot. Below are the first few lines of the method
```rust
/// Inserts an attribute to an element node.                                                                                       
/// If node is not an element or if passing an invalid attribute value, returns an Err()                                           
fn insert_attribute(&mut self, key: &str, value: &str, element_id: NodeId) -> Result<()> {                                         
    if !self.get().validate_id_attribute_value(value) {                                                                            
        return Err(Error::DocumentTask(format!(                                                                                    
            "Attribute value '{}' did not pass validation",                                                                        
            value                                                                                                                  
        )));                                                                                                                       
    } 
// ...
```
It's validating ALL attribute values like they are ID attributes, which may not be correct...

2. When inserting an ID attribute into the DOM tree, it currently does not check if the current `element_id` has an ID attribute which would need to be removed from the DOM tree (to prevent a dangling ID reference)

Ideally, if we have `<div id="myid">`, the ID `myid` is tied to this node ID. But if this changes (e.g., through JavaScript) to something like `otherid`, the old `myid` is still pointing to this node when it should be removed.

One thing to be careful about is the elements attributes are updated *before* modifying the DOM. I think the order of these operations need to switch. Below is current implementation:

```rust
 if let Some(node) = self.get_mut().get_node_by_id_mut(element_id) {                                                            
      if let NodeData::Element(element) = &mut node.data {
          // TODO: don't update the element's attributes until AFTER making changes in DOM.                                                                    
          element.attributes.insert(key.to_owned(), value.to_owned());                                                           
      } else {                                                                                                                   
          return Err(Error::DocumentTask(format!(                                                                                
              "Node ID {} is not an element",                                                                                    
              element_id                                                                                                         
          )));                                                                                                                   
      }                                                                                                                          
  } else {                                                                                                                       
      return Err(Error::DocumentTask(format!(                                                                                    
          "Node ID {} not found",                                                                                                
          element_id                                                                                                             
      )));                                                                                                                       
  }                                                                                                                              
                                                                                                                                 
  // special cases that need to sync with DOM                                                                                    
  match key {                                                                                                                    
      "id" => {
          // if ID is already in use, ignore
          if !self.get().named_id_elements.contains_key(value) {
[...]

PR #221, opened 2023-11-03
(#205) fix issues with insert_attribute

Kiyoshika:
Fixing some logical issues according to #205 

Thanks to @emwalker for cleaning up the `insert_id_attribute` implementation.

There's still some follow up work (adding support for `class` and inserting generic attributes) but those will be separate issues - all current tests are passing for now.

### DIFF
--- a/src/html5/parser/document.rs
+++ b/src/html5/parser/document.rs
@@ -5,6 +5,7 @@ use crate::html5::node::HTML_NAMESPACE;
 use crate::html5::node::{Node, NodeData, NodeId};
 use crate::html5::parser::quirks::QuirksMode;
 use crate::html5::parser::tree_builder::TreeBuilder;
+use crate::html5::util::is_valid_id_attribute_value;
 use crate::types::{Error, Result};
 use alloc::rc::Rc;
 use core::fmt;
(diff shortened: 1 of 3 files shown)


Sharktheone (approved):
Looks fine to me

jaytaph (approved):

src/html5/parser/document.rs
```diff
@@ -618,6 +603,54 @@ impl DocumentHandle {
...
+            )))?
+            .data;
+
+        if let NodeData::Element(_) = data {
+        } else {
+            return Err(Error::DocumentTask(format!(
+                "Node ID {} is not an element",
+                element_id
+            )));
+        }
```
emwalker, line 638: In a later PR, we can consolidate lines 634-637 with line 646, so that there's a single `if let` statement.

Kiyoshika: This is what I was thinking but wasn't sure if a return statement was allowed there since we were assigning to a variable (I didn't attempt it, though.) If that's possible then I can switch it

emwalker: I just checked, and it looks like Rust will allow it.

PR #224, opened 2023-11-04
add support for class and generic attributes

Kiyoshika:
This PR is an extension of my last one where I add support for the `class` attribute as well as generic attributes (anything other than `id` and `class`)

### DIFF
--- a/src/html5/parser/document.rs
+++ b/src/html5/parser/document.rs
@@ -1,3 +1,4 @@
+use crate::html5::element_class::ElementClass;
 use crate::html5::node::arena::NodeArena;
 use crate::html5::node::data::doctype::DocTypeData;
 use crate::html5::node::data::{comment::CommentData, text::TextData};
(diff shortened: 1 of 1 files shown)


src/html5/parser/document.rs
```diff
@@ -683,11 +724,8 @@ impl TreeBuilder for DocumentHandle {
     fn insert_attribute(&mut self, key: &str, value: &str, element_id: NodeId) -> Result<()> {
```
Kiyoshika, line 724: Ideally I'd like to fetch the node and do the existence/type checks here and pass the node to the other insert methods by reference but couldn't figure out how to properly do it without the borrow checker yelling at me so there's a little bit of code dupe. Open to suggestions

src/html5/parser/document.rs
```diff
@@ -1094,4 +1132,50 @@ mod tests {
...
+        let NodeData::Element(element) = &doc_read.get_node_by_id(div_id).unwrap().data else {
+            panic!()
+        };
+        assert!(element.classes.contains("one"));
+        assert!(element.classes.contains("two"));
+        assert!(element.classes.contains("three"));
+    }
+
+    #[test]
+    fn task_queue_insert_class_attribute() {}
```
Kiyoshika, line 1180: whoops, forgot to implement this test... will do now

jaytaph (approved):

PR #232, opened 2023-11-05
Parser data

jaytaph:


src/html5/parser.rs
```diff
@@ -3721,13 +3752,27 @@ impl<'chars> Html5Parser<'chars> {
         }
     }
 
+    fn parser_data(&self) -> ParserData {
+        let namespace = self
+            .get_adjusted_current_node()
```
CharlesChen0823, line 3757: this implemented is not properly, the test case html5test-com.dat case 14 failed is because this.

src/testing/tree_construction/parser.rs
```diff
@@ -296,7 +297,7 @@ fn test(i: Span) -> IResult<Span, TestSpec> {
 
             TestSpec {
                 position,
-                data: data.to_string(),
+                data: data.to_string().trim_matches(|c| c == '\n').to_string(),
```
CharlesChen0823, line 300: data should only trim one '\n', but trim_matches will trim all '\n', this should modify. tests16.dat and tests26.dat remain failed case is because this.

src/html5/parser/helper.rs
```diff
@@ -90,12 +90,7 @@ impl Html5Parser<'_> {
...
 
-    pub fn insert_text_helper(
-        &mut self,
-        node: NodeId,
-        position: InsertionPositionMode<NodeId>,
-        token: &Token,
-    ) {
+    pub fn insert_text_helper(&mut self, position: InsertionPositionMode<NodeId>, token: &Token) {
         match position {
             InsertionPositionMode::Sibling {
```
CharlesChen0823, line 95: in fn appropriate_place_insert also missing node.is_namespace(HTML_NAMESPAC)

src/testing/tree_construction/generator.rs
```diff
@@ -0,0 +1,110 @@
...
+                            "  ".repeat(indent_level),
+                            attr.0,
+                            attr.1
+                        ));
+                    }
+                }
+            }
+
+            // Template tags have an extra "content" node in the test tree ouput
+            if node.name == "template" {
```
CharlesChen0823, line 55: should add `&& node.is_namespace(HTML_NAMESPACE)`

CharlesChen0823:
IMO, after modify according my review comments, except scripted test cases, all have done.

jaytaph:
@CharlesChen0823 I'll add your suggestions to this pr.

src/testing/tree_construction/parser.rs
jaytaph: +1

src/html5/parser.rs
jaytaph: 👍

src/testing/tree_construction/generator.rs
jaytaph: fixed

src/html5/parser/helper.rs
jaytaph: fixed

jaytaph:
@CharlesChen0823 All your suggestions have been added and we have now a 99.89% pass rate. There are 4 tests that cannot pass because they require scripting to be implemented which we do not have yet.

Many thanks for the help on the parser and test fixes. Hope to have your input on other parts of the gosub browser.

benches/tree_construction.rs
```diff
@@ -1,20 +1,24 @@
...
+
+    let mut harness = Harness::new();
 
     group.bench_function("fixtures", |b| {
         b.iter(|| {
             for root in &fixtures {
                 for test in &root.tests {
                     for &scripting_enabled in test.script_modes() {
-                        let _ = test.parse(scripting_enabled).unwrap();
+                        let _ = harness.run_test(test.clone(), scripting_enabled);
```
emwalker, line 21: The clone here and in the other test harness shouldn't be necessary, as we can pass around a reference without a lifetime.  But the clone is also harmless in this case.

src/html5/tokenizer.rs
```diff
@@ -50,6 +51,21 @@ pub struct Tokenizer<'stream> {
     pub error_logger: Rc<RefCell<ErrorLogger>>,
 }
 
+/// This struct is a gateway between the parser and the tokenizer. It holds data that can be needed
+/// by the tokenizer in certain cases. See https://github.com/gosub-browser/gosub-engine/issues/230 for
+/// more information and how we should refactor this properly.
+pub struct ParserData {
+    pub adjusted_node_namespace: String,
+}
```
emwalker, line 59: I suspect this will evolve into something else at some point.  But seems fine as it is for now.

src/html5/tokenizer/character_reference.rs
```diff
@@ -6,6 +6,7 @@ use crate::bytes::{
 };
 use crate::html5::error_logger::ParserError;
 use crate::html5::tokenizer::replacement_tables::{TOKEN_NAMED_CHARS, TOKEN_REPLACEMENTS};
+
 use crate::html5::tokenizer::{Tokenizer, CHAR_REPLACEMENT};
```
emwalker, line 10: What about keeping the `use` statements together, just so that the linter can do its thing when it organizes them?

src/testing/tree_construction/parser.rs
```diff
@@ -307,7 +308,16 @@ fn test(i: Span) -> IResult<Span, TestSpec> {
...
 
-pub fn parse_str(i: &str) -> Result<Vec<TestSpec>> {
+/// Trims only a single newline from the string, even if there are multiple newlines present.
+fn trim_last_newline(s: String) -> String {
+    if s.ends_with('\n') {
+        s[..s.len() - 1].to_owned()
+    } else {
+        s
+    }
+}
```
emwalker, line 318:
```suggestion
fn trim_last_newline(s: String) -> String {
  if let Some(s) = s.strip_suffix("\n") {
    return s.to_owned();
  }
  s
}
```

tests/tree_construction.rs
```diff
@@ -1,20 +1,14 @@
...
     // tests18.dat
     "<!doctype html><template><plaintext>a</template>b",
 ];
 
-lazy_static! {
-    static ref DISABLED: HashSet<String> = DISABLED_CASES
-        .iter()
-        .map(|s| s.to_string())
-        .collect::<HashSet<_>>();
-}
```
emwalker, line 16: `&[&str].contains(...)` is `O(n)` time, compared to `HashMap::contains(...)`, which is `O(1)` time.  It doesn't matter at all in this case, though, because n is very small (1 right now).

But for a larger set of disabled cases such as the number we had recently, we'd either want the `HashMap` that is being deleted, or we'd want to use [`phf_set!`](https://docs.rs/phf/latest/phf/macro.phf_set.html) (since we're using the phf crate elsewhere).

emwalker (approved):

benches/tree_construction.rs
jaytaph: tried this, but didn't work.. something something borrower :/

src/html5/tokenizer.rs
jaytaph: yes.. this is a dirty hack that we have to get rid of.. there is bidirectional communication between the tokenizer and parser which we can avoid when introducing a tree-builder system. Then, both the tokenizer and parser will talk to it, but not to eachother, which makes the dependencies work again.

src/html5/tokenizer/character_reference.rs
jaytaph: I'm for it, but I think this is a IDE thing that doesn't know how to set them correctly..  I will try and look out for this

tests/tree_construction.rs
jaytaph: I'm not sure about this. I understand the O(1) hash lookup, but since the hash is generated each run, which is O(n), PLUS the overhead of hashing PLUS the small overhead of searching the hash.

src/bytes.rs
```diff
@@ -114,15 +114,6 @@ pub struct CharIterator {
     pub has_read_eof: bool, // True when we just read an EOF
 }
 
-pub enum SeekMode {
```
jaytaph, line 117: I've removed the possibility to seek a stream. For now, that is ok enough, and it saves a lot of time processing newlines and column end calculations.

src/bytes.rs
```diff
@@ -361,15 +306,26 @@ impl CharIterator {
         // If we still can move forward in the stream, move forwards
         if self.position.offset < self.length {
             let c = self.buffer[self.position.offset];
-            self.seek(SeekMode::SeekCur, 1);
+            if c == Ch('\n') {
```
jaytaph, line 309: even though we cannot "seek" in a stream, we can unwind a single character. This would be easy, but at a start of the line, we must know the column ending. We store this in line_offsets so we don't need to calculate it.

src/html5/parser.rs
```diff
@@ -444,6 +444,11 @@ impl<'chars> Html5Parser<'chars> {
         let mut handle_as_script_endtag = false;
 
         match &self.current_token.clone() {
+            Token::Text(value) if self.current_token.is_mixed() => {
```
jaytaph, line 447: This is now a common site when dealing with text.

At these points, special cases are needed for whitespace tokens, null tokens, or both. This section will check if there are mixed characters in the current token, if so, it will split these and insert them in the token queue. From that point on, the parser continues, and on the next loop, we are back to this point again. However, now the is_mixed() will return false (since the tokens have been separated), and we continue with the following match arms.

src/html5/parser.rs
```diff
@@ -1033,15 +1055,18 @@ impl<'chars> Html5Parser<'chars> {
 
                         if process_as_intable_anything_else {
                             let tmp = self.current_token.clone();
-                            self.current_token = Token::Text(tokens);
-
                             self.foster_parenting = true;
-                            self.handle_in_body();
-                            self.foster_parenting = false;
 
+                            let tokens = self.split_mixed_token(&pending_chars);
```
jaytaph, line 1060: special case where we split tokens manually. After this, we call handle_in_body for each of those tokens before continuing the rest of the flow.

src/html5/parser.rs
```diff
@@ -2067,21 +2116,23 @@ impl<'chars> Html5Parser<'chars> {
...
-
-                self.insert_text_element(&self.current_token.clone());
-            }
             Token::Text(..) => {
                 self.reconstruct_formatting();
 
                 self.insert_text_element(&self.current_token.clone());
 
-                self.frameset_ok = false;
+                // If this mixed token does not have whitespace chars, set frameset_ok to false
```
jaytaph, line 2132: We don't care about mixed values here. If there is a whitespace, frames are ok.

src/html5/parser.rs
```diff
@@ -3856,6 +3968,63 @@ impl<'chars> Html5Parser<'chars> {
...
+    ///
+    /// This is needed because the tokenizer does not know about the context of the text it is
+    /// so it will always try to tokenize as greedy as possible. But sometimes we need this split
+    /// to happen where a differentation between whitespaces, null and regular characters are needed.
+    /// Only in those cases, this function is called, and the token will be split into multiple
+    /// tokens.
+    /// The idea is that large blobs of javascript for instance will not be split into separate
+    /// tokens, but still be seen and parsed as a single TextToken.
+    ///
+    fn split_mixed_token(&self, text: &String) -> Vec<Token> {
```
jaytaph, line 3998: I think this can be optimized more, but for now it's ok enoughl

src/html5/tokenizer/token.rs
```diff
@@ -30,6 +30,31 @@ pub enum Token {
     Eof,
 }
 
+impl Token {
+    pub(crate) fn is_mixed(&self) -> bool {
```
jaytaph, line 34: Note we only want to check on ascii-whitespace, not unicode-whitespace

src/html5/parser.rs
```diff
@@ -630,23 +635,24 @@ impl<'chars> Html5Parser<'chars> {
     fn process_html_content(&mut self) {
         if self.ignore_lf {
             if let Token::Text(value) = &self.current_token {
-                if value.eq(&"\n".to_string()) {
-                    self.current_token = self.fetch_next_token();
+                if value.starts_with('\n') {
```
jaytaph, line 638: instead of fetching the next token, we check if the current token has a \n. If so, we modify the current token to remove the \n and continue as usual.

src/html5/parser.rs
```diff
@@ -3721,18 +3812,39 @@ impl<'chars> Html5Parser<'chars> {
...
+            let token = self
+                .tokenizer
+                .next_token(self.parser_data())
+                .expect("tokenizer error");
 
             if let Token::Text(value) = token {
-                for c in value.chars() {
-                    self.token_queue.push(Token::Text(c.to_string()));
-                }
+                self.token_queue.push(Token::Text(value));
```
jaytaph, line 3844: We can safely push the whole value directly as a single token instead of splitting each character as a separate token. This saves us A LOT of time parsing

src/html5/parser/helper.rs
```diff
@@ -239,11 +240,16 @@ impl Html5Parser<'_> {
     }
 
     pub fn insert_text_element(&mut self, token: &Token) {
-        let node = self.create_node(token, HTML_NAMESPACE);
-        let node_id = self.document.get_mut().add_new_node(node);
+        // Skip empty text nodes
+        if let Token::Text(text) = token {
```
jaytaph, line 244: It seems possible that empty tokens arrive here. We have to make sure we don't add them to the tree

src/html5/tokenizer.rs
```diff
@@ -50,6 +50,30 @@ pub struct Tokenizer<'stream> {
     pub error_logger: Rc<RefCell<ErrorLogger>>,
 }
 
+impl<'stream> Tokenizer<'stream> {
+    pub(crate) fn insert_tokens_at_queue_start(&mut self, first_tokens: Vec<Token>) {
```
jaytaph, line 54: This inserts a set of tokens at the front of the queue.

tests/tree_construction.rs
emwalker: `lazy_static!` is macro that allows you to generate once and then never generate again.  So the instantiation only happens once.

emwalker (approved):
I'd pitch for merging this sooner rather than later.

tests/tree_construction.rs
emwalker: If a "run" is a run of all of the tests, then, yes, the hash is instantiated once per run.  If a run is the run of an individual test, this would be mistaken.  The `lazy_static!` code only runs once at the start of execution, and then the result is memoized in `ref DISABLED` and is not computed again.

src/html5/parser.rs
```diff
@@ -630,23 +635,24 @@ impl<'chars> Html5Parser<'chars> {
...
-        // if self.current_token.is_eof() {
-        //     self.parser_finished = true;
-        // }
-
         match self.insertion_mode {
             InsertionMode::Initial => {
                 let mut anything_else = false;
 
                 match &self.current_token.clone() {
+                    Token::Text(value) if self.current_token.is_mixed() => {
```
CharlesChen0823, line 651: IMO, current_token.is_mixed shoud move fetch_next_token, and insert_tokens_at_queue_start could not need.

CharlesChen0823 (approved):

src/html5/parser.rs
jaytaph: @CharlesChen0823 I'm not sure what you exactly mean. Do you mean that we should do this "is_mixed" thing inside the `fetch_next_token` and not directly everywhere in the code?

jaytaph: @CharlesChen0823 for now, I want to merge this PR.. if it's ok with you, can you describe in a PR draft what your idea is to change this?

PR #235, opened 2023-11-06
refactor(css3): replace `TokenKind` with utils & rust traits

neuodev:


emwalker (approved):
LGTM.

Vaguely curious whether this changes the performance in a measurable way, but I don't know if we have anything set up to measure this.

PR #237, opened 2023-11-07
Add document querying

Kiyoshika:
This implements a `Query` builder to pass to `DocumentHandle.query()` method to traverse the tree and return all node ID(s) that match the predicate. This also supports a `FindFirst` and `FindAll` search type. The searching of the tree terminates early on `FindFirst`.

I need to add some notes in the actual `query.rs` file explaining more of its usage, but given a `Query` instance, a node must match ALL conditions provided in the query (i.e., each condition is treated as an AND condition) in order to be returned.

I also want to add a simple extension `DocumentHandle.query_multiple` (or something along those lines) which would take a `Vec<Query>` list of predicates and check if a node matches at least one in order to be returned (to support OR-like conditionals.)

### DIFF
--- a/src/html5/parser.rs
+++ b/src/html5/parser.rs
@@ -1,5 +1,6 @@
 mod attr_replacements;
 pub mod document;
+pub mod query;
 mod quirks;
 pub mod tree_builder;
 
--- a/src/html5/parser/document.rs
+++ b/src/html5/parser/document.rs
@@ -4,6 +4,8 @@ use crate::html5::node::data::doctype::DocTypeData;
 use crate::html5::node::data::{comment::CommentData, text::TextData};
 use crate::html5::node::HTML_NAMESPACE;
 use crate::html5::node::{Node, NodeData, NodeId};
+use crate::html5::parser::query::SearchType;
+use crate::html5::parser::query::{Condition, Query};
 use crate::html5::parser::quirks::QuirksMode;
 use crate::html5::parser::tree_builder::TreeBuilder;
 use crate::html5::util::is_valid_id_attribute_value;
@@ -296,6 +298,25 @@ impl Document {
         self.arena.get_node_mut(*node_id)
     }
 
+    /// Retrieves the next sibling NodeId (to the right) of the reference_node or None.
+    pub fn get_next_sibling(&self, reference_node: NodeId) -> Option<NodeId> {
+        let node = self.get_node_by_id(reference_node)?;
+        let parent = self.get_node_by_id(node.parent?)?;
+
+        let idx = parent
+            .children
+            .iter()
+            .position(|&child_id| child_id == reference_node)
+            .unwrap();
+
+        let next_idx = idx + 1;
+        if parent.children.len() > next_idx {
+            return Some(parent.children[next_idx]);
+        }
+
+        None
+    }
+
     pub fn add_new_node(&mut self, node: Node) -> NodeId {
         // if a node contains attributes when adding to the tree,
         // be sure to handle the special attributes "id" and "class"
@@ -417,6 +438,21 @@ impl Document {
     /// Returns true when the given parent_id is a child of the node_id
     pub fn has_cyclic_reference(&self, node_id: NodeId, parent_id: NodeId) -> bool {
         has_child_recursive(&self.arena, node_id, parent_id)
+    }
+
+    /// Check if a given node's children contain a certain tag name
+    pub fn contains_child_tag(&self, node_id: NodeId, tag: &str) -> bool {
+        if let Some(node) = self.get_node_by_id(node_id) {
+            for child_id in &node.children {
+                if let Some(child) = self.get_node_by_id(*child_id) {
+                    if child.name == tag {
+                        return true;
+                    }
+                }
+            }
+        }
+
+        false
     }
 }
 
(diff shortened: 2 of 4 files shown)


src/html5/parser/document.rs
```diff
@@ -16,6 +17,11 @@ use std::collections::HashMap;
 use std::fmt::Display;
 use std::ops::{Deref, DerefMut};
 
+extern crate queues;
+use queues::*;
+
+use super::query::SearchType;
```
Kiyoshika, line 23: this was auto imported by Zed... reminder for myself to change this to `use crate::...` instead

src/html5/parser/document.rs
```diff
@@ -692,6 +698,59 @@ impl DocumentHandle {
...
+    pub fn query(&self, query: &Query) -> Result<Vec<NodeId>> {
+        if query.search_type == SearchType::Uninitialized {
+            return Err(Error::Query("Query predicate is uninitialized".to_owned()));
+        }
+
+        let mut found_ids = Vec::new();
+
+        let mut node_queue: Queue<NodeId> = queue![];
+        let root_id = self.get().get_root().id;
+        let _ = node_queue.add(root_id);
```
Kiyoshika, line 714: In practice I'm not sure how often the `add/remove` methods fail on a queue, but it may be good to check these errors just for completion...

src/html5/parser/document.rs
```diff
@@ -692,6 +698,59 @@ impl DocumentHandle {
 
         Ok(())
     }
+
+    /// Perform a single query against the document.
+    /// If query search type is uninitialized, returns an error.
+    /// Otherwise, returns a vector of NodeIds that match the predicate in tree order.
```
Kiyoshika, line 704: After thinking about this before bed.. this implementation is technically not correct. According to DOM spec, tree order is depth-first which this current implementation does not (always) do. Will need to be updated

src/html5/parser/document.rs
Kiyoshika: no longer relevant, queue structure removed

src/html5/parser/document.rs
Kiyoshika: Fixed algo to search depth-first instead of breadth-first

src/html5/parser/document.rs
```diff
@@ -296,6 +298,24 @@ impl Document {
...
+
+        let idx = parent
+            .children
+            .iter()
+            .position(|&child_id| child_id == reference_node)
+            .unwrap();
+
+        if parent.children.len() > idx + 1 {
+            return Some(parent.children[idx + 1]);
+        }
```
Kiyoshika, line 314: Maybe add `let next_idx = idx + 1` instead of having it separately twice

src/html5/parser/document.rs
```diff
@@ -418,6 +438,21 @@ impl Document {
...
     }
+
+    /// Check if a given node's children contain a certain tag name
+    pub fn contains_child_tag(&self, node_id: NodeId, tag: &str) -> bool {
+        if let Some(node) = self.get_node_by_id(node_id) {
+            for child_id in &node.children {
+                if let Some(child) = self.get_node_by_id(*child_id) {
+                    if child.name == tag {
+                        return true;
+                    }
```
Kiyoshika, line 449: Replace this with `return child.name == tag`

src/html5/parser/document.rs
```diff
@@ -692,6 +727,103 @@ impl DocumentHandle {
 
         Ok(())
     }
+
+    fn check_query_predicate(
```
Kiyoshika, line 731: Could probably think of a better name for this... maybe `fn matches_query_condition`

src/html5/parser/document.rs
Kiyoshika: Nvm, this change needs to be reverted. It cases an early return when finding the first child not equal to the search tag... guess my brain wasn't working.

Kiyoshika:
I believe I have everything ready now for the first iteration of document querying. Squashed all commits during development into one.

src/html5/parser.rs
```diff
@@ -1,5 +1,6 @@
 mod attr_replacements;
 pub mod document;
+pub mod query;
```
emwalker, line 3: It vaguely looks like the `cargo bench` results are different between this branch and `main`.  I don't have a theory as to what's going on, since this branch seems purely additive.  Perhaps the benchmarks are flawed?

```sh
$ git checkout main
$ cargo bench # three times
$ git checkout feat/tree-traversal
$ cargo bench # three times
# See file:///home/walker/code/emwalker/gosub-engine/target/criterion/report/index.html (in my case)
```

src/html5/parser/document.rs
```diff
@@ -1192,4 +1326,367 @@ mod tests {
...
+        let p_id_4 = doc.create_element("p", NodeId::root(), None, HTML_NAMESPACE);
+        res = doc.insert_attribute("title", "yo", p_id_4);
+        assert!(res.is_ok());
+        res = doc.insert_attribute("style", "cat", p_id_4);
+        assert!(res.is_ok());
+
+        let query = Query::new().contains_attribute("style").find_all();
+        let found_ids = doc.query(&query).unwrap();
+        assert_eq!(found_ids.len(), 3);
+        assert_eq!(found_ids, [div_id_2, div_id_3, p_id_4]);
```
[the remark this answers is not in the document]
Kiyoshika: This would be nice. Manually constructing the document is pretty annoying but it’s what we have for now

src/html5/parser.rs
Kiyoshika: yeah that’s odd, I’m not sure what that would be caused by since I didn’t modify any of the existing implementations, only added a couple methods on the doc itself along with the new structs

emwalker: Can you double-check and see if you see the same thing?

emwalker: Ok, so something weird must be happening on my system.  Thanks for checking.

emwalker (approved):

src/html5/parser/document.rs
```diff
@@ -692,6 +728,103 @@ impl DocumentHandle {
...
+            return Err(Error::Query("Query predicate is uninitialized".to_owned()));
+        }
+
+        let mut found_ids = Vec::new();
+
+        let mut node_stack: Vec<NodeId> = Vec::new();
+        let root_id = self.get().get_root().id;
+        node_stack.push(root_id);
+
+        while !node_stack.is_empty() {
```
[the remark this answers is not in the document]
Kiyoshika: implemented this suggestion, also moved `doc_read.get()` outside the loop which probably has minimal overhead but it didn't really make sense to be inside anyways

PR #238, opened 2023-11-07
feat: add component values iterator

neuodev:


src/css3/new_parser.rs
```diff
@@ -89,6 +99,56 @@ enum ComponentValue {
...
+            ComponentValue::Function(..) | ComponentValue::SimpleBlock(..) => Token::EOF,
+        }
+    }
+}
+
+#[derive(Debug)]
+struct ValueIterator {
+    pub values: Vec<ComponentValue>,
+    pub position: usize,
+}
```
emwalker, line 115: I used the name "CharIterator," but that was just a placeholder.  It was intended to be transitional until the time when I would be able to do

```rust
impl Iterator for CharIterator {
  type Item = bytes::Bytes;

  fn next(&mut self) -> Option<Self::Item> {
    // ...
  }
}
```

(This is what would make it a real iterator.)

I now have doubts that `CharIterator` will actually be a real iterator (implementing `Iterator`), and I'm guessing the name will change to something like `BufferQueue`.  So I'd recommend not copying the the "Iterator" name unless it's an actual iterator.  What you have here is more of a `ValueBuffer` or something.  You can probably think of a better name.

emwalker (approved):
LGTM with suggestion to not use "iterator" in the name unless this will become an actual iterator.

PR #249, opened 2023-11-10
Added initial config store

jaytaph:
Initial attempt on setting up a general config store.

It can serve from multiple backends (jsonfile, memory, sqlite).

### DIFF
--- a/src/bin/config-store.rs
+++ b/src/bin/config-store.rs
@@ -1,9 +1,9 @@
 use clap::{Parser, Subcommand};
 use derive_more::Display;
-use gosub_engine::config::settings::{Setting, SettingInfo};
+use gosub_engine::config::settings::Setting;
 use gosub_engine::config::storage::json_storage::JsonStorageAdapter;
 use gosub_engine::config::storage::sqlite_storage::SqliteStorageAdapter;
-use gosub_engine::config::ConfigStore;
+use gosub_engine::config::{ConfigStore, StorageAdapter};
 
 #[derive(Debug, Parser)]
 #[clap(name = "Config-Store", version = "0.1.0", author = "Gosub")]
@@ -69,61 +69,15 @@ struct GlobalOpts {
     path: String,
 }
 
-enum Config {
-    Sqlite(ConfigStore<SqliteStorageAdapter>),
-    Json(ConfigStore<JsonStorageAdapter>),
-}
-
-impl Config {
-    fn find(&self, search: &str) -> Vec<String> {
-        match self {
-            Self::Sqlite(store) => store.find(search),
-            Self::Json(store) => store.find(search),
-        }
-    }
-
-    fn has(&self, key: &str) -> bool {
-        match self {
-            Self::Sqlite(store) => store.has(key),
-            Self::Json(store) => store.has(key),
-        }
-    }
-
-    fn get(&mut self, key: &str) -> Setting {
-        match self {
-            Self::Sqlite(store) => store.get(key),
-            Self::Json(store) => store.get(key),
-        }
-    }
-
-    fn get_info(&self, key: &str) -> Option<SettingInfo> {
-        match self {
-            Self::Sqlite(store) => store.get_info(key),
-            Self::Json(store) => store.get_info(key),
-        }
-    }
-
-    fn set(&mut self, key: &str, value: Setting) {
-        match self {
-            Self::Sqlite(store) => store.set(key, value),
-            Self::Json(store) => store.set(key, value),
-        }
-    }
-}
-
 fn main() {
     let args = Cli::parse();
 
-    let mut store: Config = match args.global_opts.engine {
-        Engine::Sqlite => {
-            let store = SqliteStorageAdapter::new(args.global_opts.path.as_str());
-            Config::Sqlite(ConfigStore::new(store, true))
-        }
-        Engine::Json => {
-            let store = JsonStorageAdapter::new(args.global_opts.path.as_str());
-            Config::Json(ConfigStore::new(store, true))
-        }
+    let storage_box: Box<dyn StorageAdapter> = match args.global_opts.engine {
+        Engine::Sqlite => Box::new(SqliteStorageAdapter::new(args.global_opts.path.as_str())),
+        Engine::Json => Box::new(JsonStorageAdapter::new(args.global_opts.path.as_str())),
     };
+
+    let mut store = ConfigStore::new(storage_box, true);
 
     match args.command {
         Commands::View { key } => {
--- a/src/bin/config-store.rs
+++ b/src/bin/config-store.rs
@@ -1,9 +1,9 @@
 use clap::{Parser, Subcommand};
 use derive_more::Display;
-use gosub_engine::config::{ConfigStore, StorageAdapter};
 use gosub_engine::config::settings::Setting;
 use gosub_engine::config::storage::json_storage::JsonStorageAdapter;
 use gosub_engine::config::storage::sqlite_storage::SqliteStorageAdapter;
+use gosub_engine::config::{ConfigStore, StorageAdapter};
 
 #[derive(Debug, Parser)]
 #[clap(name = "Config-Store", version = "0.1.0", author = "Gosub")]
@@ -56,30 +56,28 @@ impl std::str::FromStr for Engine {
     }
 }
 
-
 #[derive(Debug, Parser)]
 struct GlobalOpts {
     #[clap(short = 'e', long = "engine", global = true, default_value = "sqlite")]
     engine: Engine,
-    #[clap(short = 'p', long = "path", global = true, default_value = "settings.db")]
+    #[clap(
+        short = 'p',
+        long = "path",
+        global = true,
+        default_value = "settings.db"
+    )]
     path: String,
 }
 
 fn main() {
     let args = Cli::parse();
 
-    let storage_box: Box<dyn StorageAdapter>;
-    match args.global_opts.engine {
-        Engine::Sqlite => {
-            storage_box = Box::new(SqliteStorageAdapter::new(args.global_opts.path.as_str()));
-        }
-        Engine::Json => {
-            storage_box = Box::new(JsonStorageAdapter::new(args.global_opts.path.as_str()));
-        }
-    }
+    let storage_box: Box<dyn StorageAdapter> = match args.global_opts.engine {
+        Engine::Sqlite => Box::new(SqliteStorageAdapter::new(args.global_opts.path.as_str())),
+        Engine::Json => Box::new(JsonStorageAdapter::new(args.global_opts.path.as_str())),
+    };
 
     let mut store = ConfigStore::new(storage_box, true);
-
 
     match args.command {
         Commands::View { key } => {
@@ -92,23 +90,26 @@ fn main() {
             let value = store.get(key.as_str(), None);
 
             println!("Key            : {}", key);
-            println!("Current Value  : {}", value.to_string());
-            println!("Default Value  : {}", info.default.to_string());
+            println!("Current Value  : {}", value);
+            println!("Default Value  : {}", info.default);
             println!("Description    : {}", info.description);
         }
         Commands::List => {
-            for key in store.find("*".into()) {
+            for key in store.find("*") {
                 let value = store.get(key.as_str(), None);
-                println!("{:40}: {}", key, value.to_string());
+                println!("{:40}: {}", key, value);
             }
         }
         Commands::Set { key, value } => {
-            store.set(key.as_str(), Setting::from_string(value.as_str()).expect("incorrect value"));
+            store.set(
+                key.as_str(),
+                Setting::from_string(value.as_str()).expect("incorrect value"),
+            );
         }
         Commands::Search { key } => {
-            for key in store.find(key.as_str().into()) {
+            for key in store.find(key.as_str()) {
                 let value = store.get(key.as_str(), None);
-                println!("{:40}: {}", key, value.to_string());
+                println!("{:40}: {}", key, value);
             }
         }
     }
--- a/src/bin/config-store.rs
+++ b/src/bin/config-store.rs
@@ -101,9 +101,7 @@ fn main() {
             }
         }
         Commands::Set { key, value } => {
-            store.set(
-                &key, Setting::from_string(&value).expect("incorrect value"),
-            );
+            store.set(&key, Setting::from_str(&value).expect("incorrect value"));
         }
         Commands::Search { key } => {
             for key in store.find(key.as_str()) {
--- a/src/bin/config-store.rs
+++ b/src/bin/config-store.rs
@@ -1,0 +1,115 @@
+use clap::{Parser, Subcommand};
+use derive_more::Display;
+use gosub_engine::config::{ConfigStore, StorageAdapter};
+use gosub_engine::config::settings::Setting;
+use gosub_engine::config::storage::json_storage::JsonStorageAdapter;
+use gosub_engine::config::storage::sqlite_storage::SqliteStorageAdapter;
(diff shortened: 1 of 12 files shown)


src/config.rs
```diff
@@ -0,0 +1,153 @@
...
+#[derive(Debug, Deserialize)]
+struct JsonEntry {
+    key: String,
+    #[serde(rename = "type")]
+    entry_type: String,
+    default: String,
+    description: String,
+}
+
+trait StorageAdapter {
```
jaytaph, line 21: Any adapter that implements this can be used as a storage engine for the configuration store.

src/config.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// Stores a given setting to the storage
+    fn set_setting(&mut self, key: &str, value: Setting);
+    /// Retrieves all the settings in the storage
+    fn get_all_settings(&self) -> HashMap<String, Setting>;
+}
+
+/// Configuration store is the place where the gosub engine can find all configurable options
+pub struct ConfigStore {
+    /// A hashmap of all settings so we can search o(1) time
+    settings: HashMap<String, Setting>,
```
jaytaph, line 33: Settings are only the setting values.. they are not stored in SettingsInfo, which is only the template data for the configuration

src/config.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// A hashmap of all setting descriptions
+    settings_info: HashMap<String, SettingInfo>,
+    /// Keys of all settings so we can iterate keys easily
+    setting_keys: Vec<String>,
```
jaytaph, line 37: I'm not sure if we need this, but if we need to iterate all keys, we need to store them in a vec as well.

src/config.rs
```diff
@@ -0,0 +1,153 @@
...
+            setting_keys: Vec::new(),
+            storage
+        };
+        
+        // Populate the settings from the json file
+        store.populate_settings();
+
+        // preload the settings if requested
+        if preload {
+            let all_settings = store.storage.get_all_settings();
```
jaytaph, line 57: We might want to preload all settings from a storage engine, or lazy load them when needed. It depends a bit on the latency of the storage engine (ie: remote configuration would take a long time)

src/config.rs
```diff
@@ -0,0 +1,153 @@
...
+            for (key, value) in all_settings {
+                store.settings.insert(key, value);
+            }
+        }
+
+        store
+    }
+
+    /// Returns a list of keys that mathces the given search string (can use *)
+    fn find_setting_keys(&self, search: &str) -> Vec<String> {
```
jaytaph, line 67: This allows you to use  "dns.resolv*"  and it will match every key that starts with this. Easy for filtering

src/config.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        // Setting not found, try and load it from the storage adapter
+        if let Some(setting) = self.storage.get_setting(key) {
+            self.settings.insert(key.to_string(), setting.clone());
+            return setting.clone();
+        }
+
+        // Panic if we can't find the setting, and we don't have a default
+        if default.is_none() {
+            panic!("Setting {} not found", key);
```
jaytaph, line 96: What should we do if we cannot find the config, and there is no default given?

Should we even use a default, as we have defaults in the settingsinfo per config setting anyway

src/config/settings.rs
```diff
@@ -0,0 +1,123 @@
...
+impl Setting {
+    // first element is the type:
+    //   b:true
+    //   i:-123
+    //   u:234
+    //   s:hello world
+    //   m:foo,bar,baz
+
+    /// Converts a string to a setting or None when the string is invalid
+    pub(crate) fn from_string(p0: &str) -> Option<Setting> {
```
jaytaph, line 21: A setting can have different types. bools, ints, strings etc. This tries to accomodate that

src/config/settings.json
```diff
@@ -0,0 +1,46 @@
+{
```
jaytaph, line 1: This config file is imported into the engine lib binary, but it allows for a more flexible setup than incode configuration.

src/config.rs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        default.unwrap()
+    }
+
+    pub fn set(&mut self, key: &str, value: Setting) {
```
jaytaph, line 110: We do not do a check to see if the setting type is actually allowed.

For instance, you can set a bool (`b:true`) on a uint typed setting.

src/config.rs
jaytaph: Changed this into returning the default value as defined in the setting info for that key. This means we do not pass a default value when getting the key.

src/config.rs
jaytaph: this has been fixed as well..  it will check the type you add against the type of the default setting.

src/config/settings.rs
```diff
@@ -0,0 +1,140 @@
...
+#[derive(Clone, PartialEq, Debug)]
+pub enum Setting {
+    SInt(isize),
+    UInt(usize),
+    String(String),
+    Bool(bool),
+    Map(Vec<String>),
+}
+
+impl Serialize for Setting {
```
jaytaph, line 15: serialize / deserialize are needed for using serde_json in the json-adapter. It will convert a Setting directory to a string

src/config/storage/json_storage.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        adapter.read_file();
+
+        adapter
+    }
+}
+
+impl StorageAdapter for JsonStorageAdapter {
+    fn get_setting(&self, key: &str) -> Option<Setting> {
+        match self.elements.get(key) {
```
jaytaph, line 54: basically the json adapter reads the whole json during init, and not during every get_setting call.. This would create a secondary layer of caching and doesn't reflect any changes that are done on disk. We might want to reread the whole json on each json file, OR do something smarter like checking access/modify time

src/config/storage/json_storage.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /// Write the self.elements hashmap back to the file by truncating the file and writing the
+    /// data again.
+    fn write_file(&mut self) {
+        let mut file = self.file_mutex.lock().expect("Mutex lock failed");
+
+        let json = serde_json::to_string_pretty(&self.elements).expect("failed to serialize");
+
+        file.set_len(0).expect("failed to truncate file");
```
jaytaph, line 98: we truncate the current file, since this is the file that is guarded by the mutex

src/config/storage/json_storage.rs
```diff
@@ -0,0 +1,102 @@
...
+                let mut file = File::create(path).expect("cannot create json file");
+                file.write_all(json.as_bytes()).expect("failed to write to file");
+
+                file
+            }
+        };
+
+        let mut adapter = JsonStorageAdapter {
+            path: path.to_string(),
+            file_mutex: Arc::new(Mutex::new(file)),
```
jaytaph, line 42: A mutex is used so we can easily use this is a multi-threaded environment

src/config.rs
```diff
@@ -0,0 +1,195 @@
...
+    key: String,
+    #[serde(rename = "type")]
+    entry_type: String,
+    default: String,
+    description: String,
+}
+
+/// StorageAdapter is the interface for storing and retrieving settings
+/// This can be used to store settings in a database, json file, etc
+pub trait StorageAdapter {
```
emwalker, line 24: Just a thought:

```suggestion
pub trait Store {
```

src/config.rs
```diff
@@ -0,0 +1,195 @@
...
+    settings_info: HashMap<String, SettingInfo>,
+    /// Keys of all settings so we can iterate keys easily
+    setting_keys: Vec<String>,
+    /// The storage adapter used for persisting and loading keys
+    storage: Box<dyn StorageAdapter>,
+}
+
+impl ConfigStore {
+    /// Creates a new store with the given storage adapter and preloads the store if needed
+    pub fn new(storage: Box<dyn StorageAdapter>, preload: bool) -> Self {
```
emwalker, line 49: I suspect we can avoid using `Box<dyn ...>` if we replace it with a generic struct parameter.  But in this specific case I doubt it would make a difference, since this is not (presumably) critical path code.

src/bin/config-store.rs
```diff
@@ -0,0 +1,116 @@
...
+            );
+        }
+        Commands::Search { key } => {
+            for key in store.find(key.as_str()) {
+                let value = store.get(key.as_str());
+                println!("{:40}: {}", key, value);
+            }
+        }
+    }
+}
```
emwalker, line 116: I would add this line to the `Makefile` for this command:

```patch
diff --git a/Makefile b/Makefile
index e25ba44..7a362d3 100644
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ test_fmt:
 test_commands:
        cargo run --bin parser-test >/dev/null
+       cargo run --bin config-store list >/dev/null
 
 help: ## Display available commands
        echo "Available make commands:"
```

src/bin/config-store.rs
```diff
@@ -0,0 +1,116 @@
...
+        }
+        Commands::Set { key, value } => {
+            store.set(
+                key.as_str(),
+                Setting::from_string(value.as_str()).expect("incorrect value"),
+            );
+        }
+        Commands::Search { key } => {
+            for key in store.find(key.as_str()) {
+                let value = store.get(key.as_str());
```
emwalker, line 111:
```suggestion
                let value = store.get(key);
```

src/bin/config-store.rs
```diff
@@ -0,0 +1,116 @@
...
+        Commands::List => {
+            for key in store.find("*") {
+                let value = store.get(key.as_str());
+                println!("{:40}: {}", key, value);
+            }
+        }
+        Commands::Set { key, value } => {
+            store.set(
+                key.as_str(),
+                Setting::from_string(value.as_str()).expect("incorrect value"),
```
emwalker, line 106:
```suggestion
                &key, Setting::from_string(&value).expect("incorrect value"),
```

src/bin/config-store.rs
```diff
@@ -0,0 +1,116 @@
...
+            let value = store.get(key.as_str());
+
+            println!("Key            : {}", key);
+            println!("Current Value  : {}", value);
+            println!("Default Value  : {}", info.default);
+            println!("Description    : {}", info.description);
+        }
+        Commands::List => {
+            for key in store.find("*") {
+                let value = store.get(key.as_str());
```
emwalker, line 99:
```suggestion
                let value = store.get(&key);
```

src/bin/config-store.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    match args.command {
+        Commands::View { key } => {
+            if !store.has(key.as_str()) {
+                println!("Key not found");
+                return;
+            }
+
+            let info = store.get_info(key.as_str()).unwrap();
+            let value = store.get(key.as_str());
```
emwalker, line 90:
```suggestion
            let info = store.get_info(&key).expect("missing required key");
            let value = store.get(&key);
```

src/bin/config-store.rs
```diff
@@ -0,0 +1,116 @@
...
+    let storage_box: Box<dyn StorageAdapter> = match args.global_opts.engine {
+        Engine::Sqlite => Box::new(SqliteStorageAdapter::new(args.global_opts.path.as_str())),
+        Engine::Json => Box::new(JsonStorageAdapter::new(args.global_opts.path.as_str())),
+    };
+
+    let mut store = ConfigStore::new(storage_box, true);
+
+    match args.command {
+        Commands::View { key } => {
+            if !store.has(key.as_str()) {
```
emwalker, line 84:
```suggestion
            if !store.has(&key) {
```

src/config.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /// Returns a list of keys that matches the given search string (can use ? and *) for search
+    /// wildcards.
+    pub fn find(&self, search: &str) -> Vec<String> {
+        let search = WildMatch::new(search);
+
+        let mut keys = Vec::new();
+        for key in &self.setting_keys {
+            if search.matches(key.as_str()) {
```
emwalker, line 83:
```suggestion
            if search.matches(key) {
```

src/config.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /// Populates the settings in the store from the settings.json file
+    fn populate_settings(&mut self) {
+        let json_data = serde_json::from_str(SETTINGS_JSON);
+        if json_data.is_err() {
+            panic!("Failed to parse settings.json");
+        }
+
+        let json_data = json_data.unwrap();
```
emwalker, line 143:
```suggestion
        let json_data: Value =
            serde_json::from_str(SETTINGS_JSON).expect("Failed to parse settings.json");

```

src/config/settings.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+impl Display for Setting {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Setting::SInt(value) => write!(f, "i:{}", value),
+            Setting::UInt(value) => write!(f, "u:{}", value),
+            Setting::String(value) => write!(f, "s:{}", value),
+            Setting::Bool(value) => write!(f, "b:{}", value),
```
emwalker, line 45:
```suggestion
            Setting::SInt(value) => write!(f, "i:{value}"),
            Setting::UInt(value) => write!(f, "u:{value}"),
            Setting::String(value) => write!(f, "s:{value}"),
            Setting::Bool(value) => write!(f, "b:{value}"),
```

src/config/settings.rs
```diff
@@ -0,0 +1,146 @@
...
+impl Setting {
+    // first element is the type:
+    //   b:true
+    //   i:-123
+    //   u:234
+    //   s:hello world
+    //   m:foo,bar,baz
+
+    /// Converts a string to a setting or None when the string is invalid
+    pub fn from_string(key: &str) -> Option<Setting> {
```
emwalker, line 68:
```suggestion
    pub fn from_str(key: &str) -> Option<Setting> {
```

emwalker (approved):
LGTM with suggestions.

src/bin/config-store.rs
emwalker: I would also add `settings.db` to `.gitignore` or put it under a `tmp/` directory that is ignored:

```patch
diff --git a/.gitignore b/.gitignore
index 5ce6f87..0a3c97f 100644
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@ target/
 *.swp*
 
 # Local testing 
-local/
\ No newline at end of file
+local/
+settings.db
```

src/bin/config-store.rs
```diff
@@ -101,11 +101,11 @@ fn main() {
             }
         }
         Commands::Set { key, value } => {
-            store.set(&key, Setting::from_str(&value).expect("incorrect value"));
+            store.set(&key, Setting::from_string(&value).expect("incorrect value"));
```
emwalker, line 104: Curious why this was reverted.  The name `from_str` is common in Rust, while the name `from_string` looks like someone didn't know about `from_str`.

jaytaph: @emwalker this is due to an warning that is emitted by rust itself... it clashes with a standard `from_str` that is found in a ::std trait and could cause conflicts

emwalker: Ah, gotcha.  Perhaps that means we should be implementing that trait instead of defining `from_string`.  It would not be hard.

```rust
impl FromStr for Setting {
  fn from_str(value: &str) -> Self {
    todo!()
  }
}
```

PR #251, opened 2023-11-11
Add document tree iterator

Kiyoshika:
This adds a `TreeIterator` which takes a `&DocumentHandle` and constructs an iterator which walks through the tree one node at a time in tree order.

EDIT: below statement is incorrect, mutations are reflected in the iterator if it's still kept "open." I added a warning in the struct description about this.
~~This iterator is a snapshot of the document at the time the iterator is created and is not updated from mutations. In that case, a new iterator must be created.~~

At the moment, the iterator returns `NodeId`s (thanks for the help from @emwalker) which is good enough for now but ideally I would like this to return some type of `&Node` or `Rc<Node>` that the user agent can use directly without having to touch the document's "internals"

`DocumentHandle.query()` has been updated to use this new iterator as well which still passes all tests

### DIFF
--- a/src/html5/parser/document.rs
+++ b/src/html5/parser/document.rs
@@ -185,20 +185,28 @@ impl TreeBuilder for DocumentTaskQueue {
         new_id
     }
 
-    fn create_text(&mut self, content: &str, parent_id: NodeId) {
+    fn create_text(&mut self, content: &str, parent_id: NodeId) -> NodeId {
         let text = DocumentTask::CreateText {
             content: content.to_owned(),
             parent_id,
         };
+        let new_id = self.next_node_id;
+        self.next_node_id = self.next_node_id.next();
         self.tasks.push(text);
+
+        new_id
     }
 
-    fn create_comment(&mut self, content: &str, parent_id: NodeId) {
+    fn create_comment(&mut self, content: &str, parent_id: NodeId) -> NodeId {
         let comment = DocumentTask::CreateComment {
             content: content.to_owned(),
             parent_id,
         };
+        let new_id = self.next_node_id;
+        self.next_node_id = self.next_node_id.next();
         self.tasks.push(comment);
+
+        new_id
     }
 
     fn insert_attribute(&mut self, key: &str, value: &str, element_id: NodeId) -> Result<()> {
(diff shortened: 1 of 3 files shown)


src/html5/parser/document.rs
```diff
@@ -185,20 +185,28 @@ impl TreeBuilder for DocumentTaskQueue {
         new_id
     }
 
-    fn create_text(&mut self, content: &str, parent_id: NodeId) {
+    fn create_text(&mut self, content: &str, parent_id: NodeId) -> NodeId {
```
Kiyoshika, line 188: Making `create_text` and `create_comment` return a node ID is not mandatory but I did it for convenience (in the tests) and also it makes the tree builder interface more consistent anyways..

src/html5/parser/document.rs
```diff
@@ -786,25 +794,13 @@ impl DocumentHandle {
...
-        let mut node_stack: Vec<NodeId> = Vec::new();
-        let root_id = self.get().get_root().id;
-        node_stack.push(root_id);
-
         let doc_read = self.get();
+        let mut found_ids = Vec::new();
+        let tree_iterator = TreeIterator::new(self);
 
-        while let Some(current_node_id) = node_stack.pop() {
+        for current_node_id in tree_iterator {
```
Kiyoshika, line 801: cool side effect of using the iterator, we get this nice Python-like iteration syntax

src/html5/parser/document.rs
```diff
@@ -898,10 +894,54 @@ impl DocumentBuilder {
     }
 }
 
+/// Constructs an iterator from a given DocumentHandle.
+/// Note that this iterator is a snapshot of the DocumentHandle
+/// at the time of creation; it does NOT update with tree mutations
+/// and a new iterator must be used.
```
Kiyoshika: now that I think about it, this may actually be wrong. We are just cloning the reference to the document so technically mutations would be reflected, the iterator would likely just be unreliable depending on what nodes have been inserted/mutated

Kiyoshika: Nvm, after actually testing this, the iterator is "consumed" if calling next() until the last element and will return None even if appending more nodes to the doc.

jaytaph (approved):

src/html5/parser/document.rs
```diff
@@ -1688,4 +1728,65 @@ mod tests {
...
+            p_id_2,
+            text_id_2,
+            p_id_3,
+            text_id_3,
+        ];
+
+        let mut traversed_nodes = Vec::new();
+        for current_node_id in tree_iterator {
+            traversed_nodes.push(current_node_id);
+        }
```
emwalker, line 1767: Also:

```rust
let traversed_nodes = tree_iterator.collect::<Vec<_>>();
```

PR #260, opened 2023-11-13
deprecated 'make fix'

jaytaph:
Use 'make format' instead, as this is more descriptive

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,10 @@ build: ## Build the project
 	section "Cargo build" ;\
 	cargo build
 
-fix:  ## Fix formatting and clippy errors
+fix: ## Fix formatting and clippy errors (deprecated)
+	echo "Use 'make format' instead"
+
+format:  ## Fix formatting and clippy errors
 	cargo fmt
 	cargo clippy --fix --allow-dirty --allow-staged
 


Makefile
```diff
@@ -14,7 +14,10 @@ build: ## Build the project
 	section "Cargo build" ;\
 	cargo build
 
-fix:  ## Fix formatting and clippy errors
+fix: ## Fix formatting and clippy errors (deprecated)
+	echo "Use 'make format' instead"
+
```
Sharktheone, line 19: Why not remove the old completely? I mean keeping this has no big benefits I think.

Sharktheone (changes requested):
You can keep it if you want, but i don't see a huge benefit of keeping this.

Kiyoshika (approved):

Makefile
Kiyoshika: My guess is for anyone used to using `make fix` that doesn't see this PR and maybe confused when it's gone. Granted, it's not hard to look at the Makefile but it'll at least tell you it's moved; then maybe just completely remove it at some point

Although I don't have a preference for either way, whether we remove it now or later

emwalker (approved):

PR #262, opened 2023-11-14
initial commit for DNS local resolver

jaytaph:
This is the initial commit for the DNS resolver component. 
In consists of a single entrypoint which returns a DNSEntry. It is up to the caller to actually extract the correct ips it wants (either the ipv4() or ipv6().

There are 3 resolvers that are called in sequence until one of them returns a hit:

  - caching resolver
  - local table resolver
  - remote resolver

The caching resolver has a fixed size cache that will retrieve quickly a dns record if it is available.
The local table resolver contains a list of items that will be checked against (including wildcards). This allows to override hosts just like an /etc/host file or dnsmasq would do
The remote resolver calls hickory-dns, which will resolve the query.

It is not 100% finished, but the major work is done. As soon as we have the config store in place, we can use that for configuration purposes.

Another things left to do is expiring of DNS records.

```
    Finished test [unoptimized + debuginfo] target(s) in 0.08s
     Running unittests src/lib.rs (target/debug/deps/gosub_engine-761d2bb186cfbf97)
2023-11-17T21:57:37.971Z INFO  [gosub_engine::dns] Resolving example.org for Ipv4
2023-11-17T21:57:37.971Z DEBUG [gosub_engine::dns] Trying resolver: cache resolver
2023-11-17T21:57:37.971Z DEBUG [gosub_engine::dns] Trying resolver: local table resolver
2023-11-17T21:57:37.971Z TRACE [gosub_engine::dns::local] example.org: not found in local table
2023-11-17T21:57:37.971Z DEBUG [gosub_engine::dns] Trying resolver: remote resolver
2023-11-17T21:57:37.971Z TRACE [gosub_engine::dns::remote] example.org: resolving with ["ipv4"]
2023-11-17T21:57:37.977Z TRACE [gosub_engine::dns::remote] example.org: found ipv4 address 93.184.216.34
2023-11-17T21:57:37.978Z DEBUG [gosub_engine::dns] Found entry DnsEntry { domain: "example.org", ips: [93.184.216.34], has_ipv4: true, has_ipv6: false, iter: 0, expires: 0 }
2023-11-17T21:57:37.978Z TRACE [gosub_engine::dns::cache] example.org: announcing to cache
2023-11-17T21:57:37.978Z TRACE [gosub_engine::dns::cache] adding new entry to cache
ipv4: 93.184.216.34
Took 6626 microseconds.
2023-11-17T21:57:37.978Z INFO  [gosub_engine::dns] Resolving example.org for Ipv6
2023-11-17T21:57:37.978Z DEBUG [gosub_engine::dns] Trying resolver: cache resolver
2023-11-17T21:57:37.978Z TRACE [gosub_engine::dns::cache] example.org: no ipv6 addresses found in entry
2023-11-17T21:57:37.978Z DEBUG [gosub_engine::dns] Trying resolver: local table resolver
2023-11-17T21:57:37.978Z TRACE [gosub_engine::dns::local] example.org: not found in local table
2023-11-17T21:57:37.978Z DEBUG [gosub_engine::dns] Trying resolver: remote resolver
2023-11-17T21:57:37.978Z TRACE [gosub_engine::dns::remote] example.org: resolving with ["ipv6"]
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::remote] example.org: found ipv6 address 2606:2800:220:1:248:1893:25c8:1946
2023-11-17T21:57:37.986Z DEBUG [gosub_engine::dns] Found entry DnsEntry { domain: "example.org", ips: [2606:2800:220:1:248:1893:25c8:1946], has_ipv4: false, has_ipv6: true, iter: 0, expires: 0 }
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] example.org: announcing to cache
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] example.org: updating existing entry to cache
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] current entries: [93.184.216.34]
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] new entries: [2606:2800:220:1:248:1893:25c8:1946]
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] merged entries: [93.184.216.34, 2606:2800:220:1:248:1893:25c8:1946]
ipv6: 2606:2800:220:1:248:1893:25c8:1946
Took 8253 microseconds.
2023-11-17T21:57:37.986Z INFO  [gosub_engine::dns] Resolving example.org for Ipv4
2023-11-17T21:57:37.986Z DEBUG [gosub_engine::dns] Trying resolver: cache resolver
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] example.org: found in cache with correct resolve type
2023-11-17T21:57:37.986Z DEBUG [gosub_engine::dns] Found entry DnsEntry { domain: "example.org", ips: [93.184.216.34, 2606:2800:220:1:248:1893:25c8:1946], has_ipv4: true, has_ipv6: true, iter: 0, expires: 0 }
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] example.org: announcing to cache
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] example.org: updating existing entry to cache
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] current entries: [93.184.216.34, 2606:2800:220:1:248:1893:25c8:1946]
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] new entries: [93.184.216.34, 2606:2800:220:1:248:1893:25c8:1946]
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] merged entries: [93.184.216.34, 2606:2800:220:1:248:1893:25c8:1946]
ipv4: 93.184.216.34
Took 57 microseconds.
2023-11-17T21:57:37.986Z INFO  [gosub_engine::dns] Resolving example.org for Both
2023-11-17T21:57:37.986Z DEBUG [gosub_engine::dns] Trying resolver: cache resolver
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] example.org: found in cache with correct resolve type
2023-11-17T21:57:37.986Z DEBUG [gosub_engine::dns] Found entry DnsEntry { domain: "example.org", ips: [93.184.216.34, 2606:2800:220:1:248:1893:25c8:1946], has_ipv4: true, has_ipv6: true, iter: 0, expires: 0 }
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] example.org: announcing to cache
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] example.org: updating existing entry to cache
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] current entries: [93.184.216.34, 2606:2800:220:1:248:1893:25c8:1946]
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] new entries: [93.184.216.34, 2606:2800:220:1:248:1893:25c8:1946]
2023-11-17T21:57:37.986Z TRACE [gosub_engine::dns::cache] merged entries: [93.184.216.34, 2606:2800:220:1:248:1893:25c8:1946]
ipv4: 93.184.216.34
ipv6: 2606:2800:220:1:248:1893:25c8:1946
Took 45 microseconds.
```

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -70,24 +70,9 @@ impl DnsEntry {
     fn ipv6(&self) -> Vec<IpAddr> {
         self.ips.iter().filter(|x| x.is_ipv6()).copied().collect()
     }
-}
 
-impl Iterator for DnsEntry {
-    type Item = IpAddr;
-
-    /// With next() you can simply iterate over all the ips in the list
-    fn next(&mut self) -> Option<Self::Item> {
-        if self.iter >= self.ips.len() {
-            // reset iterator at the end
-            self.iter = 0;
-
-            return None;
-        }
-
-        let ip = self.ips[self.iter];
-        self.iter += 1;
-
-        Some(ip)
+    fn iter(&self) -> impl Iterator<Item = &IpAddr> {
+        self.ips.iter()
     }
 }
 
@@ -106,17 +91,10 @@ trait DnsResolver {
     /// Resolves a domain name for a given resolver_type
     fn resolve(&mut self, domain: &str, resolve_type: ResolveType) -> Result<DnsEntry>;
     /// Announces the resolved dns entry for the domain to a resolver
-    fn announce(&mut self, domain: &str, entry: &DnsEntry);
-
+    fn announce(&mut self, _domain: &str, _entry: &DnsEntry) {}
     // name for debugging purposes
     fn name(&self) -> &'static str;
 }
-
-// impl fmt::Debug for dyn DnsResolver {
-//     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-//         write!(f, "dns resolver")
-//     }
-// }
 
 trait DnsCache {
     /// Flush all domains
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -1,0 +1,217 @@
+mod cache;
+mod local;
+mod remote;
+
+use crate::types;
+use crate::types::Result;
+use core::str::FromStr;
+use derive_more::Display;
+use log::{debug, info};
+use std::net::IpAddr;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+/// A DNS entry is a mapping of a domain to zero or more IP address mapping
+#[derive(Default, Clone, Debug, PartialEq)]
+struct DnsEntry {
+    // domain name
+    domain: String,
+    // // Ip type that is stored in this entry (could be Ipv4, IPv6 or Both)
+    // ip_type:   ResolveType,
+    // List of addresses for this domain
+    ips: Vec<IpAddr>,
+
+    /// True when the ips list has ipv4 addresses
+    has_ipv4: bool,
+    /// True when the ips list has ipv6 addresses
+    has_ipv6: bool,
+
+    // Internal iterator pointer
+    iter: usize,
+    /// expiry time after epoch
+    expires: u64,
+}
+
+impl DnsEntry {
+    /// Instantiate a new domain name entry with set of ips
+    pub(crate) fn new(domain: &str, ips: Vec<&str>) -> DnsEntry {
+        let mut entry = DnsEntry {
+            domain: domain.to_string(),
+            ..Default::default()
+        };
+
+        for ip in ips {
+            if let Ok(ip) = IpAddr::from_str(ip) {
+                if ip.is_ipv4() {
+                    entry.has_ipv4 = true;
+                }
+                if ip.is_ipv6() {
+                    entry.has_ipv6 = true;
+                }
+                entry.ips.push(ip);
+            }
+        }
+
+        entry
+    }
+
+    /// Returns true if the dns entry has expired
+    pub fn expired(&self) -> bool {
+        self.expires
+            < SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
+                .as_secs()
+    }
+
+    fn ipv4(&self) -> Vec<IpAddr> {
+        self.ips.iter().filter(|x| x.is_ipv4()).copied().collect()
+    }
+
+    fn ipv6(&self) -> Vec<IpAddr> {
+        self.ips.iter().filter(|x| x.is_ipv6()).copied().collect()
+    }
+}
+
+impl Iterator for DnsEntry {
+    type Item = IpAddr;
+
+    /// With next() you can simply iterate over all the ips in the list
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.iter >= self.ips.len() {
+            // reset iterator at the end
+            self.iter = 0;
+
+            return None;
+        }
+
+        let ip = self.ips[self.iter];
+        self.iter += 1;
+
+        Some(ip)
+    }
+}
+
+/// Type of DNS resolution
+#[derive(Display, Debug, PartialEq, Clone)]
+pub enum ResolveType {
+    /// Only resolve IPV4 addresses (A)
+    Ipv4,
+    /// Only resolve IPV6 addresses (AAAA)
+    Ipv6,
+    /// Resolve both IPV4 and IPV6 addresses
+    Both,
+}
+
+trait DnsResolver {
+    /// Resolves a domain name for a given resolver_type
+    fn resolve(&mut self, domain: &str, resolve_type: ResolveType) -> Result<DnsEntry>;
+    /// Announces the resolved dns entry for the domain to a resolver
+    fn announce(&mut self, domain: &str, entry: &DnsEntry);
+
+    // name for debugging purposes
+    fn name(&self) -> &'static str;
+}
+
+// impl fmt::Debug for dyn DnsResolver {
+//     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+//         write!(f, "dns resolver")
+//     }
+// }
+
+trait DnsCache {
+    /// Flush all domains
+    fn flush_all(&mut self);
+    /// Flush a single domain
+    fn flush_entry(&mut self, domain: &str);
+}
+
+struct Dns {
+    resolvers: Vec<Box<dyn DnsResolver>>,
+}
+
+impl Dns {
+    pub fn new() -> Self {
+        let mut resolvers: Vec<Box<dyn DnsResolver>> = vec![];
+        resolvers.push(Box::new(cache::CacheResolver::new(1000)));
+        resolvers.push(Box::new(local::LocalTableResolver::new()));
+
+        let opts = remote::RemoteResolverOptions::default();
+        resolvers.push(Box::new(remote::RemoteResolver::new(opts)));
+
+        Dns { resolvers }
+    }
+
+    /// Resolves a domain name to a set of IP addresses based on the resolve_type.
+    /// It can resolve either Ipv4, ipv6 or both addresses.
+    ///
+    /// Each request will be resolved by the resolvers in the order they are added.
+    /// The first resolver is usually the cache resolver, which caches any entries (according to their TTL)
+    /// The second resolver is usually the local table resolver, which resolves any local overrides
+    /// The third resolver is usually the remote resolver, which resolves any remote entries by querying external DNS server(s)
+    ///
+    pub fn resolve(&mut self, domain: &str, resolve_type: ResolveType) -> Result<DnsEntry> {
+        let mut entry = None;
+
+        info!("Resolving {} for {:?}", domain, resolve_type);
+
+        for resolver in self.resolvers.iter_mut() {
+            debug!("Trying resolver: {}", resolver.name());
+
+            if let Ok(e) = resolver.resolve(domain, resolve_type.clone()) {
+                debug!("Found entry {:?}", e);
+                entry = Some(e);
+                break;
+            }
+        }
+
+        if entry.is_none() {
+            return Err(types::Error::DnsDomainNotFound);
+        }
+
+        // Iterate all resolvers and add to all cache systems (normally, this is only the first resolver)
+        for resolver in self.resolvers.iter_mut() {
+            resolver.announce(domain, &entry.clone().unwrap().clone());
+        }
+
+        Ok(entry.unwrap().clone())
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use simple_logger::SimpleLogger;
+    use std::time::Instant;
+
+    #[test]
+    fn resolver() {
+        SimpleLogger::new().init().unwrap();
+
+        let mut dns = Dns::new();
+
+        let now = Instant::now();
+        let e = dns.resolve("example.org", ResolveType::Ipv4).unwrap();
+        let elapsed_time = now.elapsed();
+        e.ipv4().iter().for_each(|x| println!("ipv4: {}", x));
+        println!("Took {} microseconds.", elapsed_time.as_micros());
+
+        let now = Instant::now();
+        let e = dns.resolve("example.org", ResolveType::Ipv6).unwrap();
+        let elapsed_time = now.elapsed();
+        e.ipv6().iter().for_each(|x| println!("ipv6: {}", x));
+        println!("Took {} microseconds.", elapsed_time.as_micros());
+
+        let now = Instant::now();
+        let e = dns.resolve("example.org", ResolveType::Ipv4).unwrap();
+        let elapsed_time = now.elapsed();
+        e.ipv4().iter().for_each(|x| println!("ipv4: {}", x));
+        println!("Took {} microseconds.", elapsed_time.as_micros());
+
+        let now = Instant::now();
+        let e = dns.resolve("example.org", ResolveType::Both).unwrap();
+        let elapsed_time = now.elapsed();
+        e.ipv4().iter().for_each(|x| println!("ipv4: {}", x));
+        e.ipv6().iter().for_each(|x| println!("ipv6: {}", x));
+        println!("Took {} microseconds.", elapsed_time.as_micros());
+    }
+}
--- a/src/dns/cache.rs
+++ b/src/dns/cache.rs
@@ -1,0 +1,165 @@
+use crate::dns::{DnsCache, DnsEntry, DnsResolver, ResolveType};
+use crate::types::Result;
+use log::trace;
+use std::collections::{HashMap, VecDeque};
+
+pub(crate) struct CacheResolver {
+    values: HashMap<String, DnsEntry>,
+    max_entries: usize,
+    lru: VecDeque<String>,
+}
+
+impl DnsResolver for CacheResolver {
+    fn resolve(&mut self, domain: &str, resolve_type: ResolveType) -> Result<DnsEntry> {
+        if let Some(entry) = self.values.get(domain) {
+            if !entry.has_ipv4 && !entry.has_ipv6 && resolve_type == ResolveType::Both {
+                trace!("{}: no addresses found in entry", domain);
+                return Err(crate::types::Error::DnsNoIpAddressFound);
+            }
+            if !entry.has_ipv4 && resolve_type == ResolveType::Ipv4 {
+                trace!("{}: no ipv4 addresses found in entry", domain);
+                return Err(crate::types::Error::DnsNoIpAddressFound);
+            }
+            if !entry.has_ipv6 && resolve_type == ResolveType::Ipv6 {
+                trace!("{}: no ipv6 addresses found in entry", domain);
+                return Err(crate::types::Error::DnsNoIpAddressFound);
+            }
+
+            trace!("{}: found in cache with correct resolve type", domain);
+            self.lru.retain(|x| x != domain);
+            self.lru.push_back(domain.to_string());
+
+            return Ok(entry.clone());
+        }
+
+        Err(crate::types::Error::DnsNoIpAddressFound)
+    }
+
+    /// When a domain is resolved, it will be announced to all resolvers. This cache resolver
+    /// will store it into the cache.
+    fn announce(&mut self, domain: &str, entry: &DnsEntry) {
+        trace!("{}: announcing to cache", domain);
+
+        self.lru.retain(|x| x != domain);
+        self.lru.push_back(domain.to_string());
+
+        if let Some(current_entry) = self.values.get_mut(domain) {
+            trace!("{}: updating existing entry to cache", domain);
+
+            trace!("current entries: {:?}", current_entry.ips);
+            trace!("new entries: {:?}", entry.ips);
+            current_entry.has_ipv4 |= entry.has_ipv4;
+            current_entry.has_ipv6 |= entry.has_ipv6;
+
+            for ip in entry.ips.iter() {
+                if current_entry.ips.iter().any(|x| x == ip) {
+                    continue;
+                }
+                current_entry.ips.push(*ip);
+            }
+            trace!("merged entries: {:?}", current_entry.ips);
+        } else {
+            trace!("adding new entry to cache");
+
+            // Clear out if we have too many entries
+            if self.values.len() >= self.max_entries {
+                if let Some(key) = self.lru.pop_front() {
+                    self.values.remove(&key);
+                }
+            }
+
+            self.values.insert(domain.to_string(), entry.clone());
+        }
+    }
+
+    fn name(&self) -> &'static str {
+        "cache resolver"
+    }
+}
+
+impl DnsCache for CacheResolver {
+    fn flush_all(&mut self) {
+        self.values.clear();
+        self.lru.clear();
+    }
+
+    fn flush_entry(&mut self, domain: &str) {
+        self.values.remove(domain);
+        self.lru.retain(|x| x != domain);
+    }
+}
+
+impl CacheResolver {
+    pub(crate) fn new(max_entries: usize) -> CacheResolver {
+        CacheResolver {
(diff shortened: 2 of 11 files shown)


src/dns.rs
```diff
@@ -0,0 +1,58 @@
...
+use crate::types;
+
+/// A DNS entry is a simple domain name to IP address mapping
+struct DNSEntry {
+    domain_name: String,
+    ip: IpAddr,
+}
+
+/// The main DNS resolver struct
+struct DnsResolver {
```
jaytaph, line 15: This needs to be filled in later.. this pr just does the local resolver

src/dns/local.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// per domain part (e.g. com, org, net, etc.)
+    entries: Vec<DNSEntry>,
+    round_robin_ptrs: HashMap<String, usize>,
+}
+
+impl Default for LocalOverrideTable {
+    fn default() -> LocalOverrideTable {
+        LocalOverrideTable {
+            entries: Vec::new(),
+            round_robin_ptrs: HashMap::new(),
```
jaytaph, line 19: if a domain has multiple ips, we store a round robin pointer so the next call to resolve will return the next ip in the list.

src/dns/local.rs
```diff
@@ -0,0 +1,134 @@
...
+        // For now it's ok to iterate the entries list, but later we should use a more efficient
+        // data structure.
+        for entry in &self.entries {
+            // We separate wildcard matches and specific matches.
+            if WildMatch::new(&entry.domain_name).matches(domain) {
+                wildcard_ips.push(entry.ip);
+            }
+            if entry.domain_name == domain {
+                ips.push(entry.ip);
+            }
```
Sharktheone, line 39: Here I would change it then (when we us a vec in DNSEntry)

```suggestion
            if entry.domain_name == domain {
                ips.append(entry.ips);
                break;
            } else if WildMatch::new(&entry.domain_name).matches(domain) { //when we have a more specific domain name don't use the WildMatch
                 wildcard_ips.append(entry.ips);
                 break;
            }

```

Sharktheone (changes requested):

src/dns.rs
jaytaph: yes, this will be behind a trait

src/dns.rs
```diff
@@ -0,0 +1,58 @@
...
+use crate::types::Result;
+use std::net::IpAddr;
+use crate::dns::local::LocalOverrideTable;
+use crate::types;
+
+/// A DNS entry is a simple domain name to IP address mapping
+struct DNSEntry {
+    domain_name: String,
+    ip: IpAddr,
+}
```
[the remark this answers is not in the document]
jaytaph: The idea is that the resolver always returns a single address (either ipv4 or v6), depending on settings (not sure if these are resolver settings, or user agent settings). If there are multiple ips for a single host, we roundrobin them and this should be done by the resolver.

It is possible to change this into a vec if we find multiple entries, but we should have our round-robin pointer in a dns entry struct as well.

src/dns.rs
```diff
@@ -1,58 +1,85 @@
...
-            }
-        }
+    /// Returns true when this entry has a ipv6 address
+    fn has_ipv6(&self) -> bool {
+        !self.ipv6.is_empty()
+    }
+
+    /// Retrieves the next ipv4 address in the round robin list
+    fn next_ipv4(&mut self) -> IpAddr {
+        let entry = self.ipv4.get(self.rr_ipv4_ptr).expect("Invalid round robin pointer");
```
Sharktheone, line 55: I think, the expect shouldn't be here. This error is recoverable (just ignore the ip ptr), so maybe just use a Result or an Option / "silently" recover the error may be better.

src/dns/local.rs
```diff
@@ -1,134 +1,166 @@
...
-        for entry in &self.entries {
-            // We separate wildcard matches and specific matches.
+        // data structure in the future
+        for (idx, entry) in self.entries.iter().enumerate() {
+            // We separate wildcard matches and specific matches (e.g. *.example.com vs example.com)
             if WildMatch::new(&entry.domain_name).matches(domain) {
-                wildcard_ips.push(entry.ip);
+                matched_wildcard_entries.push(idx);
             }
             if entry.domain_name == domain {
```
Sharktheone, line 42: Maybe put this over the wildcard match because it can optimize the time this a tiny bit. (But feel free to ignore)

src/dns/local.rs
```diff
@@ -1,134 +1,166 @@
...
+            // Entry most specific match
+            matched_entry = Some(self.find_most_specific_match(matched_wildcard_entries))
         }
 
-        // If there is only one ip for a domain, return it
-        if ips.len() == 1 {
-            return Some(*ips.get(0).unwrap());
-        }
+        // Fetch the actual entry in the entries list
+        let found_entry = self.entries.get_mut(matched_entry.expect("No idx found")).expect("No entry found");
```
Sharktheone, line 57: Here again, don't use expect, I mean this function already returns a Result. Maybe it would be nice to implement the functionality for the `?` syntax. If this is not already done, i can do this if you want.

Sharktheone (changes requested):

src/dns.rs
jaytaph: The DNS resolver is now a trait.. I will have a few: the global dns resolver (that uses hickory dns), the local override table resolver and maybe others in the future. The dns components will consist of a global dns resolver, a local dns resolver, and a cache layer.  First, the resolver checks the cache layer. If time expired for the entry, or the entry is not found, the local dns resolver is queried. If there is nothing found, it will check the global dns resolver.

src/dns.rs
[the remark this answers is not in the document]
jaytaph: Changed the dns entry as a whole now

src/dns.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use crate::types::Result;
+use std::net::IpAddr;
+use std::time;
+
+/// A single IP entry in the DNS cache. Each ip entry has an expiration time (based on the TTL)
+/// and a last-used timestamp. When the entry is expired, the entry is found to be stale and should
+/// be refreshed by the resolver. When expired == 0, the entry is considered to be a local override
+/// and should not be refreshed by the resolver.
+/// @todo: Does that information belong here? Or should it be in the cache layer?
```
jaytaph, line 13: as stated in the todo: i'm not sure if both expired and last-used should be part of the ip entry. I think the only place where ttl / expiration is important, would be the caching layer. This means the resolver should return a tuple (ipaddress, ttl) so the layer above can cache it accordingly. Time last used may not be needed.

src/dns.rs
```diff
@@ -0,0 +1,97 @@
...
+    ip: IpAddr,
+    /// Epoch when ip address expires
+    expired: u64,
+    /// Epoch when this entry was last used
+    last_used: u64,
+}
+
+/// A DNS entry is a simple domain name to IP address mapping
+#[derive(Default)]
+struct DNSEntry {
```
jaytaph, line 25: The DNS entry now consists of a domain name, and 2 lists of ip addresses, the ipv4 andipv6. It now contains the roundrobin pointers as well. The whole round robin is encapsulated in the DNSEntry so we can simply do next_ipv4() and not need to worry about it from the outside world.

src/dns.rs
```diff
@@ -0,0 +1,97 @@
...
+        // Update the last used timestamp
+        entry.last_used = time::Instant::now().as_secs();
+
+        entry.ip
+    }
+}
+
+/// Type of DNS resolution
+#[derive(Debug, PartialEq, Clone)]
+pub enum ResolveType {
```
jaytaph, line 84: I don't know exactly who will be in charge of this, but somewhere we must decide if we resolve the ipv4 or ipv6 of a domain. This is the ResolveType we can add to the resolve() function.

src/dns/cache.rs
```diff
@@ -0,0 +1,82 @@
+use std::collections::HashMap;
+use crate::dns::DNSEntry;
+
+struct Cache {
```
jaytaph, line 4: The cache has a fixed size of max_entries elements and uses a lru strategy for evicting old dns entries. It's not yet complete implemented, but the lru seems to work according to the tests

src/dns/local.rs
```diff
@@ -0,0 +1,190 @@
...
+use domain_lookup_tree::DomainLookupTree;
+use std::str::FromStr;
+use std::time;
+
+/// Local override table that can be used instead of using /etc/hosts or similar 3rd party dns system.
+pub struct LocalOverrideTable {
+    /// Entries in the local override table.
+    entries: HashMap<String, DNSEntry>,
+    /// Domaintree is a hierarchical lookup tree for quick scanning of (wildcard) domains
+    tree: DomainLookupTree,
```
jaytaph, line 14: The wildcard matcher has been replaced by the tree lookup crate. It will match deeper wildcards better (*.example.org would match foo.bar.baz.example.org, and we don't want that).

Also, we now do not have to do a difficult process of matching both specific and wildcard matching.

src/dns/local.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        let found_entry = self.entries.get_mut(&domain_entry).expect("No entry found");
+        if resolve_type == ResolveType::Ipv4 && found_entry.has_ipv4() {
+            return Ok(found_entry.next_ipv4());
+        }
+        if resolve_type == ResolveType::Ipv6 && found_entry.has_ipv6() {
+            return Ok(found_entry.next_ipv6());
+        }
+        if resolve_type == ResolveType::Both {
+            // IPv6 gets precedence over IPv4
```
jaytaph, line 54: Could be configurable by  "dns.prefer.ipv6.over.ipv4" maybe

src/dns/local.rs
```diff
@@ -0,0 +1,190 @@
...
+    ///
+    /// if the list is:
+    ///      - .org
+    ///      - .foo.example.org
+    ///      - .example.org
+    ///
+    /// and the domain is: foo.example.org, then the most specific match is foo.example.org
+    ///
+    /// if the domain is test.foo.example.org, .foo.example.org is the best match
+    fn find_most_specific_match(&self, entries: Vec<usize>) -> usize {
```
jaytaph, line 137: we can remove this func

src/types.rs
```diff
@@ -44,8 +44,23 @@ pub enum Error {
     #[error("document task error: {0}")]
     DocumentTask(String),
 
-    #[error("query error: {0}")]
+    #[error("query: generic error: {0}")]
     Query(String),
+
+    #[error("dns: generic error: {0}")]
```
jaytaph, line 50: I like the way we can simply create errors and return them in Result<> , but should we add them all inside a generic enum like this? Wouldn't we want the errors to be per component? (`src/dns/errors.rs`, `src/html5/errors.rs` etc)

src/dns.rs
```diff
@@ -0,0 +1,97 @@
...
+    // domain name
+    domain_name: String,
+    // List of IPv4 addresses for this domain name
+    ipv4: Vec<IpEntry>,
+    // List of IPv6 addresses for this domain name
+    ipv6: Vec<IpEntry>,
+    // Round robin pointer in case we need to round robin
+    rr_ipv4_ptr: usize,
+    // Round robin pointer in case we need to round robin
+    rr_ipv6_ptr: usize,
```
emwalker, line 35: Should this be an enum?  Also, curious why we aren't making use of a simple wrapper around an existing Rust library.  This looks more low-level than I was anticipating.

jaytaph: We are using hickory dns in the remote resolver. I'll draw up a diagram to give you an idea of the architecture (its not too complicated)

src/dns.rs
```diff
@@ -0,0 +1,217 @@
...
+struct DnsEntry {
+    // domain name
+    domain: String,
+    // // Ip type that is stored in this entry (could be Ipv4, IPv6 or Both)
+    // ip_type:   ResolveType,
+    // List of addresses for this domain
+    ips: Vec<IpAddr>,
+
+    /// True when the ips list has ipv4 addresses
+    has_ipv4: bool,
```
jaytaph, line 24: an entry has either resolved ipv4 addresses, resolver ipv6 addresses or both. WHen we have a record in cache, but we want to have ipv6 addresses for that record, we can see if these are already resolved for this record.

src/dns.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// True when the ips list has ipv6 addresses
+    has_ipv6: bool,
+
+    // Internal iterator pointer
+    iter: usize,
+    /// expiry time after epoch
+    expires: u64,
```
jaytaph, line 31: We don't use the expires time yet. This will be taken from a TTL that is returned by the remote resolver.

src/dns/cache.rs
```diff
@@ -0,0 +1,165 @@
...
+
+            return Ok(entry.clone());
+        }
+
+        Err(crate::types::Error::DnsNoIpAddressFound)
+    }
+
+    /// When a domain is resolved, it will be announced to all resolvers. This cache resolver
+    /// will store it into the cache.
+    fn announce(&mut self, domain: &str, entry: &DnsEntry) {
```
jaytaph, line 40: Each resolver has an "announce", which will be called when a new dns entry has been found. It is only used by the cache resolver to store this in cache.

src/types.rs
Kiyoshika: I think it'd make sense to make them component specific now that the list is growing (and will grow more in the future.) I remember this was added maybe a month or two back when there were only 2-3 error types, but now I'd be in favour to move groups of them into their own components

Kiyoshika (approved):

src/dns.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+impl Iterator for DnsEntry {
+    type Item = IpAddr;
+
+    /// With next() you can simply iterate over all the ips in the list
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.iter >= self.ips.len() {
+            // reset iterator at the end
+            self.iter = 0;
```
emwalker, line 82: This cleanup was confusing for me, and now I see what the source of the confusion is.  Iterators should be implemented as separate structs that fall out of scope / are discarded after they're exhausted.  So a cleanup / reset step shouldn't be needed. See `TreeIterator` for an example of this pattern.  If there's a cleanup step, it suggests a new struct needs to be added.  

But in this case, a separate struct for the iterator isn't needed, because you can get an iterator over the items you want through other means:

```patch
diff --git a/src/dns.rs b/src/dns.rs
index 36b6a6d..cb2b42e 100644
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -70,24 +70,9 @@ impl DnsEntry {
     fn ipv6(&self) -> Vec<IpAddr> {
         self.ips.iter().filter(|x| x.is_ipv6()).copied().collect()
     }
-}
-
-impl Iterator for DnsEntry {
-    type Item = IpAddr;
-
-    /// With next() you can simply iterate over all the ips in the list
-    fn next(&mut self) -> Option<Self::Item> {
-        if self.iter >= self.ips.len() {
-            // reset iterator at the end
-            self.iter = 0;
-
-            return None;
-        }
-
-        let ip = self.ips[self.iter];
-        self.iter += 1;
 
-        Some(ip)
+    fn iter(&self) -> impl Iterator<Item = &IpAddr> {
+        self.ips.iter()
     }
 }
```

src/dns.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    // name for debugging purposes
+    fn name(&self) -> &'static str;
+}
+
+// impl fmt::Debug for dyn DnsResolver {
+//     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+//         write!(f, "dns resolver")
+//     }
+// }
```
emwalker, line 119: Dead code.

src/dns.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// The second resolver is usually the local table resolver, which resolves any local overrides
+    /// The third resolver is usually the remote resolver, which resolves any remote entries by querying external DNS server(s)
+    ///
+    pub fn resolve(&mut self, domain: &str, resolve_type: ResolveType) -> Result<DnsEntry> {
+        let mut entry = None;
+
+        info!("Resolving {} for {:?}", domain, resolve_type);
+
+        for resolver in self.resolvers.iter_mut() {
+            debug!("Trying resolver: {}", resolver.name());
```
emwalker, line 158: Nice use of the log crate.

src/dns/remote.rs
```diff
@@ -0,0 +1,113 @@
...
+        match resolve_type {
+            ResolveType::Ipv4 => {
+                ip_types.push("ipv4");
+            }
+            ResolveType::Ipv6 => {
+                ip_types.push("ipv6");
+            }
+            ResolveType::Both => {
+                ip_types.push("ipv6");
+                ip_types.push("ipv4");
```
emwalker, line 31: It strikes me as unusual that we're using a static string here as a type marker.

src/types.rs
emwalker: I'm guessing the rule of thumb on when to create a new error type and when to stick with an existing type is when you want to match against a specific kind of error and not other errors.  So the number of error types becomes the number of specific kinds of error you need to match on in high-level application code.

The matching in error handling code is what is important here.  I can imagine needing a total of three kinds of error in all, or 20, depending on what kind of error handling is needed.

emwalker (approved):
It would be nice if you would update the iterator code I left a note on.  But I also don't want to hold up this PR.  I suspect some of the details will change later.

This code makes me nervous, because the network stack exposes a large surface area for security attacks.  But I can also see why it's needed.  I do wonder whether there's already some canonical Rust crates out there that do some or all of this.  In the Rust ecosystem, semi-official Rust crates (e.g., `serde`) seem to play a bigger role than third-party libraries in other language ecosystems.

src/dns/remote.rs
Sharktheone: I guess we can just use the ResolverType itself and if it is Both we Push `ResolverType::Ipv4` and `ReolverType::Ipv6`

src/dns.rs
```diff
@@ -0,0 +1,217 @@
...
+    Ipv6,
+    /// Resolve both IPV4 and IPV6 addresses
+    Both,
+}
+
+trait DnsResolver {
+    /// Resolves a domain name for a given resolver_type
+    fn resolve(&mut self, domain: &str, resolve_type: ResolveType) -> Result<DnsEntry>;
+    /// Announces the resolved dns entry for the domain to a resolver
+    fn announce(&mut self, domain: &str, entry: &DnsEntry);
```
Sharktheone, line 109: Maybe add a default implementation here, so we only need to override it in the CacheResolver
```suggestion
    fn announce(&mut self, domain: &str, entry: &DnsEntry) {};
```

Sharktheone (approved):
Looks fine to me, just some simple things. But you can ignore them if you want.

src/dns.rs
jaytaph: ah, didn't know this was possible.. very nice

src/dns.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn flush_entry(&mut self, domain: &str);
+}
+
+struct Dns {
+    resolvers: Vec<Box<dyn DnsResolver>>,
+}
+
+impl Dns {
+    pub fn new() -> Self {
+        let mut resolvers: Vec<Box<dyn DnsResolver>> = vec![];
```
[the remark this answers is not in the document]
jaytaph: I would not worry about this too much. It's not used in any thight loops, and the first push will do an allocation for 4 elements anyway. 
So instead of with_capacity(3), you get a with_capacity(4), which is ok enough.

src/dns.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// With next() you can simply iterate over all the ips in the list
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.iter >= self.ips.len() {
+            // reset iterator at the end
+            self.iter = 0;
+
+            return None;
+        }
+
+        let ip = self.ips[self.iter];
```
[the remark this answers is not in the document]
jaytaph: 👍

src/dns/remote.rs
jaytaph: 👍

src/dns.rs
jaytaph: removed

Sharktheone (approved):

src/dns.rs
```diff
@@ -0,0 +1,195 @@
...
+        self.ips.iter().filter(|x| x.is_ipv4()).copied().collect()
+    }
+
+    fn ipv6(&self) -> Vec<IpAddr> {
+        self.ips.iter().filter(|x| x.is_ipv6()).copied().collect()
+    }
+
+    fn iter(&self) -> impl Iterator<Item = &IpAddr> {
+        self.ips.iter()
+    }
```
emwalker, line 76: After I thought about it more, I realized you might have been wanting to use the iterator trait as a state machine for implementing round-robin.  If so, I think something other than `trait Iterator` would be used for this, e.g., a custom method that updates the state to the next ip address, possibly in a new struct that this struct delegates to.

jaytaph: i've actually removed the round robin inside the dns entries because it ended up a bit too complex. I leave it up to the caller to deal with round robin. There might be other ways they want to use the set of resolved ips: when one fails, use another, etc..

PR #267, opened 2023-11-19
Global config store

jaytaph:
The config store is now usuable as a global instance through macros (in the same spirit as you can use warn! info! etc from the logger).

You can use `config!(string, "my.key")` to get the given key or it will return a default value when the key is not found. There is always the possibily to do checks on the actual store:

```rust
   let exist = config::config_store().has("does.this.key.exist")
```

This setup makes it possible for any user agent to set their own configuration store:

```rust
fn main() {
  // Set the config through `config_store_write()` instead of `config_store()`
  config::config_store_write().set_storage(new SqliteStorage("config.db"));

  // do the rest
}
```


src/config.rs
```diff
@@ -8,10 +8,59 @@ use serde_json::Value;
 use std::collections::HashMap;
 use std::mem;
 use std::str::FromStr;
+use std::sync::RwLock;
 use wildmatch::WildMatch;
+use lazy_static::lazy_static;
 
 const SETTINGS_JSON: &str = include_str!("./config/settings.json");
 
+struct GlobalStorage {
```
jaytaph, line 17: I'm not sure what the correct path is here... I have a global storage, which currently holds only the config store (but other things can be in there as well)... but unfortunately, this doesn't work because it is not safe to use.

src/config.rs
```diff
@@ -8,10 +8,59 @@ use serde_json::Value;
...
+    };
+    (sint $key:expr) => {
+        GLOBAL_STORAGE.config_store().get($key).to_sint()
+    };
+    (map $key:expr) => {
+        GLOBAL_STORAGE.config_store().get($key).to_map()
+    };
+}
+
+unsafe impl Send for GlobalStorage {}
```
jaytaph, line 62: This is telling the compiler:   nah, it's allright to use in multi threading...  

but i'm not sure that it is... we want to be able to lock in certain conditions (when writing the json file, or doing db writing)... so we have to fix that from the storage adapters..  when we know for sure that this is safe for multithreading, we somehow have to convince the compiler that this is all good...

we are working around the compiler here, so it might be tricky to get it right..

src/config/settings.rs
```diff
@@ -14,6 +15,92 @@ pub enum Setting {
     Map(Vec<String>),
 }
 
+impl Setting {
+    pub fn to_bool(&self) -> bool {
+        if ! matches!(self, Setting::Bool(_)) {
+            warn!("setting is not a boolean");
+        }
+
+        match self {
```
jaytaph, line 24: even if we have a bool-setting, it may be possible to read it as another type. It's not recommended in most cases, but i can see that it sometimes can be. However, we will always log a warning when we do so.

src/config/settings.rs
```diff
@@ -14,6 +15,92 @@ pub enum Setting {
...
+            Setting::Map(values) => values.clone(),
+            _ => Vec::new(),
+        }
+    }
+}
+
+fn is_bool_value(s: &str) -> bool {
```
jaytaph, line 95: There are multiple string values that can be seen as a bool(true).

src/config.rs
```diff
@@ -194,4 +243,13 @@ mod test {
             Setting::String("wont accept strings".into()),
         );
     }
+
+    #[test]
+    fn find_settings() {
+        let enabled = config!(bool "dns.local_resolver.enabled");
```
jaytaph, line 249: this is ultimately what i want to do:   have a single macro that allows us to fetch a value from the config store, whereever it may be..   Since it's the user-agent that initializes the config store, we must have some global place where this information can be found.. we can't init our own config store within the engine.

We therefor must have a place where the engine always can find the store.. it sounds like a singleton (not a big fan though).. but we can't pass a config store to each deep function just in case we need it somewhere.

src/config.rs
```diff
@@ -8,10 +8,59 @@ use serde_json::Value;
...
+lazy_static! {
+    static ref GLOBAL_STORAGE: RwLock<GlobalStorage> = RwLock::new(GlobalStorage {
+        config_store: None,
+    });
+}
+
+#[macro_export]
+macro_rules! config {
+    (string $key:expr) => {
+        let lock = GLOBAL_STORAGE.config_store().read();
```
jaytaph, line 43: acquire a read lock, do a read, and drop the lock. This will work with a rwlock:  multiple read locks can be acquired, but only one write lock which blocks other locks.

src/config.rs
emwalker: Yeah, needing `unsafe` feels strange in this case.  I think you can instantiate it some place under `main` and then

1. Copy it once to each thread, assuming it won't change
2. Wrap it in an `Arc` so that it is `Send`

src/config.rs
emwalker: To lock resources against shared access across different threads, you'd normally gate the resource behind an `Arc` / `Mutex`.  [Channels](https://doc.rust-lang.org/std/sync/mpsc/index.html) might also be useful here, depending on the specific thing being done.

src/config.rs
```diff
@@ -1,17 +1,128 @@
 pub mod settings;
 pub mod storage;
 
+use std::cell::Cell;
 use crate::config::settings::{Setting, SettingInfo};
 use crate::types::Result;
```
Sharktheone, line 6: This is your problem, you import the "own" Result enum and not the one from the rust stdlib.

src/bin/config-store.rs
```diff
@@ -1,10 +1,11 @@
...
 use clap::{Parser, Subcommand};
 use derive_more::Display;
+use gosub_engine::config;
 use gosub_engine::config::settings::Setting;
-use gosub_engine::config::storage::json_storage::JsonStorageAdapter;
-use gosub_engine::config::storage::sqlite_storage::SqliteStorageAdapter;
-use gosub_engine::config::{ConfigStore, Store};
+use gosub_engine::config::storage::json::JsonStorageAdapter;
+use gosub_engine::config::storage::sqlite::SqliteStorageAdapter;
+use gosub_engine::config::StorageAdapter;
```
jaytaph, line 8: The name "store" was a bit ambiguous. They are actually storage adapters since we already have a config store.

src/config.rs
```diff
@@ -2,16 +2,118 @@ pub mod settings;
...
 use wildmatch::WildMatch;
 
+/// Settings are stored in a json file, but this is included in the binary for mostly easy editting.
 const SETTINGS_JSON: &str = include_str!("./config/settings.json");
 
+/// StoreAdapter is the interface for storing and retrieving settings
+/// This can be used to storage settings in a database, json file, etc
+/// Note that we need to implement Send so we can send the storage adapter
+/// to other threads.
+pub trait StorageAdapter: Send + Sync {
```
jaytaph, line 23: Send + Sync are important here, otherwise we cannot share them with other threads. We must make sure the adapters are thread-aware.

src/config.rs
```diff
@@ -2,16 +2,118 @@ pub mod settings;
...
+    /// can get mutable.
+    fn set(&self, key: &str, value: Setting);
+
+    /// Retrieves all the settings in the storage in one go. This is used for preloading the settings
+    /// into the ConfigStore and is more performant normally than calling get_setting manually for each
+    /// setting.
+    fn all(&self) -> crate::types::Result<HashMap<String, Setting>>;
+}
+
+lazy_static! {
```
jaytaph, line 38: This is static reference, but through `lazy_static!` we can actually initialize it just like regular variables.

src/config.rs
```diff
@@ -2,16 +2,118 @@ pub mod settings;
...
+    // Initial config store will have a memory storage adapter. It will save within the session, but not
+    // persist this on disk.
+    static ref CONFIG_STORE: RwLock<ConfigStore> = RwLock::new(ConfigStore::default());
```
jaytaph, line 41: RwLock since we can have multiple reads (for instance, https://github.com/gosub-browser/gosub-engine/pull/267/files#diff-d55130faaf1118028faf388a306896363c166c769c0bab6bdff2b23051659575R101)

src/config.rs
```diff
@@ -2,16 +2,118 @@ pub mod settings;
...
+}
+
+/// Returns a reference to the config store, which is locked by a mutex.
+/// Any callers of the config store can just do  config::config_store().get("dns.local_resolver.enabled")
+pub fn config_store() -> std::sync::RwLockReadGuard<'static, ConfigStore> {
+    CONFIG_STORE.read().unwrap()
```
jaytaph, line 47: this returns a regular read locked store

src/config.rs
```diff
@@ -21,58 +123,59 @@ struct JsonEntry {
...
-
-/// Configuration store is the place where the gosub engine can find all configurable options
+/// Configuration storage is the place where the gosub engine can find all configurable options
 pub struct ConfigStore {
     /// A hashmap of all settings so we can search o(1) time
-    settings: HashMap<String, Setting>,
+    /// The mutex allows to share between multiple threads,
+    /// The refcell allows us to use mutable references in a non-mutable way (ie: settings can be
+    /// stored while doing a immutable get())
+    settings: std::sync::Mutex<std::cell::RefCell<HashMap<String, Setting>>>,
```
jaytaph, line 132: A mutex to share the config store between threads
A RefCell to allow us to mutate from a immutable config store

src/config.rs
```diff
@@ -21,58 +123,59 @@ struct JsonEntry {
...
+        self.storage = storage;
 
-        // preload the settings if requested
-        if preload {
-            let all_settings = store.storage.get_all_settings()?;
+        // Find all keys, and add them to the configuration store
+        if let Ok(all_settings) = self.storage.all() {
             for (key, value) in all_settings {
-                store.settings.insert(key, value);
+                self.settings
```
jaytaph, line 167: lock the mutex,
unwrap the mutex guard,
borrow the refcel as mutable reference
and insert the key/value into the hashmap

src/config.rs
```diff
@@ -97,46 +200,60 @@ impl ConfigStore {
...
     /// Returns the setting with the given key. If the setting is not found in the current
-    /// store, it will load the key from the storage. If the key is still not found, it will
+    /// storage, it will load the key from the storage. If the key is still not found, it will
     /// return the default value for the given key. Note that if the key is not found and no
     /// default value is specified, this function will panic.
-    pub fn get(&mut self, key: &str) -> Setting {
+    pub fn get(&self, key: &str) -> Option<Setting> {
         if !self.has(key) {
-            panic!("Setting {} not found", key);
+            return None;
```
jaytaph, line 208: This was a panic, but we just return None.. it's up to the caller to deal with it

src/config.rs
```diff
@@ -97,46 +200,60 @@ impl ConfigStore {
...
-            return setting.clone();
+        if let Some(setting) = self.settings.lock().unwrap().borrow().get(key) {
+            return Some(setting.clone());
         }
 
         // Setting not found, try and load it from the storage adapter
-        if let Some(setting) = self.storage.get_setting(key) {
-            self.settings.insert(key.to_string(), setting.clone());
-            return setting.clone();
+        if let Some(setting) = self.storage.get(key) {
```
jaytaph, line 216: If the key is not found in the current settings, we load it from the storage, and save it in the settings for the next time.

src/config/storage/json.rs
```diff
@@ -1,25 +1,24 @@
...
 use std::fs::File;
 use std::io::{Read, Seek, Write};
-use std::sync::{Arc, Mutex};
+use std::sync::Mutex;
 
 pub struct JsonStorageAdapter {
     path: String,
-    file_mutex: Arc<Mutex<File>>,
-    elements: HashMap<String, Setting>,
+    elements: Mutex<HashMap<String, Setting>>,
```
jaytaph, line 14: Mutex to guard against multiple threads. This needs to be fixed better when saving the json

src/config/storage/json.rs
```diff
@@ -53,38 +51,45 @@ impl TryFrom<&String> for JsonStorageAdapter {
     }
 }
 
-impl Store for JsonStorageAdapter {
-    fn get_setting(&self, key: &str) -> Option<Setting> {
-        self.elements.get(key).cloned()
+impl StorageAdapter for JsonStorageAdapter {
+    fn get(&self, key: &str) -> Option<Setting> {
+        let lock = self.elements.lock().unwrap();
```
jaytaph, line 56: We read during a lock. This is fast enough. I don't even know if we really need it here.

src/bin/config-store.rs
jaytaph: @Sharktheone Can you guve an example on what you mean?

src/config.rs
jaytaph: This might indeed be false. I'll check

src/bin/config-store.rs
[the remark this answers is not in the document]
jaytaph: I personally like dashes in binaries and options :)

src/config.rs
```diff
@@ -2,16 +2,118 @@ pub mod settings;
...
+    CONFIG_STORE.write().unwrap()
+}
+
+/// These macro's can be used to simplify the calls to the config store. You can simply do:
+///
+/// let enabled = config!(bool "dns.local_resolver.enabled").unwrap();
+/// config_set!(bool "dns.local_resolver.enabled", false);
+///
+/// Note that when you cannot find the key, it will return a default value. This is not always
+/// what you want, but you can test for existence of the key with config_store().has("key")
```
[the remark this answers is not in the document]
jaytaph: If the config key is not found, we have a default value in the json configuration. There might be a chance that we will be using user defined keys later on (for instance for plugins).

src/bin/config-store.rs
Sharktheone: I mean, doing this in `storage.rs`

```rust
mod json;
mod memory;
mod sqlite;

pub use json::*;
pub use memory::*;
pub use sqlite::*;
```

Then you can do this
```rust
use gosub_engine::config::storage::*
```

and still use `JsonStorageAdapter` and `SqliteStorageAdapter`. This makes things cleaner. Most of the time when you import one of the storage atapters, you most likely need all.

src/config.rs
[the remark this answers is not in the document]
Sharktheone: Ah, yeah... then it makes definitely sense to not do this.

src/config.rs
```diff
@@ -2,16 +2,118 @@ pub mod settings;
...
+    };
+    (map $key:expr) => {
+        match crate::config::config_store().get($key) {
+            Some(setting) => setting.to_map(),
+            None => std::collections::HashMap::new(),
+        }
+    };
+}
+
+#[allow(clippy::crate_in_macro_def)]
```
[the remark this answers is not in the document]
jaytaph: no, it won't work. It gives an unknown crate issue. I'll leave it for now

src/config.rs
jaytaph: fixed this. This flow was not correct after some refactoring here and there.. surprised it still passed the tests

src/config.rs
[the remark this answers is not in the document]
jaytaph: you need to unwrap the guard before you can use it. I'm open for suggestions on better ways

src/config.rs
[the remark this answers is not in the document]
jaytaph: this flow was a bit off. This should be fixed now:

get:
  - if key exists in store: return key
  - if key exists in storage: save key in store and return key
  - if key info exists: return default value from info

set:
  - if key info exist: set in store and set in storage (persist)

Later on, we can register keys to the configuration by registering an settings_info struct. From that point on, it should work exactly like it works with our "global" configuration keys.  I'm not sure if this works the way i envision it, but it allows plugins to actually register their own keys  (only in the user.* or plugin.* namespace, for instance)

Maybe we can do dynamic keys as well.. but i'm not 100% sure how that would work

src/config/settings.rs
```diff
@@ -14,6 +15,93 @@ pub enum Setting {
...
+    }
+
+    pub fn to_map(&self) -> Vec<String> {
+        if !matches!(self, Setting::Map(_)) {
+            warn!("setting is not a map");
+        }
+
+        match self {
+            Setting::Map(values) => values.clone(),
+            _ => Vec::new(),
```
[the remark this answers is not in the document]
jaytaph: yeah, makes sense I guess

src/config.rs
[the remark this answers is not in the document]
Sharktheone: Maybe return a `Result` which you can just use the `?` on in functions that also return a `Result`. Replacing it with a `expect` is also already better

src/config/settings.rs
```diff
@@ -155,7 +155,7 @@ impl FromStr for Setting {
 
     /// Converts a string to a setting or None when the string is invalid
     fn from_str(key: &str) -> Result<Setting> {
-        let (key_type, key_value) = key.split_once(':').unwrap();
+        let (key_type, key_value) = key.split_once(':').expect("");
```
Sharktheone, line 158: something that says, what has failed in the expect would be nice, else it is more or less the same as `unwrap`

src/dns.rs
```diff
@@ -164,10 +164,7 @@ mod test {
     #[test]
     fn resolver() {
         // Add simple logger, if not possible, that's fine too
-        match SimpleLogger::new().init() {
-            Ok(_) => {}
-            Err(_) => {}
-        }
+        let _ = SimpleLogger::new().init();
```
Sharktheone, line 167: When i think about this, I'm not sure if we want to to this here. It maybe would make sense to use the `SimpleLogger` instead of our own and invoke the logger init somewhere in the main fn. The other option I'd see, to upgrade our logger and use this instead. I guess it makes sense to only have one logger.

Sharktheone (approved):

src/config/settings.rs
```diff
@@ -14,6 +15,93 @@ pub enum Setting {
...
+    pub fn to_sint(&self) -> isize {
+        if !matches!(self, Setting::SInt(_)) {
+            warn!("setting is not an signed integer");
+        }
+
+        match self {
+            Setting::SInt(value) => *value,
+            Setting::UInt(value) => *value as isize,
+            Setting::Bool(value) => *value as isize,
+            Setting::String(value) => is_bool_value(value) as isize,
```
Kiyoshika, line 42: For `to_sint` and `to_uint`, should the string convert to the actual integer representation or leave it as bool here?

If there was a setting `s:123` stored as a string and we tried converting it to its integer representation then it would return 1 which may be unexpected.

I'm not sure what cases if/when that would ever come up, but something to consider

Kiyoshika: Of course, we shouldn't store integer settings as strings, but.. :P

src/config.rs
```diff
@@ -97,46 +200,65 @@ impl ConfigStore {
...
+            }
+        };
+
         if mem::discriminant(&info.default) != mem::discriminant(&value) {
-            panic!("value is of different type than setting expects")
+            warn!(
+                "config: Setting {} is of different type than setting expects",
+                key
+            );
+            return;
```
Sharktheone, line 248: I guess we could try to convert the setting here to the expected type. (e.g int to string)

src/config/settings.rs
Sharktheone: Makes sense to me, to do that.

Sharktheone (approved):
LGTM. The only thing i noticed is the `unwrap`s but i think they are okay for now. I also don't really have a better solution that is convenient. I might look into that at some point. I maybe add a todo on tasks project

Kiyoshika (approved):

src/bin/config-store.rs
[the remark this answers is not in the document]
emwalker: Same here — I prefer dashes to underscores in binary executable names.  E.g., in Linux, the executable name is `google-chrome`.  (For binary executables, even better is a short name with no dashes or underscores.)

PR #107, opened 2023-10-07
replace older mod.rs style with newer <module>.rs style

Sharktheone:
continuing from #102, since I pushed to main... 🤦

PR #268, opened 2023-11-19
Improving Code Idiomacy

koopa1338:
This pull request focuses on enhancing the idiomatic expression within the Rust codebase. By adhering more closely to Rust's best practices and conventions, this update aims to improve readability, maintainability, and performance where applicable.

#### Changes Made
1. *Iterator Usage*: Utilized iterators and functional programming paradigms in certain loops, enhancing code clarity.
2. *Options*: Utilize methods of options to avoid unnecessary if-else
3. *Constructors*: Removed empty constructors and replaced them with their `Default` implementation
4. *Traits*: Where possible, used `impl Trait`s in function arguments and replaced methods with their semantic trait implementation
5. *Derives*: Sorted derives and add `Default` where possible
6. *Modules*: Replaced named module files that contained only module declaration with corresponding `mod.rs` files


Sharktheone:
For point 6: We have moved from the mod.rs syntax in #107. It can get quite messy and chaotic when having 100 files called mod.rs, especially when you have multiple of them open in your editor.

src/testing/tree_construction/fixture.rs
```diff
@@ -15,37 +15,35 @@ pub struct FixtureFile {
 }
 
 /// Reads a given test file and extract all test data
-pub fn read_fixture_from_path(path: &PathBuf) -> Result<FixtureFile, Error> {
-    let input = fs::read_to_string(path)?;
-    let path = path.to_string_lossy().into_owned();
+pub fn read_fixture_from_path(path: impl AsRef<Path>) -> Result<FixtureFile, Error> {
```
Kiyoshika, line 18: Not that I'm disagreeing, but for my own knowledge (still new to rust) - what's the benefit of using `AsRef` here versus a "regular" `&` reference?

Kiyoshika: Read through the rust docs, I guess it's for anything that can be converted to a `PathBuf` reference. I'm not familiar enough with this portion of the code to know if that's done, but I don't see the harm

koopa1338: > Read through the rust docs, I guess it's for anything that can be converted to a PathBuf reference. I'm not familiar enough with this portion of the code to know if that's done, but I don't see the harm

Exactly, I guess it isn't necessary at this point, I'm also not familiar with the code, so I only "_fixed_" things in a way that is common in Rust and I would also do myself. In this case it makes the api more flexible for future usage.

Sharktheone:
It breaks many of the tests tough, and you may consider running `make format` and fixing some cargo clippy warnings

koopa1338:
yeah it's still wip, I will make sure that all tests run again and the ci is happy :wink:

jaytaph:
@koopa1338 sorry for the delay. Didn't see it was still pending. If we have approval, we can merge the PR. If you like, you can add your name to the AUTHORS file as well.

koopa1338:
No worries, I added myself to the authors file :wink:

Kiyoshika:
I looked through a good chunk of this earlier, will finish reviewing hopefully soon, trying to wrap up some of the bindings real quick

Kiyoshika:
@jaytaph do you want the commits squashed prior to merging? I know lately we usually stick to 1-2 commits per PR

src/html5/parser.rs
```diff
@@ -3667,7 +3667,7 @@ impl<'chars> Html5Parser<'chars> {
                     let new_name = SVG_ADJUSTMENTS_ATTRIBUTES
                         .get(name)
                         .expect("svg adjustments");
-                    new_attributes.insert(new_name.to_string(), value.clone());
+                    new_attributes.insert((*new_name).to_owned(), value.clone());
```
Sharktheone, line 3670: This seems a bit odd, i mean defer the element and then call `to_owned` if you could just call `to_string`

Sharktheone (commented):
I like most of the changes, especially adding the "missing" semicolons. But i really don't like the things with `default`. I personally prefer `new` over `default`.

Sharktheone:
And can you maybe look at #107/#102 where we removed all the `mod.rs` files.

emwalker (changes requested):
@koopa1338 I like about 80 percent of the changes in this PR, and I can live with all of them.  My main reservation is the size of the PR; in the future, perhaps tackling one thing at a time will make it easier to review.

Some of the changes are a little arbitrary, e.g., `to_string() -> to_owned()`, but I don't mind them, even if that's the case.  This kind of update is a good way of familiarizing oneself with the code base.

For now, my own thought is to merge what you've done here.  But first:

1. There are merge conflicts that will need to be manually resolved
1. Once the conflicts are resolved, you'll want to squash all commits into a single commit and force push

Sharktheone:
Holy, this screenshot is like a flashbang... I can also live with all of the changes, but as I said above, I don't really like the "renaming" all `new` functions to `default`.

jaytaph:
As far as I can tell, one of the benefits of having "Default" instead of New (considering new() would not have any arguments), is that it's easier to use them in scenario's like "unwrap_or_default() and such.

Sharktheone:
I mean you can have both `new` and `default`, you can just call `new` in default

emwalker:
I personally don't mind a `new()`, and I think `new()` can even delegate to `default()`:

```rust
impl Something {
  fn new() -> Self {
    Self::default()
  }
}
```

My own thought on the changes in this PR are that they're a vague improvement, because we're getting `Default` defined, and we're using it in a bunch of places where it makes sense.  And the PR allows another developer to learn the code.  We can always backfill `new()` back selectively as we go when it makes sense.  But I'm just one developer here, so I'm happy to go with whatever other people think.

src/html5/parser.rs
koopa1338: Yeah, I can see that this is kind of weird, It would be fine to revert this to `to_string`. The deref is necessary because we get a `&&str` out of the hashmap. Another option that might be better to read would be this:
```rust
let &new_name = SVG_ADJUSTMENTS_ATTRIBUTES
    .get(name)
    .expect("svg adjustments");
new_attributes.insert(new_name.to_owned(), value.clone());
```
What do you think?

koopa1338:
> For now, my own thought is to merge what you've done here. But first:
>
> 1. There are merge conflicts that will need to be manually resolved
> 2. Once the conflicts are resolved, you'll want to squash all commits into a single commit and force push

I would do another rebase, squash my commits and fix merge conflicts along the way.

@emwalker I was a bit unsure about the whole `default` and `new` thing, in general I thought that a `new` constructor without parameters could be replaced with a call to default, however there also some use cases where I see that a `new` constructor makes sense semantically, like for arenas or buffers.

I've stumbled on [this clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/new_without_default) where the `fix` section seems like a sane alternative and if everyone is ok with that I would change it to that style.

emwalker:
> I would do another rebase, squash my commits and fix merge conflicts along the way.

Judging from experience, your rebase might be easier if you squash your commits first so that you only have one or two commits to rebase off of latest `main`.  But I'm sure you'll figure something out.

> I've stumbled on [this clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/new_without_default) where the fix section seems like a sane alternative and if everyone is ok with that I would change it to that style.

Feel free to change some of those `::default()` calls back to `::new()`.  On the off chance you were thinking of updating the linting rules, I'd skip that for this PR.

src/html5/node.rs
```diff
@@ -235,86 +237,96 @@ impl Clone for Node {
...
+            parent,
+            children,
             data: NodeData::Document(DocumentData::new()),
-            name: "".to_string(),
-            namespace: None,
+            name,
+            namespace,
             document: Document::clone(document),
-            is_registered: false,
+            is_registered,
```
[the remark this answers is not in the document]
koopa1338: I looked into it, and there are some issue with the default implementation that I couldn't solve, for example I would have to use a default for the `NodeData` and neither of the variants looks like a good fit for a default value imo. Any suggestions?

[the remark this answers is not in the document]
Sharktheone: I guess, you could use a comment as a "Default" NodeData. But Document could also make sense, because it's an empty struct. I guess it's up to you, if no one as a other opinion on this.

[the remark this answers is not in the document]
koopa1338: I would skip that for now, the PR is already big enough and I guess there might be more discussion about that, so I will open an Issue regarding this.

jaytaph:
@emwalker @Sharktheone @Kiyoshika CI is happy.. feel free to merge

src/render_tree/properties.rs
```diff
@@ -10,19 +10,20 @@ pub struct Rectangle {
 
 impl Default for Rectangle {
     fn default() -> Self {
+        Self::new()
+    }
```
Sharktheone, line 14: Couldn't you use `#[derive(Default)]` here because the default rectangle uses zeros for the new method?

Sharktheone (changes requested):
Only some small changes, I might have missed last time.

Sharktheone:
> @emwalker @Sharktheone @Kiyoshika CI is happy.. feel free to merge

Yup, LGTM. I only found one small thing, but it wouldn't be problematic if it wouldn't be changed.

src/render_tree/properties.rs
koopa1338: I guess I could, I just wanted to be consistent with the other default implementations. A pro imo is that, if you change the `new` constructor to accept parameters, then you are forced to reason about what the default implementation should do, or you even need one. If we derive the trait, then this connection is lost and this could lead to surprises in scenarios where you called `default` instead of `new`.

emwalker (changes requested):
> @emwalker @Sharktheone @Kiyoshika CI is happy.. feel free to merge

We still need to rebase into a single commit.  @jaytaph I think you still don't like using the "Squash and merge" button?  Please advise.  We want atomic commits that will make using `git bisect` unproblematic.

@koopa1338 can you rebase this PR into a single commit and force push the result back to your branch?

koopa1338:
Sorry for that, didn't know that the squash and commit is disabled, squashed it all in a single commit, should be fine now.

emwalker (approved):
Congratulations, @koopa1338, on your first Gosub PR.  Many thanks for the help.

PR #272, opened 2023-11-20
first progress on render tree

Kiyoshika:
EDIT:
This PR depends on #275 being merged into this branch before this is merged into main. Big thanks to @Sharktheone for cleaning up lots of redundant duplication.

This is the introduction to the `RenderTree`. It is far, far from complete, but it at least helps with rendering pure text which is working in my proof of concept shown below (note that some calculations are still slightly off but it's very close)

image

It's a separate tree with a separate iterator and separate nodes that contain visual information (margins, fonts, sizes, position, etc.) A render tree is build from a `DocumentHandle` and is traversed internally to build the render tree with `RenderTree.build()`. Side note, I also want to add something similar to what @jaytaph was talking about in chat earlier, an "Engine API" that the user agent communicates with the engine easily. Right now in my UA I am instantiating the `CharIterator`, `DocumentBuilder` and `Html5Parser` and would be better to do something like `Engine.parse_html("...")` to pass raw html and return a `RenderTree`. That's probably a natural next step now that this tree is at least minimally functional

There are likely going to be some slight design changes as we start needing to do more complex calculations for the layout, but right now, for rendering pure text, this seems like a decent enough first step.

We've already got some benefits since this significantly simplified my user agent code.

### DIFF
--- a/src/render_tree.rs
+++ b/src/render_tree.rs
@@ -1,0 +1,436 @@
+use std::borrow::BorrowMut;
+use std::{cell::RefCell, rc::Rc};
+
+use crate::html5::node::NodeData;
+use crate::html5::parser::document;
+use crate::html5::parser::document::{Document, DocumentHandle};
+use crate::render_tree::{properties::Rectangle, text::TextNode};
+
+pub mod properties;
+pub mod text;
+pub mod util;
+
+/// The position of the render cursor used to determine where
+/// to draw an object
+#[derive(Debug, PartialEq)]
+pub struct Position {
+    pub x: f64,
+    pub y: f64,
+}
+
+impl Position {
+    pub fn new() -> Self {
+        Self { x: 0., y: 0. }
+    }
+
+    // TODO: might be more idiomatic to use From trait for this?
+    pub fn new_from_existing(position: &Position) -> Self {
+        Self {
+            x: position.x,
+            y: position.y,
+        }
+    }
+
+    /// Move position to (x, y)
+    pub fn move_to(&mut self, x: f64, y: f64) {
+        self.x = x;
+        self.y = y;
+    }
+
+    /// Move position relative to another position.
+    /// x = relative.x + x_offset
+    /// y = relative.y + y_offset
+    pub fn move_relative_to(&mut self, relative_position: &Position, x_offset: f64, y_offset: f64) {
+        self.x = relative_position.x + x_offset;
+        self.y = relative_position.y + y_offset;
+    }
+
+    /// Adjust y by an offset.
+    /// y += offset_y
+    pub fn offset_y(&mut self, offset_y: f64) {
+        self.y += offset_y;
+    }
+
+    /// Adjust x by an offset.
+    /// x += offset_x
+    pub fn offset_x(&mut self, offset_x: f64) {
+        self.x += offset_x;
+    }
+}
+
+impl Default for Position {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// A RenderTree is a data structure to be consumed by a user agent
+/// that combines the DOM and CSSOM to compute layouts and styles
+/// for objects to draw on the screen.
+pub struct RenderTree {
+    /// Pointer to the underlying document that builds this render tree
+    // TODO: Add CSSOM here as well when we get to that point
+    document: DocumentHandle,
+    /// Entry point of the render tree
+    // TODO: make this a NodeHandle to make operations easier and not
+    // have to keep doing borrow(), borrow_mut() etc.
+    pub root: Rc<RefCell<Node>>,
+    /// Global render cursor position managed by the tree
+    /// that determines where to draw objects on the screen.
+    /// Whenever a node is created, the current value of the
+    /// position is copied into the node.
+    pub position: Position,
+}
+
+impl RenderTree {
+    pub fn new(document: &DocumentHandle) -> Self {
+        Self {
+            document: Document::clone(document),
+            root: Rc::new(RefCell::new(Node::new())),
+            position: Position::new(),
+        }
+    }
+
+    pub fn build(&mut self) {
+        // start with a clean root (if build is called multiple times)
+        self.root = Rc::new(RefCell::new(Node::new()));
+
+        let tree_iterator = document::TreeIterator::new(&self.document);
+        let mut reference_element = Rc::clone(&self.root);
+
+        for current_node_id in tree_iterator {
+            let doc_read = self.document.get();
+            if let Some(current_node) = doc_read.get_node_by_id(current_node_id) {
+                match &current_node.data {
+                    NodeData::Element(element) => {
+                        let new_node = match element.name.as_str() {
+                            "h1" => Node::new_heading1,
+                            "h2" => Node::new_heading2,
+                            "h3" => Node::new_heading3,
+                            "h4" => Node::new_heading4,
+                            "h5" => Node::new_heading5,
+                            "h6" => Node::new_heading6,
+                            "p" => Node::new_paragraph,
+                            _ => continue, /* add more here, for now ignore */
+                        }(&mut self.position);
+
+                        let new_node = Rc::new(RefCell::new(new_node));
+
+                        util::add_node_to_element(&reference_element, &new_node);
+                        reference_element = Rc::clone(&new_node);
+                    }
+                    NodeData::Text(text) => {
+                        let mut mut_element_ref = reference_element.as_ref().borrow_mut();
+                        if let NodeType::Text(element_text) = &mut mut_element_ref.node_type {
+                            element_text.value.push_str(text.value());
+                        }
+                    }
+                    _ => { /* ignore */ }
+                }
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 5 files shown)


src/render_tree/text.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub font: String,
+    pub font_size: f64,
+    pub is_bold: bool,
+    // TODO: color, styles, visiblity, etc
+}
+
+impl TextNode {
+    /*
+        NOTE: I got the default font sizes from https://stackoverflow.com/a/70720104
+    */
```
Kiyoshika, line 17: ah, the pain of comments. Originally this was true but now it's not...

src/render_tree.rs
```diff
@@ -0,0 +1,563 @@
...
+        Self {
+            document: Document::clone(document),
+            root: Rc::new(RefCell::new(Node::new())),
+            position: Position::new(),
+        }
+    }
+
+    pub fn build(&mut self) {
+        // start with a clean root (if build is called multiple times)
+        self.root = Rc::new(RefCell::new(Node::new()));
```
Kiyoshika, line 96: something I am not sure of, if we replace the root like this, are all the children of the (previous) tree deallocated? Does this cause a leak? Checking for leaks on MacOS is terrible compared to linux (I miss valgrind...)

src/render_tree.rs
```diff
@@ -0,0 +1,563 @@
...
+/// A RenderTree Node mimics a Node from the DOM but
+/// contains more visual information such as width,
+/// height, font sizes, colors, etc. A RenderTree Node
+/// can have children just like regular DOM nodes.
+#[derive(Debug, PartialEq)]
+pub struct Node {
+    pub node_type: NodeType,
+    // TODO: set default margin/padding for all headings/paragraph
+    // items. Maybe we need a Node::new_heading1() which is a wrapper
+    // around Text::new_heading1() to set the additional information
```
Kiyoshika, line 175: outdated comment, reminder for myself to remove

src/render_tree.rs
```diff
@@ -0,0 +1,563 @@
...
+            margin: Rectangle::with_values(margin, 0., 0., margin),
+            padding: Rectangle::new(),
+            parent: None,
+            next_sibling: None,
+            children: Vec::new(),
+            position: Position::new_from_existing(position),
+        };
+
+        position.offset_y(font_size);
+        position.offset_y(margin);
```
Kiyoshika, line 219: can probably just reduce this to `position.offset_y(font_size + margin)`

jaytaph (approved):

Kiyoshika:
> I'm ok with this for now as this shows great progress. But we should try and follow the common way of rendering:
> 
> * step 1: generate render tree: walk the document tree and check together with the CSS which of the nodes are visible and which are not. (for now we can only check node types i guess). This results in the render-tree where all CSS properties of a node is calculated.
> * step 2: generate box layout from render tree:  as we have the properties, we can create rectangles on the screen where things should be placed. These are not filled in yet.
> * step 3: paint: render the content of the boxes. This is a large task that can be outsourced to a GPU.
> 
> This is where we want to go, but obviously we need a few steps before that (calculate CSS properties from the CCSOM).. I hope @neuodev can help us with this

yeah as of right now I haven’t touched anything CSS related since I wanted to utilize the CSSOM (when we have it) instead of using an elements attributes and manually trying to do it. This would handle visibility of nodes as well as all other styling properties

the box layout definitely needs to be done, right now it’s a pretty “raw” layout by using margin and font size to place the items but at the very least I wanted to get something working on the UA. I need to do some reading this week on the box layout to see how I want to build that in

Sharktheone:
I hopefully can finish my review later... I have some other tasks currently.

src/render_tree.rs
```diff
@@ -0,0 +1,563 @@
...
+                                reference_element = Rc::clone(&new_node);
+                            }
+                            "p" => {
+                                let new_node =
+                                    Rc::new(RefCell::new(Node::new_paragraph(&mut self.position)));
+                                util::add_text_node(&reference_element, &new_node);
+                                reference_element = Rc::clone(&new_node);
+                            }
+                            _ => { /* add more here, for now ignore */ }
+                        }
```
Sharktheone, line 150: You have repeating code here, this can be improved significantly. I have something ready, I can commit it if you want.

src/render_tree.rs
```diff
@@ -0,0 +1,563 @@
...
+            next_sibling: None,
+            children: Vec::new(),
+            position: Position::new_from_existing(position),
+        };
+
+        position.offset_y(font_size);
+        position.offset_y(margin);
+
+        new_node
+    }
```
Sharktheone, line 337: I think, here it is also possible to get rid of huge chunks of the code because it's repeating very much. I can also commit something for this here if you want.

src/render_tree/text.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    pub fn new_paragraph() -> Self {
+        Self {
+            value: "".to_owned(),
+            font: DEFAULT_FONT.to_owned(),
+            font_size: 18.5,
+            is_bold: false,
+        }
+    }
```
Sharktheone, line 79: same for here, I would think about a function like `new_text` and `new_headingX` calls into it and passes the font_size to it.

Sharktheone (commented):
For some of them I have something ready because i needed to test it. I can commit it.

src/render_tree.rs
Kiyoshika: can you commit into a separate branch? that way if we decide not to go forward we doubt have to deal with reverting commits. Otherwise we could then commit into this branch and rebase

src/render_tree.rs
Kiyoshika: You can put this under the same branch mentioned above

src/render_tree/text.rs
Sharktheone: I meant, you define the fontsize in the function, we can have a functions `new_headingX` and `new_paragraph` that calls into a (private) function `new_text`, which wants the font size (this is defined in the public function. It would look like this then:


```rust
pub fn new_paragraph() -> {
   Self::new_text(18.5)
}
```

If you want, i can commit this as a extra commit and if we only want the others, we can just rebase without it.

Kiyoshika: Ohhh gotcha, ok I misunderstood. In that case it would probably be better

Kiyoshika: I have to start work soon so I will be able to review around lunch time, approx 12pm EDT

Kiyoshika:
@Sharktheone looks like your branch merged smoothly, I will test the user agent during lunch time to make sure everything still renders as expected then will merge this into main. Thanks for the help

src/render_tree.rs
```diff
@@ -0,0 +1,438 @@
...
+                            "h5" => Node::new_heading5,
+                            "h6" => Node::new_heading6,
+                            "p" => Node::new_paragraph,
+                            _ => continue, /* add more here, for now ignore */
+                        }(&mut self.position);
+
+                        let new_node = Rc::new(RefCell::new(new_node));
+
+                        util::add_text_node(&reference_element, &new_node);
+                        reference_element = Rc::clone(&new_node);
```
Kiyoshika, line 119: I just realized, the `util::add_text_node` is called here which is in the scope of the `NodeData::Element`. We will have to change this when we encounter different node types besides text (`span`, `div`, etc.)

Maybe we could update a `node_type` or something in the `new_node` match and call the appropriate util function for that node type.

For now this is fine, but will need update it when we get to that point

Sharktheone: Ah... Completely missed that one. Maybe we can return a tuple with the `Node` creation function and the util function to add the node

Kiyoshika: Yeah, that could work. I'm not too worried about it at the moment since we only have one type right now and shouldn't be hard to extend it when we get to that point. I will probably just leave a note for now as a reminder. In the meantime I want to figure out the FFI stuff before I come back to the render tree (unless it's something you're interested in)

ashwinvin (changes requested):

src/render_tree.rs
Kiyoshika: ah, the function name is actually misleading. It has nothing to do with text nodes, it actually works for any generic node. I think I had that name in place before I abstracted it a bit, not sure. I'm just going to change the name and leave it how we have it

src/render_tree/properties.rs
```diff
@@ -0,0 +1,34 @@
...
+#[derive(Debug, PartialEq)]
+pub struct Rectangle {
+    pub top: f64,
+    pub left: f64,
+    pub right: f64,
+    pub bottom: f64,
+}
+
+impl Rectangle {
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
Kiyoshika: I moved the initializer to `default` but I still prefer using `Rectangle::new`... I guess it's an opinionated preference even if it's less idiomatic

Kiyoshika:
squashed all 10 commits into 1 for the final merge

PR #275, opened 2023-11-21
Remove some duplicate code in render tree

Sharktheone:
This tries to get rid of most of the duplicate code in the creation of text nodes for the RenderTree. For the discussion see #272

### DIFF
--- a/src/render_tree.rs
+++ b/src/render_tree.rs
@@ -110,7 +110,7 @@ impl RenderTree {
                             "h5" => Node::new_heading5,
                             "h6" => Node::new_heading6,
                             "p" => Node::new_paragraph,
-                            _ => todo!(), /* add more here, for now ignore */
+                            _ => continue, /* add more here, for now ignore */
                         }(&mut self.position);
 
                         let new_node = Rc::new(RefCell::new(new_node));
--- a/src/render_tree.rs
+++ b/src/render_tree.rs
@@ -164,168 +164,74 @@ impl Node {
         }
     }
 
+    fn new_text(node: TextNode, margin: f64, position: &mut Position) -> Self {
+        position.offset_y(margin);
+        let fs = node.font_size;
+        let new_node = Node {
+            node_type: NodeType::Text(node),
+            margin: Rectangle::with_values(margin, 0., 0., margin),
+            padding: Rectangle::new(),
+            parent: None,
+            next_sibling: None,
+            children: Vec::new(),
+            position: Position::new_from_existing(position),
+        };
+
+        position.offset_y(fs + margin);
+
+        new_node
+    }
+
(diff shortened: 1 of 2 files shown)


src/render_tree.rs
```diff
@@ -215,148 +177,61 @@ impl Node {
...
-    pub fn new_heading2(position: &mut Position) -> Self {
-        let margin = 9.96;
-        position.offset_y(margin);
+    // I took the margins/font sizes from Chrome dev tools.
+    // There are still some slight differences but it's very close
 
-        let new_heading = TextNode::new_heading2();
-        let font_size = new_heading.font_size;
+    pub fn new_heading1(position: &mut Position) -> Self {
+        let margin = 10.72;
```
Sharktheone, line 189: It might make sense to remove the variable for the margin and just write it directly in `Node::new_text`

src/render_tree.rs
```diff
@@ -103,51 +102,21 @@ impl RenderTree {
...
-                        }
+                        let new_node = match element.name.as_str() {
+                            "h1" => Node::new_heading1,
+                            "h2" => Node::new_heading2,
+                            "h3" => Node::new_heading3,
+                            "h4" => Node::new_heading4,
+                            "h5" => Node::new_heading5,
+                            "h6" => Node::new_heading6,
+                            "p" => Node::new_paragraph,
+                            _ => todo!(), /* add more here, for now ignore */
```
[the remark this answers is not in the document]
Sharktheone: Yeah, okay i look into this, it probably is harder because of the direct attached function call

src/render_tree.rs
Kiyoshika: yes, I'd probably agree with that.

src/render_tree/text.rs
```diff
@@ -15,66 +15,41 @@ impl TextNode {
...
     }
 
     pub fn new_heading3() -> Self {
-        Self {
-            value: "".to_owned(),
-            font: DEFAULT_FONT.to_owned(),
-            font_size: 21.5,
-            is_bold: true,
-        }
+        TextNode::new(21.5,true)
```
Kiyoshika, line 37: formatter might complain about having no space between the commas (compared to the earlier `TextNode::new`.) It's minor but if we fix it here, we wouldn't have to fix it in my branch after merging

Sharktheone: Fixed. It's sometimes hard to see, RustRover and all the other Jetbrains IDEs have this "annotation" or something with the name before the value

Kiyoshika: Cool, I am ok to merge this into my branch now to make sure that everything still passes and the rendering on the user agent still works. Thanks again!

PR #277, opened 2023-11-23
Added config settings from the store into dns

jaytaph:
This will setup the dns with configuration settings from the config store.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -42,7 +42,7 @@ lazy_static! {
 }
 
 /// Returns a reference to the config store, which is locked by a mutex.
-/// Any callers of the config store can just do  config::config_store().get("dns.local_resolver.enabled")
+/// Any callers of the config store can just do  config::config_store().get("dns.local.enabled")
 pub fn config_store() -> std::sync::RwLockReadGuard<'static, ConfigStore> {
     CONFIG_STORE.read().unwrap()
 }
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -88,7 +88,7 @@ macro_rules! config {
     (map $key:expr) => {
         match crate::config::config_store().get($key) {
             Some(setting) => setting.to_map(),
-            None => std::collections::HashMap::new(),
+            None => Vec::new(),
```
jaytaph, line 91: maps are vecs, not hashmaps

src/dns.rs
```diff
@@ -109,11 +109,26 @@ struct Dns {
 
 impl Dns {
     pub fn new() -> Self {
+        // Cache resolver
+        let max_entries = config!(uint "dns.cache.max_entries");
```
jaytaph, line 113: we create a configuration file based on the config settings from the store

src/dns/remote.rs
```diff
@@ -80,10 +81,20 @@ impl DnsResolver for RemoteResolver {
...
-    pub fn new(_opts: RemoteResolverOptions) -> RemoteResolver {
+    pub fn new(dns_opts: RemoteResolverOptions) -> RemoteResolver {
         // @todo: do something with the options
-        let config = ResolverConfig::default();
-        let opts = ResolverOpts::default();
+        let mut config = ResolverConfig::default();
+        let mut opts = ResolverOpts::default();
+
+        for nameserver in dns_opts.nameservers.iter() {
+            if let Ok(ip) = IpAddr::from_str(nameserver.as_str()) {
```
Sharktheone, line 90: I guess, you don't need `as_str` here, because it is already a reference to a string. `&String` also counts as `&str`. But feel free to ignore.

Sharktheone (approved):

src/dns/remote.rs
emwalker: I'd use `IpAddr::from_str(&nameserver)` and only fall back to `as_str()` if the compiler complains.

emwalker (approved):
LGTM with suggestion

PR #278, opened 2023-11-24
Http request / response objects and initial API endpoint

jaytaph:
Starting on a network layer and engine api entrypoint.  We use a custom http request/response since the crates found are a bit overkill for now.  This PR adds a simple http response and request objects that are used for returning in the gosub engine API to the user-agents.

The current api consists of a single function:
```
fn fetch_url(
    method: &str,
    url: &str,
    headers: Headers,
    cookies: CookieJar,
) -> Result<FetchResponse, Error> {
```

That will retrieve the URL and returns a fetch response which is a collection of information (request send, response received, document tree, render tree (later), parser errors and timings.  

This is not a finished system, but a simple start.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,6 @@ simple_logger = "4.2.0"
 shared_singleton = "0.1.0"
 testing_logger = "0.1.1"
 cookie = { version = "0.18.0", features = ["secure", "private"] }
-reqwest = { version = "0.11.22", features = ["blocking"] }
 http = "1.0.0"
 url = "2.5.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,10 @@ hickory-resolver = "0.24.0"
 simple_logger = "4.2.0"
 shared_singleton = "0.1.0"
 testing_logger = "0.1.1"
+cookie = { version = "0.18.0", features = ["secure", "private"] }
+reqwest = { version = "0.11.22", features = ["blocking"] }
+http = "1.0.0"
+url = "2.5.0"
 
 [dev-dependencies]
 criterion = { version = "0.5", features = ["html_reports"] }
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -275,6 +275,14 @@ impl CharIterator {
         result
     }
 
+    /// Read directly from bytes
+    pub fn read_from_bytes(&mut self, bytes: &[u8], e: Option<Encoding>) -> io::Result<()> {
+        self.u8_buffer = bytes.to_vec();
+        self.force_set_encoding(e.unwrap_or(Encoding::UTF8));
+        self.reset();
+        Ok(())
+    }
+
     /// Populates the current buffer with the contents of given file f
     pub fn read_from_file(&mut self, mut f: File, e: Option<Encoding>) -> io::Result<()> {
         // First we read the u8 bytes into a buffer
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -12,6 +12,8 @@ use core::fmt::Debug;
 use std::io::Read;
 use url::Url;
 
+const USER_AGENT: &str = "Mozilla/5.0 (compatible; gosub/0.1; +https://gosub.io)";
+
 const MAX_BYTES: u64 = 10_000_000;
 
 /// Response that is returned from the fetch function
@@ -77,7 +79,9 @@ fn fetch_url(
     fetch_response.timings.start(Timing::DnsLookup);
 
     let mut resolver = crate::dns::Dns::new();
-    let hostname = parts.host_str().expect("no hostname");
+    let Some(hostname) = parts.host_str() else {
+        return Err(Error::Generic(format!("invalid hostname: {}", url)));
+    };
     let _ = resolver.resolve(hostname, ResolveType::Ipv4)?;
 
     fetch_response.timings.end(Timing::DnsLookup);
@@ -85,12 +89,8 @@ fn fetch_url(
     // Fetch the HTML document from the site
     fetch_response.timings.start(Timing::ContentTransfer);
 
-    let mut req = match method.to_ascii_lowercase().as_str() {
-        "get" => ureq::get(url),
-        "post" => ureq::post(url),
-        _ => return Err(Error::Generic(format!("unknown method: {method}"))),
-    };
-
+    let agent = ureq::agent();
+    let mut req = agent.request(method, url).set("User-Agent", USER_AGENT);
     for (key, value) in headers.sorted() {
         req = req.set(key, value);
     }
@@ -105,6 +105,7 @@ fn fetch_url(
                     fetch_response.response.headers.set(key.as_str(), value);
                 }
             }
+            // TODO: cookies
             // for cookie in resp.cookies() {
             //     fetch_response.response.cookies.insert(cookie.name().to_string(), cookie.value().to_string());
             // }
@@ -150,8 +151,6 @@ fn fetch_url(
 #[cfg(test)]
 mod tests {
     use super::*;
-
-    const USER_AGENT: &str = "Mozilla/5.0 (compatible; gosub/0.1; +https://gosub.io)";
 
     #[test]
     fn test_fetch_url() {
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1,0 +1,187 @@
+use crate::bytes::{CharIterator, Confidence, Encoding};
+use crate::dns::ResolveType;
+use crate::html5::parser::document::{Document, DocumentBuilder, DocumentHandle};
+use crate::html5::parser::Html5Parser;
+use crate::net::errors::Error;
+use crate::net::http::headers::Headers;
+use crate::net::http::request::Request;
+use crate::net::http::response::Response;
+use crate::timing::{Timing, TimingTable};
+use crate::types::ParseError;
+use cookie::CookieJar;
+use core::fmt::Debug;
+use core::str::FromStr;
+use reqwest::header::HeaderName;
+use reqwest::Method;
+use url::Url;
+
+/// Response that is returned from the fetch function
+pub struct FetchResponse {
+    /// Request that has been send
+    pub request: Request,
+    /// Response that has been received
+    pub response: Response,
+    /// Document tree that is made from the response
+    pub document: DocumentHandle,
+    /// Parse errors that occurred while parsing the document tree
+    pub parse_errors: Vec<ParseError>,
+    /// Rendertree that is generated from the document tree and css tree
+    pub render_tree: String,
+    /// Timing table that contains all the timings
+    pub timings: TimingTable,
+}
+
+impl Debug for FetchResponse {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        writeln!(f, "Request:")?;
+        writeln!(f, "{}", self.request)?;
+        writeln!(f, "Response:")?;
+        writeln!(f, "{}", self.response)?;
+        writeln!(f, "Document tree:")?;
+        writeln!(f, "{}", self.document)?;
+        writeln!(f, "Parse errors:")?;
+        for error in &self.parse_errors {
+            writeln!(f, "  ({}:{}) {}", error.line, error.col, error.message)?;
+        }
+        writeln!(f, "Render tree:")?;
+        writeln!(f, "{}", self.render_tree)?;
+        writeln!(f, "Timings:")?;
+        writeln!(f, "{}", self.timings)?;
+
+        Ok(())
+    }
+}
+
+fn fetch_url(
+    method: &str,
+    url: &str,
+    headers: Headers,
+    cookies: CookieJar,
+) -> Result<FetchResponse, Error> {
+    let mut http_req = Request::new(method, url, "HTTP/1.1");
+    http_req.headers = headers.clone();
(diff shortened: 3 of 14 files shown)


Cargo.toml
```diff
@@ -58,6 +58,10 @@ hickory-resolver = "0.24.0"
 simple_logger = "4.2.0"
 shared_singleton = "0.1.0"
 testing_logger = "0.1.1"
+cookie = { version = "0.18.0", features = ["secure", "private"] }
+reqwest = { version = "0.11.22", features = ["blocking"] }
```
jaytaph, line 62: we do blocking reqwest for now.. this will be changed into something better later on

src/bytes.rs
```diff
@@ -275,6 +275,14 @@ impl CharIterator {
         result
     }
 
+    /// Read directly from bytes
+    pub fn read_from_bytes(&mut self, bytes: &[u8], e: Option<Encoding>) -> io::Result<()> {
```
jaytaph, line 279: Reqwest (and pretty much every system) will return an array of bytes, so we should be able to read that. I'm not 100% sure about ownership and who needs to take it.. We don't want this to be copied a few times.

src/engine.rs
```diff
@@ -0,0 +1,187 @@
...
+use crate::types::ParseError;
+use cookie::CookieJar;
+use core::fmt::Debug;
+use core::str::FromStr;
+use reqwest::header::HeaderName;
+use reqwest::Method;
+use url::Url;
+
+/// Response that is returned from the fetch function
+pub struct FetchResponse {
```
jaytaph, line 19: This is the response from the gosub-engine when you fetch a url. Probably will change in the future, but I think its a start to generate data on a useragent

src/engine.rs
```diff
@@ -0,0 +1,187 @@
...
+        parse_errors: vec![],
+        render_tree: String::new(),
+        timings: TimingTable::default(),
+    };
+
+    // For now, we do a DNS lookup here. We don't use this information yet, but it allows us to
+    // measure the DNS lookup time.
+    fetch_response.timings.start(Timing::DnsLookup);
+
+    let mut resolver = crate::dns::Dns::new();
```
jaytaph, line 83: Doesn't do anything, but we use it anyway for timing purposes.

src/engine.rs
```diff
@@ -0,0 +1,187 @@
...
+    if res.is_err() {
+        return Err(Error::Generic(format!("Failed to resolve domain: {}", url)));
+    }
+
+    fetch_response.timings.end(Timing::DnsLookup);
+
+    // Fetch the HTML document from the site
+    fetch_response.timings.start(Timing::ContentTransfer);
+
+    let m = Method::from_str(method).unwrap();
```
jaytaph, line 94: Mostly converting our http request/response to reqwest objects. This can be done better, but this will be removed later on anyway

src/net/http/headers.rs
```diff
@@ -0,0 +1,50 @@
+use std::collections::HashMap;
+
+#[derive(Default, Debug, Clone)]
+pub struct Headers {
+    headers: HashMap<String, String>,
+}
+
+impl Headers {
```
jaytaph, line 8: This is a very simple encapsulation of a hashmap. I think reqwest has a different way that may or may not be better.. but for now this will do.

src/net/http/headers.rs
```diff
@@ -0,0 +1,50 @@
...
+        self.headers.get(key)
+    }
+
+    /// Returns all the header entries. Note that there is no ordering in here!
+    pub fn all(&self) -> &HashMap<String, String> {
+        &self.headers
+    }
+
+    pub fn sorted(&self) -> Vec<(&String, &String)> {
+        let mut sorted = self.headers.iter().collect::<Vec<_>>();
```
jaytaph, line 29: Hashmaps are not deterministic in output/iteration, so we use a sorted() to always get the same results. A bit of overhead, but again, good enough for now

src/timing.rs
```diff
@@ -0,0 +1,183 @@
...
+    ServerProcessing,
+    ContentTransfer,
+    HtmlParse,
+    CssParse,
+    JsParse,
+    RenderTree,
+}
+
+#[derive(Default, Debug, Clone)]
+pub struct Timer {
```
jaytaph, line 17: Timer is just a start and end time, and calculates the differences between (all in milliseconds)

src/timing.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    pub fn duration(&self) -> u64 {
+        self.end - self.start
+    }
+}
+
+/// Timing information for a single request. Timing is measured in milliseconds.
+#[derive(Default, Debug, Clone)]
+pub struct TimingTable {
+    /// Time spent making DNS queries.
```
jaytaph, line 57: The timing sections. We probably need more but that's easy enough to do.

Cargo.toml
emwalker: We had `reqwest` in one of the binaries at one point, and that brought in Tokio.  It was replaced with `ureq`, which is simple and synchronous and doesn't bring in an async runtime.  Is there a reason `ureq` won't be adequate here? I'd pitch for using `ureq` where you're thinking of using `reqwest`.

src/engine.rs
```diff
@@ -0,0 +1,187 @@
...
+        Ok(())
+    }
+}
+
+fn fetch_url(
+    method: &str,
+    url: &str,
+    headers: Headers,
+    cookies: CookieJar,
+) -> Result<FetchResponse, Error> {
```
emwalker, line 60:
```suggestion
) -> Result<FetchResponse> {
```

To get this to work, you'd need to use the `Result` type from `crate::types`, and you might need to translate the error.

Cargo.toml
jaytaph: Ureq is prob fine too. Ill change that

emwalker:
@jaytaph I agree with implementing our own type for requests and responses.  I do not agree with the reasoning, i.e., that none of them are quite adequate for what we need.  The reason I think we want wrappers for this kind of thing is so that where a struct is widely used in our code, we can control the interface carefully and potentially swap out the third-party crate later on without too much hassle.  (I.e., most of the code is not tightly coupled with the third party crate.)

I took a pass at the contents of this PR and have the following suggested changes, which will be easier to show in a patch rather than line by line:

```patch
diff --git a/src/engine.rs b/src/engine.rs
index d6e78d3..9ae5834 100644
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -2,19 +2,18 @@ use crate::bytes::{CharIterator, Confidence, Encoding};
 use crate::dns::ResolveType;
 use crate::html5::parser::document::{Document, DocumentBuilder, DocumentHandle};
 use crate::html5::parser::Html5Parser;
-use crate::net::errors::Error;
 use crate::net::http::headers::Headers;
 use crate::net::http::request::Request;
 use crate::net::http::response::Response;
 use crate::timing::{Timing, TimingTable};
-use crate::types::ParseError;
+use crate::types::{Error, ParseError, Result};
 use cookie::CookieJar;
 use core::fmt::Debug;
-use core::str::FromStr;
-use reqwest::header::HeaderName;
-use reqwest::Method;
+use std::io::Read;
 use url::Url;
 
+const MAX_BYTES: u64 = 10_000_000;
+
 /// Response that is returned from the fetch function
 pub struct FetchResponse {
     /// Request that has been send
@@ -57,15 +56,12 @@ fn fetch_url(
     url: &str,
     headers: Headers,
     cookies: CookieJar,
-) -> Result<FetchResponse, Error> {
+) -> Result<FetchResponse> {
     let mut http_req = Request::new(method, url, "HTTP/1.1");
     http_req.headers = headers.clone();
     http_req.cookies = cookies.clone();
 
-    let parts = Url::parse(url);
-    if parts.is_err() {
-        return Err(Error::Generic(format!("Failed to parse URL: {}", url)));
-    }
+    let parts = Url::parse(url)?;
 
     let mut fetch_response = FetchResponse {
         request: http_req,
@@ -81,39 +77,47 @@ fn fetch_url(
     fetch_response.timings.start(Timing::DnsLookup);
 
     let mut resolver = crate::dns::Dns::new();
-    let res = resolver.resolve(parts.unwrap().host_str().unwrap(), ResolveType::Ipv4);
-    if res.is_err() {
-        return Err(Error::Generic(format!("Failed to resolve domain: {}", url)));
-    }
+    let hostname = parts.host_str().expect("no hostname");
+    let _ = resolver.resolve(hostname, ResolveType::Ipv4)?;
 
     fetch_response.timings.end(Timing::DnsLookup);
 
     // Fetch the HTML document from the site
     fetch_response.timings.start(Timing::ContentTransfer);
 
-    let m = Method::from_str(method).unwrap();
-    let u = Url::parse(url).unwrap();
-    let mut req = reqwest::blocking::Request::new(m, u);
+    let mut req = match method.to_ascii_lowercase().as_str() {
+        "get" => ureq::get(url),
+        "post" => ureq::post(url),
+        _ => return Err(Error::Generic(format!("unknown method: {method}"))),
+    };
+
     for (key, value) in headers.sorted() {
-        req.headers_mut()
-            .insert(HeaderName::from_str(key).unwrap(), value.parse().unwrap());
+        req = req.set(key, value);
     }
 
-    match reqwest::blocking::Client::new().execute(req) {
+    match req.call() {
         Ok(resp) => {
             fetch_response.response = Response::new();
-            fetch_response.response.status = resp.status().as_u16();
-            fetch_response.response.version = format!("{:?}", resp.version());
-            for (key, value) in resp.headers().iter() {
-                fetch_response
-                    .response
-                    .headers
-                    .set(key.as_str(), value.to_str().unwrap());
+            fetch_response.response.status = resp.status();
+            fetch_response.response.version = format!("{:?}", resp.http_version());
+            for key in &resp.headers_names() {
+                for value in resp.all(key) {
+                    fetch_response.response.headers.set(key.as_str(), value);
+                }
             }
             // for cookie in resp.cookies() {
             //     fetch_response.response.cookies.insert(cookie.name().to_string(), cookie.value().to_string());
             // }
-            fetch_response.response.body = resp.bytes().unwrap().to_vec();
+
+            let len = if let Some(header) = resp.header("Content-Length") {
+                header.parse::<usize>().unwrap_or_default()
+            } else {
+                MAX_BYTES as usize
+            };
+
+            let mut bytes: Vec<u8> = Vec::with_capacity(len);
+            resp.into_reader().take(MAX_BYTES).read_to_end(&mut bytes)?;
+            fetch_response.response.body = bytes;
         }
         Err(e) => {
             return Err(Error::Generic(format!("Failed to fetch URL: {}", e)));
@@ -179,6 +183,7 @@ mod tests {
         let cookies = CookieJar::new();
 
         let resp = fetch_url("GET", url, headers, cookies);
+        dbg!(&resp);
         assert!(resp.is_ok());
 
         let resp = resp.unwrap();
diff --git a/src/types.rs b/src/types.rs
index 8e77176..bcf0560 100644
--- a/src/types.rs
+++ b/src/types.rs
@@ -20,8 +20,8 @@ pub enum Error {
     #[error("config error: {0}")]
     Config(String),
 
-    #[error("ureq error")]
-    Request(#[from] Box<ureq::Error>),
+    #[error("reqwest error: {0}")]
+    Reqwest(#[from] reqwest::Error),
 
     #[error("io error: {0}")]
     IO(#[from] std::io::Error),
@@ -61,6 +61,12 @@ pub enum Error {
 
     #[error("dns: domain not found")]
     DnsDomainNotFound,
+
+    #[error("there was a problem: {0}")]
+    Generic(String),
+
+    #[error("failed to parse url: {0}")]
+    Url(#[from] url::ParseError),
 }
 
 /// Result that can be returned which holds either T or an Error
```

Cargo.toml
emwalker: It looks like it won't be hard.  See the code snippet in a comment below.

emwalker (approved):

Sharktheone (approved):
Just some small things, so feel free to ignore

src/engine.rs
```diff
@@ -0,0 +1,169 @@
...
+        render_tree: String::new(),
+        timings: TimingTable::default(),
+    };
+
+    // For now, we do a DNS lookup here. We don't use this information yet, but it allows us to
+    // measure the DNS lookup time.
+    fetch_response.timings.start(Timing::DnsLookup);
+
+    let mut resolver = crate::dns::Dns::new();
+    let hostname = parts.host_str().expect("no hostname");
```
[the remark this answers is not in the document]
jaytaph: fixed

src/engine.rs
```diff
@@ -0,0 +1,169 @@
...
+    let _ = resolver.resolve(hostname, ResolveType::Ipv4)?;
+
+    fetch_response.timings.end(Timing::DnsLookup);
+
+    // Fetch the HTML document from the site
+    fetch_response.timings.start(Timing::ContentTransfer);
+
+    let mut req = match method.to_ascii_lowercase().as_str() {
+        "get" => ureq::get(url),
+        "post" => ureq::post(url),
```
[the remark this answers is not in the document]
jaytaph: I use "request" now:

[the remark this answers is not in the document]
jaytaph: let mut req = agent.request(method, url).set("User-Agent", USER_AGENT);

src/timing.rs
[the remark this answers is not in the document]
jaytaph: Looks similar. I will let it what it is since it works and is not that important

PR #279, opened 2023-11-25
Introducing the C API

Kiyoshika:
This introduces the first set of C bindings for gosub which, at the moment, just exposes the render tree that we can use in Swift to render the text on the screen.

Ideally, we will build out an "engine API" and wrap that so the user agents can use, instead of using gosub-internal data structures, but right now this builds the foundation for the C API.

EDIT: unsafe code has been completely moved out of the core engine and instead a wrapper has been introduced in the bindings to convert between engine structs and C-friendly structs.
~~I'm trying to keep as much unsafe code out of the engine as possible, but in some cases this is unavoidable to keep the structs compatible with the C interface. We may be able to move this stuff out of the engine at some point.~~

Currently the only known problem with the C API is that we are leaking memory from the `DocumentHandle` held by the `RenderTree`. I'm currently investigating this but unfortunately can't find the root cause after trying various different ways to "free" this memory. We may have to use something other than a `DocumentHandle` to get around it... not sure yet

After this is merged, I'd like to write some docs on the gosub wiki / github wiki about how to use the C API "safely" which will also be useful to others building native clients (the main consumers of the API)


src/render_tree.rs
```diff
@@ -122,7 +126,22 @@ impl RenderTree {
...
+                                );
+                                if !try_realloc.is_null() {
+                                    element_text.value = try_realloc as *mut c_char;
+                                    libc::strncat(
+                                        element_text.value,
+                                        text.value().as_ptr() as *const c_char,
+                                        text.value().len() as size_t,
+                                    );
+                                }
+                            }
```
Kiyoshika, line 144: this is the main piece of "unsafe" code in the engine (outside of the bindings themselves)

We can have multiple text nodes tied to a node, these need to be "concatenated" together to create the text body of that node. Since the underlying type is a raw C ptr, we have to use some libc functions that (attempt to) rellocate the appropriate amount of memory for the new string and call `strncat` to append the text (with one extra byte for the null terminator)

if the reallocation fails, this exits early and the additional text nodes (in theory) should be ignored

src/render_tree.rs
```diff
@@ -152,7 +172,7 @@ pub struct Node {
 impl Node {
     pub fn new() -> Self {
         Self {
-            node_type: NodeType::Root,
+            node_type: NodeType::Root(true),
```
Kiyoshika, line 175: This is a dummy value that needs to be included otherwise it seems like it's ignored going from Rust -> C

src/render_tree.rs
```diff
@@ -14,8 +18,8 @@ pub mod util;
 /// to draw an object
 #[derive(Debug, PartialEq)]
 pub struct Position {
-    pub x: f64,
-    pub y: f64,
+    pub x: f32,
+    pub y: f32,
```
Kiyoshika, line 22: Using `f32` to correspond to C's `float` type

src/render_tree.rs
```diff
@@ -330,107 +349,4 @@ mod tests {
             panic!()
         }
     }
-
-    #[test]
-    fn text_nodes() {
```
Kiyoshika, line 335: This test is moved to `gosub-bindings/tests`

src/render_tree/text.rs
```diff
@@ -1,20 +1,21 @@
...
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone, Copy)]
+#[repr(C)]
 pub struct TextNode {
     /// Body of the text node that will be drawn
-    pub value: String,
-    pub font: String,
-    pub font_size: f64,
+    pub value: *mut c_char,
+    pub font: *mut c_char,
```
Kiyoshika, line 8: Rust strings are not compatible with C, and using `CString` is too inconvenient since we need to append text, so storing raw pointers here is the most convenient

src/render_tree/text.rs
```diff
@@ -1,20 +1,21 @@
...
 }
 
 impl TextNode {
-    fn new(fs: f64, bold: bool) -> Self {
+    fn new(fs: f32, bold: bool) -> Self {
         Self {
-            value: "".to_owned(),
-            font: DEFAULT_FONT.to_owned(),
+            value: CString::new("").expect("").into_raw(),
+            font: CString::new("Times New Roman").expect("").into_raw(),
```
Kiyoshika, line 18: Probably worth adding a useful message if the allocation fails

gosub-bindings/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+/// If the HTML fails to parse, returns a NULL pointer.
+///
+/// # Safety
+/// Takes a read-only pointer owned from the C API representing the HTML source
+/// to build a render tree. DO NOT take ownership of this pointer in Rust or the
+/// universe might collapse.
+///
+/// Moves an owning pointer to the render_tree using Box::into_raw() to the C API.
+/// This pointer MUST be passed to gosub_render_tree_free() after usage for proper cleanup.
+pub unsafe extern "C" fn gosub_render_tree_init(html: *const c_char) -> *mut RenderTree {
```
Sharktheone, line 26: For now, this is okay, but i guess we want this most likely to be a URL and not the HTML document, so we can use the browser's dns system

gosub-bindings/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+    let html_str: &str;
+    unsafe {
+        html_str = CStr::from_ptr(html).to_str().unwrap();
+    }
+    let mut chars = CharIterator::new();
+    chars.read_from_str(html_str, Some(Encoding::UTF8));
+    chars.set_confidence(Confidence::Certain);
+
+    let doc = DocumentBuilder::new_document();
+    let parse_result = Html5Parser::parse_document(&mut chars, Document::clone(&doc), None);
```
Sharktheone, line 36: I've noticed here, we maybe should not return a Result with a `Vec<ParseError>` as the `Ok` type. Maybe i missed something here?

gosub-bindings/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+#[no_mangle]
+/// Free the iterator pointer obtained from gosub_render_tree_iterator_init()
+///
+/// # Safety
+/// This takes ownership of the pointer from the C API and transfers it to Rust so it can
+/// be deallocated.
+pub unsafe extern "C" fn gosub_render_tree_iterator_free(tree_iterator: *mut TreeIterator) {
+    let _ = Box::from_raw(tree_iterator);
```
Sharktheone, line 108: Have you tested using `drop_in_place` here? At leased you didn't wrote something in Zulip.

src/render_tree.rs
Sharktheone: `f64` should be C's `double`, so you can leave that as `f64`

src/render_tree/text.rs
Sharktheone: I don't really like having a raw ptr somewhere in the engine. Are you sure, you can't do something like a `cstring_append` function?

Sharktheone (changes requested):
I don't really like the usage of mut ptrs in the engine itself. I think, there must be a other alternative to deal with this. Maybe you can add other struct variants for example `CTextNode` and convert it to this. But I'm not sure, what the performance would say about this.

src/render_tree/text.rs
Kiyoshika: I did not finding any appending functionality with CString. The only alternative I can think of at the moment is storing them as strings in the engine, but in the bindings converting them to a different struct that will convert String to the raw pointer and transfer ownership to the C API. That might be doable

Kiyoshika: Ah, just saw your other message below, I guess we had similar thinking

gosub-bindings/src/lib.rs
Kiyoshika: Yeah, I would want to move to the DNS system eventually, but was not available at the time I wrote this

gosub-bindings/src/lib.rs
Kiyoshika: I did but it segfaulted. There's something odd about the document, I noted it down for myself for future research but could not figure it out yet.

I might move away from DocumentHandle and try Box-ing the Document itself to see if that would work

src/render_tree.rs
Kiyoshika: I can change the API to use `double` instead of `float` and revert back to `f64`

gosub-bindings/src/lib.rs
Kiyoshika: Not sure, @jaytaph would have more insight to that, I am just a consumer of the parser :P

Kiyoshika:
Converting to draft for the following changes:
* Introduce different struct to avoid raw pointers and unsafe code in the engine (and use String again). Convert to this new struct in the bindings that will convert the `String` to a raw pointer and transfer ownership to the C API
* Change `f32` back to `f64` and update C API to use `double` instead of `float`
* move the `#[no_mangle]` tags down to the actual function instead of above the comments

src/render_tree/text.rs
Kiyoshika: Looking through the `String` docs, looks like we can convert it to a box pointer and from there convert it to a raw ptr

`Box::into_raw(my_string.into_boxed_str())`

So maybe for the `CTextNode` we can implement a `From` trait that takes a regular `TextNode` containing the strings and perform the conversions to send back to C

src/render_tree.rs
Kiyoshika: No longer relevant, unsafe code has been removed from the engine

gosub-bindings/src/gosub_api.c
```diff
@@ -0,0 +1,55 @@
+#include "gosub_api.h"
+
+struct render_tree_t *render_tree_init(const char *html) {
+  struct render_tree_t *render_tree = malloc(sizeof(*render_tree));
```
Sharktheone, line 4: Has this a specific usage, or why are you using malloc and store it on the heap? From my C knowledge, you can initialize the struct like `struct render_tree_t *render_tree`.

gosub-bindings/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+/// universe might collapse.
+///
+/// Moves an owning pointer to the render_tree using Box::into_raw() to the C API.
+/// This pointer MUST be passed to gosub_render_tree_free() after usage for proper cleanup.
+#[no_mangle]
+pub unsafe extern "C" fn gosub_render_tree_init(html: *const c_char) -> *mut RenderTree {
+    let html_str: &str;
+    unsafe {
+        html_str = CStr::from_ptr(html).to_str().unwrap();
+    }
```
Sharktheone, line 34:
```suggestion
    let html_str = unsafe {
        if let Ok(html_str) = CStr::from_ptr(html).to_str() {
            html_str
        } else {
            return ptr::null_mut();
        }
    };
```

Returning null to C instead of crashing the program when something fails is probably better

gosub-bindings/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+/// to the next node.
+#[no_mangle]
+pub unsafe extern "C" fn gosub_render_tree_next_node(
+    tree_iterator: *mut TreeIterator,
+) -> *const Node {
+    let next = (*tree_iterator).next();
+    if next.is_none() {
+        return ptr::null();
+    }
+    next.unwrap().as_ptr() as *const Node
```
Sharktheone, line 80: Just a bit of cleaning up

```suggestion
    if let Some(next) = next {
        next.as_ptr() as *const Node
    } else {
        ptr::null()
    }
```

src/render_tree.rs
```diff
@@ -2,8 +2,9 @@ use std::borrow::BorrowMut;
 use std::{cell::RefCell, rc::Rc};
 
 use crate::html5::node::NodeData;
-use crate::html5::parser::document;
+use crate::html5::parser::document::{self};
```
Sharktheone, line 5:
```suggestion
use crate::html5::parser::document;
```

src/render_tree.rs
Sharktheone: Not sure if you want to have this tests only over the C API

Sharktheone (changes requested):
Some minor things, but at most of them feel free to ignore.
Note that i cannot say too much about the C side.

gosub-bindings/src/gosub_api.c
Kiyoshika: It doesn’t have to be on the heap but it simplifies the initialization a bit. Otherwise the caller will have to create a render tree struct on the stack and pass it to the init function by address. I don’t really have a strong opinion as to whether or not it should be on the stack or heap since it’s pretty small holding only a few pointers

gosub-bindings/src/lib.rs
Kiyoshika: yep, I agree

gosub-bindings/src/gosub_api.c
Sharktheone: I'd say, when it can be on the stack, it should be on the stack

Kiyoshika: That’s fair enough, I can change it

src/render_tree.rs
Kiyoshika: This was a weird auto import from Zed, guess I didn't catch it

src/render_tree.rs
Kiyoshika: yeah I was on the fence about this... my thought was the main purpose of this portion of the test was more relevant for the C API, and it might be good to have both tests but then I'd have to maintain two versions of the same test, and I'd rather have the C API be right (which would also imply the render tree implementation is correct.) But I can see the argument for having it in the render tree itself as well

Kiyoshika:
@Sharktheone added your suggestions and also switched the `render_tree_init` to use a stack-allocated struct instead of allocating one on the heap

gosub-bindings/include/gosub_api.h
```diff
@@ -18,9 +18,9 @@ struct render_tree_t {
   struct node_t *data;
 };
 
-/// Construct a new render tree.
-/// Returns NULL if unsuccessful.
-struct render_tree_t *render_tree_init(const char *html);
+/// Initialize a render tree by passing a stack-allocated
+/// struct by address.
+void render_tree_init(struct render_tree_t *render_tree, const char *html);
```
Sharktheone, line 23: I don't know too much about this, but shouldn't it be possible to create the `render_tree_t` in the function? I've seen this approach in many different C Libraries, so it's probably the best solution, when we want to have it on the stack.

gosub-bindings/src/gosub_api.c
```diff
@@ -1,21 +1,16 @@
...
-  struct render_tree_t *render_tree = malloc(sizeof(*render_tree));
-  if (!render_tree)
-    return NULL;
+void render_tree_init(struct render_tree_t *render_tree, const char *html) {
 
   render_tree->tree = gosub_render_tree_init(html);
   if (!render_tree->tree) {
-    free(render_tree);
-    return NULL;
+    return;
```
Sharktheone, line 7: Wouldn't it maybe make sense to return a integer to tell the caller if something went wrong?

Sharktheone (approved):
LGTM, just a minor thing, feel free to ignore.

gosub-bindings/src/gosub_api.c
Kiyoshika: That's usually a common pattern in C, but I don't really think it'll be useful here... It doesn't matter to the caller whether the tree iterator or tree itself failed to allocate (at least, I don't know what exactly you would do with that level of detail)

gosub-bindings/include/gosub_api.h
Kiyoshika: You can't return a pointer to something stack-allocated in a function, C doesn't transfer ownership. If you tried dereferencing it, you would segfault

Kiyoshika: Unless they are doing something that I've never heard of before... I feel like that may be possible in C++ with its move semantics, but I haven't really seen anything like that in C

Kiyoshika: Unless you mean returning the struct by value, I guess I didn't really think of that, but not something I do often since it seems kind of weird to copy the value around (even tho the compiler might do some optimizations on return values)

Kiyoshika: Yes, after some thought, just returning the struct by value might be fine in this case, it's very small anyways

Sharktheone: ah, okay. I don't have much C experience. I mostly used C++ and not raw C.

Kiyoshika: If you look at my personal github, I came from mostly raw C... but usually the structs I work with are relatively large so returning by value didn't come across my mind since I am so used to returning pointers

Sharktheone: > Yes, after some thought, just returning the struct by value might be fine in this case, it's very small anyways

I guess, you can just leave it the way it currently is, but if you want to change it just do it.

gosub-bindings/src/gosub_api.c
Kiyoshika: oh, I am too tired, I was still thinking we are returning a pointer, I forgot I changed this to void. Returning `-1` on failure and `0` on success is something I am good with in this case. I need some sleep :D

Kiyoshika: When I first read that comment, I was thinking of an enum that described which part actually failed, not that a failure in general occurred. My bad

Sharktheone: All good, happens to me all the time... 😄

gosub-bindings/include/gosub_api.h
Kiyoshika: I think I will leave it as a parameter and return the `int` to check if a failure occurred. Then I think this will be ready..

Sharktheone (approved):
LGTM

gosub-bindings/src/lib.rs
emwalker: Are we guaranteed that the Swift Objective C side will never free this `*const c_char` pointer while we're in Rust code?

emwalker (approved):
I'm fully out of my depth.  I have only dabbled in C over the years, so I have no way to approve this PR beyond the impression that you and @Sharktheone seem to know C a lot more than I do.  Godspeed.

gosub-bindings/Makefile
```diff
@@ -0,0 +1,26 @@
...
+	cp ./target/debug/libgosub_bindings.a lib/
+	cp ./libgosub.a lib/
+
+TEST_SRC_DIR := tests
+test: # build and run tests for bindings
+	$(CC) $(TEST_SRC_DIR)/render_tree_test.c -L./ -lgosub -L./target/debug -lgosub_bindings -lsqlite3 -lm -o $(TEST_SRC_DIR)/render_tree_test $(CPPFLAGS) $(CFLAGS)
+	./$(TEST_SRC_DIR)/render_tree_test
+
+format:
+	cargo clippy --fix --allow-dirty --allow-staged
```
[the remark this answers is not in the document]
Kiyoshika: You should only have to `cd gosub-bindings`, `cargo build` then run the make files

Kiyoshika:
@emwalker for some reason I can't reply to your comment

> Are we guaranteed that the Swift Objective C side will never free this *const c_char pointer while we're in Rust code?

When this function is called from swift and the pointer is passed, only the Rust code (technically the C wrapper) will be doing work until after the scope closes and Swift continues, so we wouldn't be able to accidentially free this while it's being used by Rust/C

gosub-bindings/Makefile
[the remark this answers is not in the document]
emwalker: Ah, I see.  Yes, that works.



I see that that leaves a `target` directory.  It seems that we're now getting into the realm of a [Cargo workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html#creating-the-second-package-in-the-workspace), which is what I'm thinking we should switch to at some point soon if you want to keep this crate separate from the main one in this repo.  (Workspaces allow you to publish a lot of crates under a single repo.)

Kiyoshika:
Relevant swift portions if you're curious:

Calling render tree constructor (swift wrapper) - https://github.com/gosub-browser/gosub-client-macos/blob/110d52daaebe09154b2e422cbf2f63d33be51d9d/gosub-client-macos/ContentView.swift#L18C14-L18C14

The internals that call the C code - https://github.com/gosub-browser/gosub-client-macos/blob/110d52daaebe09154b2e422cbf2f63d33be51d9d/gosub-client-macos/RenderTree.swift#L8C14-L8C14

gosub-bindings/Makefile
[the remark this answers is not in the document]
Sharktheone: I also thought about this a while ago, but we only had one crate at that time.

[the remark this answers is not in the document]
emwalker: Might be as easy as this directory structure:

* gosub-engine/
  * engine/
    * `Cargo.toml`
  * bindings/
    * `Cargo.toml`
  * `Cargo.toml` (for all the crates)

```
$ cat Cargo.toml
[workspace]

members = [
    "engine",
    "bindings",
]
```

Sharktheone:
> I'm fully out of my depth. I have only dabbled in C over the years, so I have no way to approve this PR beyond the impression that you and @Sharktheone seem to know C a lot more than I do. Godspeed.

I also don't have too much knowledge about C. My main understanding comes from C++ and only minimal C. But I've read some libraries that are coded in C and have a bit of understanding from C because of that, but also not too much.

Kiyoshika:
I've written quite a lot of C (as that makes up almost my entire github history) but none of it was professional and almost all of it was entirely solo that I've learned over time. But I've never really used it in the context of interacting with other languages (except one case with Python and ctypes for a numpy clone I wrote forever ago)

emwalker:
This talk by a Mozilla developer that wrote a Rust crate that is used in Firefox goes into some challenges he ran into with dropping pointers that vaguely sound a little like the problem you were running into:

Kiyoshika:
My next step is to go down this rabbit hole and try to figure out how to clean up the leak. I think using `Rc<>` will present a lot of challenges with cyclic reference counts that can lead to leaks, but not sure how that's happening in the specific case of the render tree

PR #286, opened 2023-11-27
Css3 parser tryout

jaytaph:
This PR is a css3 parser created as much as possible according the specs to gain more knowledge about the css specification. Main goal would be fill in the blanks in the current css3 parser that we current have (new_parser). This PR tries to follow the specifications as close as possible.

Main entrypoints are the parse_* functions. There are multiple depending on what you like to parse (i'm not 100% sure how this actually works).
The consume_* functions are the parts that will consume tokens and return css structures like Rules, QualifiedValues etc.


src/css3.rs
```diff
@@ -1,120 +1,959 @@
...
+        TokenStream { tokens, index: 0 }
+    }
+
+    fn new_from_componentvalues(values: Vec<ComponentValue>) -> TokenStream {
+        let tokens = values.iter().filter(|v| {
+            if let ComponentValue::PreservedToken(_) = v {
+                true
+            } else {
+                false
+            }
```
Sharktheone, line 41:
```suggestion
            matches!(v, ComponentValue::PreservedToken(_))
```

src/css3.rs
```diff
@@ -1,120 +1,959 @@
...
+        }).map(|v| {
+            if let ComponentValue::PreservedToken(token) = v {
+                token.clone()
+            } else {
+                panic!("we should not be here");
+            }
+        }).collect::<Vec<Token>>().into();
```
Sharktheone, line 48: I think, you can remove the `into` here
```suggestion
        }).collect::<Vec<Token>>();
```

src/css3.rs
```diff
@@ -1,120 +1,959 @@
...
+    }
+}
+
+impl TokenStreamer for TokenStream {
+    fn current(&self) -> Token {
+        if self.index >= self.tokens.len() {
+            return Token::Eof;
+        }
+
+        self.tokens[self.index].clone()
```
Sharktheone, line 63: You can probably change this to

```suggestion
        self.tokens.get(self.index).unwrap_or(&Token::Eof).clone()
```

or if you don't want to clone the Token::Eof you can use

```suggestion
        if let Some(token) = self.tokens.get(self.index) {
            token.clone()
        } else {
            Token::Eof
        }
```

Sharktheone: (same for the others)

src/css3.rs
```diff
@@ -1,120 +1,959 @@
...
+        if self.index > 0 {
+            self.index -= 1;
+        }
+    }
+}
+
+// =================================================================================================
+
+pub struct CSS3ParserTng<'a> {
+    token_stream: &'a mut dyn TokenStreamer,
```
Sharktheone, line 94: When i think about this, it's probably better to use Nico's suggestion. (with the lifetime in the `Box`) because then you can store the `TokenStreamer` actually in the struct. If it doesn't matters, you can leave it as it is.

src/css3.rs
```diff
@@ -1,120 +1,959 @@
...
+                            _ => {}
+                        }
+
+                        if let Some(component_value) = self.consume_component_value() {
+                            match component_value {
+                                ComponentValue::PreservedToken(token) => {
+                                    tmp_input.push(token);
+                                }
+                                ComponentValue::Function(_function) => {
+                                    panic!("we should not have a function here");
```
Sharktheone, line 544: I assume, you'll change this to a be a parser error in the future?

src/css3.rs
```diff
@@ -1,120 +1,959 @@
...
+        self.consume_whitespaces();
+
+        loop {
+            match self.token_stream.consume() {
+                Token::Eof => break,
+                _ => {
+                    if let Some(component_value) = self.consume_component_value() {
+                        declaration.value.push(component_value);
+                    }
+                }
```
Sharktheone, line 667: I guess, you can change this to the `if let` syntax as in #268 recommended.

src/css3.rs
```diff
@@ -1,120 +1,959 @@
...
+        let current_token = self.token_stream.current();
+        match self.token_stream.consume() {
+            Token::LCurly | Token::LBracket | Token::LParen => {
+                match self.consume_simple_block(current_token) {
+                    Some(block) => {
+                        return Some(ComponentValue::SimpleBlock(block));
+                    }
+                    None => {
+                        // parse error
+                    }
```
Sharktheone, line 702: I guess, here it also could be a `if let` especially when you ignore the `None` case. (same for 706-712 and some more, i won't add a comment for all of them)

src/css3.rs
```diff
@@ -1,120 +1,959 @@
...
+
+impl CssAst {
+    pub fn new() -> CssAst {
+        CssAst {
+            tree: Node {
+                name: "".to_string(),
+                attributes: HashMap::new(),
+                children: Vec::new(),
+            },
+        }
```
Sharktheone, line 784: This has no default implementation

src/css3.rs
```diff
@@ -1,120 +1,959 @@
...
+
+    fn compile_norm_rule(&mut self, _norm_rule: NormRule) -> Node {
+        let node = Node {
+            name: "norm-rule".to_string(),
+            attributes: HashMap::new(),
+            children: Vec::new(),
+        };
+
+        node
+
```
Sharktheone, line 837:
```suggestion
        Node::new("norm-rule)
```
(same for the other functions under this one)

src/css3.rs
```diff
@@ -1,120 +1,959 @@
...
+    pub fn display_tree(&self) {
+        self.display_tree_node(&self.tree, 0);
+    }
+
+    fn display_tree_node(&self, node: &Node, level: usize) {
+        println!("{}{}: {:?}", " ".repeat(level), node.name, node.attributes);
+        for child in &node.children {
+            self.display_tree_node(&child, level + 1);
+        }
+    }
```
Sharktheone, line 827:
```suggestion
    pub fn display_tree(&self) {
        Self::display_tree_node(&self.tree, 0);
    }

    fn display_tree_node(node: &Node, level: usize) {
        println!("{}{}: {:?}", " ".repeat(level), node.name, node.attributes);
        for child in &node.children {
            Self::display_tree_node(&child, level + 1);
        }
    }
```

src/css3.rs
```diff
@@ -1,120 +1,959 @@
...
+
+    fn parse_prelude(&mut self, input: Vec<Token>) {
+        // let mut selectors = Vec::new();
+        let mut cur_selector = Vec::new();
+
+        let mut idx = 0;
+        loop {
+            let t = input[idx].clone();
+            idx += 1;
```
Sharktheone, line 877: Are you somewhere exiting this loop? Because else it can (and probably will) panic because of an overflow. Maybe I miss something. (and please add the semicolons after the `push` and maybe change the double quotes to single ones in line 880)

src/css3/ast.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    pub fn to_token(&self) -> Token {
+        return self.0[0].clone();
+    }
+}
+
+impl InputTake for Span<'_> {
+    fn take(&self, count: usize) -> Self {
+        let tokens = &self.0[0..count];
```
Sharktheone, line 47: When you give it a too high count here, it will panic

src/css3/new_tokenizer.rs
```diff
@@ -227,13 +235,64 @@ macro_rules! consume {
...
+    /// Position on the NEXT read to consume. If it's outside the vec list, it will return EOF
     position: usize,
     /// Full list of all tokens produced by the tokenizer
-    tokens: Vec<Token>,
+    pub tokens: Vec<Token>,
+}
+
+impl TokenStreamer for Tokenizer<'_> {
+    fn current(&self) -> Token {
+        if self.position == 0 || self.position > self.tokens.len() {
```
Sharktheone, line 257:
```suggestion
        if self.position == 0 || self.position >= self.tokens.len() {
```
panic-alert

src/css3/nom_selector.rs
```diff
@@ -0,0 +1,350 @@
...
+        let node = match combinator {
+            Some(combinator) => {
+                let mut node = Node::new("Combinator");
+                node.children.push(combinator);
+                node.children.push(selector);
+                node
+            }
+            None => {
+                selector
+            }
```
Sharktheone, line 78: `if let` for idiomacy

Sharktheone (changes requested):
Can't say too much about the functionality and the `nom` part

jaytaph:
Old tryout. Closing in favor of a new setup

PR #288, opened 2023-11-30
Github action for notifiying of missing email in AUTHOR file

jaytaph:
Does the following:

  - fetches the first 10 committers in the diff and checks if either their name or email (or both) are found in the AUTHORS file.
  - if not, it will post a comment saying the user is invited to add their name(s).
  - if found (during a new commit after the first one), and the names are found, the comment is automatically removed.

### DIFF
--- a/.github/workflows/check_authors.sh
+++ b/.github/workflows/check_authors.sh
@@ -12,13 +12,13 @@ is_in_authors() {
 # Only check the first 10 committers found in the PR
 COMMITTERS=$(git log $1 --pretty=format:"%an;%ae" | sort | uniq | head -n 10)
 
-for COMMITTER in "$COMMITTERS" ; do
+for committer in "$COMMITTERS" ; do
   # split sentence in two parts seperated by a ;
-  NAME=$(echo $COMMITTER | cut -d ";" -f 1)
-  EMAIL=$(echo $COMMITTER | cut -d ";" -f 2)
+  local name=$(echo $committer | cut -d ";" -f 1)
+  local email=$(echo $committer | cut -d ";" -f 2)
 
-  if is_in_authors "$EMAIL" == 0 && is_in_authors "$NAME" == 0; then
-    echo "This author is not found in the AUTHORS file"
+  if is_in_authors "$email" == 0 && is_in_authors "$name" == 0; then
+    echo "Author $name <$email> was not found in the AUTHORS file"
     exit 1
   fi
 done
--- a/.github/workflows/check_authors.sh
+++ b/.github/workflows/check_authors.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+is_in_authors() {
+  TARGET=$1
+  if grep -Fq "$TARGET" AUTHORS; then
+    return 1
+  else
+    return 0
+  fi
+}
+
+# Only check the first 10 committers found in the PR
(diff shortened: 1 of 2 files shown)


jaytaph:
don't worry.. just testing some things out on github actions...   always trail and error...  will ping when this is finalized (and rebased)

Sharktheone (commented):
I guess, change it so, that it doesn't post multiple times. Besides that looks good.

.github/workflows/check_authors.sh
```diff
@@ -0,0 +1,18 @@
...
+    echo "Author's email is in the AUTHORS file."
+  else
+    # Post a comment on the PR
+    COMMENT="Hello! Welcome and thank you for your contribution. It appears that you might be a new contributor to our project. To acknowledge your work appropriately, we kindly ask you to add your email to the AUTHORS file. This helps us maintain a record of all our valuable contributors. Thanks again for your involvement!"
+    PAYLOAD=$(jq -n --arg body "$COMMENT" '{body: $body}')
+    COMMENTS_URL=$(jq -r .pull_request.comments_url "$GITHUB_EVENT_PATH")
+
+    curl -s -S -H "Authorization: Bearer $GITHUB_TOKEN" --header "Content-Type: application/json" --data "$PAYLOAD" "$COMMENTS_URL"
+  fi
+done
```
emwalker, line 18: Wondering whether we can allow names without emails.  I already get a lot of spam at my main email address that I'm reluctant to put it in a file that is easy to crawl.

jaytaph: sure.. that's not hard to implement

.github/workflows/check_authors.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+is_in_authors() {
+  TARGET=$1
+  if grep -Fq "$TARGET" AUTHORS; then
+    return 1
+  else
+    return 0
+  fi
+}
```
emwalker, line 10:
```suggestion
is_in_authors() {
  local target=$1
  if grep -Fq "$target" AUTHORS; then
    return 1
  else
    return 0
  fi
}
```

.github/workflows/check_authors.sh
```diff
@@ -0,0 +1,27 @@
...
+for COMMITTER in "$COMMITTERS" ; do
+  # split sentence in two parts seperated by a ;
+  NAME=$(echo $COMMITTER | cut -d ";" -f 1)
+  EMAIL=$(echo $COMMITTER | cut -d ";" -f 2)
+
+  if is_in_authors "$EMAIL" == 0 && is_in_authors "$NAME" == 0; then
+    echo "This author is not found in the AUTHORS file"
+    exit 1
+  fi
+done
```
emwalker, line 24:
```suggestion
for committer in "$COMMITTERS" ; do
  # split sentence in two parts seperated by a ;
  local name=$(echo $committer | cut -d ";" -f 1)
  local email=$(echo $committer | cut -d ";" -f 2)

  if is_in_authors "$email" == 0 && is_in_authors "$name" == 0; then
    echo "Author $name / $email was not found in the AUTHORS file"
    exit 1
  fi
done
```

emwalker (approved):
Small suggestions.  LGTM.

Issue #280, opened 2023-11-25
Investigate the memory leak by DocumentHandle within the RenderTree


Kiyoshika:
Somehow the `DocumentHandle` within the `RenderTree` is leaking when running `gosub-bindings/tests/render_tree_test` under valgrind.

May need to use a different type of pointer or wrap the `Document` directly in a `Box`.

Sharktheone:
You said something about the Rc not being leak-safe. I think it's only not leak safe when you have a reference cycle. So you could use `Rc::downgrade` to convert it to `Weak` which is a weak reference and doesn't count as a real reference (so not preventing a de allocation). But i'm not sure if the problem is a reference cycle.

Kiyoshika:
> You said something about the Rc not being leak-safe. I think it's only not leak safe when you have a reference cycle. So you could use `Rc::downgrade` to convert it to `Weak` which is a weak reference and doesn't count as a real reference (so not preventing a de allocation). But i'm not sure if the problem is a reference cycle.

Yes, that’s also what I’ve read but going through the code I didn’t really understand where it was happening

Kiyoshika:
At the moment this is just conjecture, but it's likely the source of the leak is cloning the `DocumentHandle` in each `Node` which is probably not being cleaned up properly and causing lots of weird cyclical references.

PR #290, opened 2023-11-30
refactor node to use weak pointer to document

Kiyoshika:
This is a small refactor that changes the `Node`'s `DocumentHandle` pointer to a `Weak` pointer to avoid cyclic reference problems and cleans up memory leaks (#280)

I was initially having troubles with the `Weak` pointer staying alive since it appears to be very sensitive to scopes, but @emwalker discovered we had to move around some things to get it to work, now all tests are passing and we verified the memory leak is cleaned up:

* `cd crates/gosub-bindings`
* `cargo build`
* `make bindings`
* `make test`
* `valgrind --leak-check=full ./tests/render_tree_test`


Kiyoshika:
I see there are conflicts from the code idiomacy PR, I'll fix these when I have some time after work.

src/html5/node.rs
```diff
@@ -121,13 +122,18 @@ pub struct Node {
...
     // Returns true when the given node is registered into an arena
     pub is_registered: bool,
 }
 
+impl PartialEq for Node {
+    fn eq(&self, other: &Node) -> bool {
+        self.id == other.id
+    }
+}
+
```
Kiyoshika, line 136: This was added since `Weak<RefCell>` doesn't have a `PartialEq` implementation and `#[derive(PartialEq)]` no longer worked. I think two Nodes should be equal if their `id`s are equal

Kiyoshika:
Nvm the conflicts were a lot smaller than I was imagining, resolved now.

emwalker (approved):

PR #293, opened 2023-12-01
expose node position in C API

Kiyoshika:
This is a small extension to the C API to expose the node position. After this is merged the user agent can be changed to use positions instead of the default vertical-stack layout

### DIFF
--- a/crates/gosub-bindings/include/nodes.h
+++ b/crates/gosub-bindings/include/nodes.h
@@ -13,8 +13,14 @@ void render_tree_node_free(struct node_t **node);
 
 enum node_type_e { NODE_TYPE_ROOT = 0u, NODE_TYPE_TEXT };
 
+struct position_t {
+  double x;
+  double y;
+};
+
 struct node_t {
   enum node_type_e type;
+  struct position_t position;
   union data {
     bool root;               // NODE_TYPE_ROOT
     struct node_text_t text; // NODE_TYPE_TEXT
@@ -22,6 +28,8 @@ struct node_t {
 };
 
 struct node_t *render_tree_node_init();
+double render_tree_node_get_x(const struct node_t *node);
+double render_tree_node_get_y(const struct node_t *node);
 void render_tree_node_free_data(struct node_t *node);
 void render_tree_node_free(struct node_t **node);
 
--- a/crates/gosub-bindings/src/lib.rs
+++ b/crates/gosub-bindings/src/lib.rs
@@ -4,8 +4,6 @@ use std::ptr;
 
 pub mod wrapper;
 
-use crate::wrapper::text::CTextNode;
-
 use gosub_engine::{
     bytes::{CharIterator, Confidence, Encoding},
     html5::parser::{
@@ -14,6 +12,7 @@ use gosub_engine::{
     },
     render_tree::{Node, NodeType, RenderTree, TreeIterator},
 };
+use wrapper::node::CNode;
 
 /// Initialize a render tree and return an owning pointer to it.
 /// If the HTML fails to parse or the html string fails to be converted to Rust,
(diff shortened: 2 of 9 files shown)


crates/gosub-bindings/src/lib.rs
```diff
@@ -91,12 +90,10 @@ pub unsafe extern "C" fn gosub_render_tree_next_node(
 /// and a mutable pointer owned by the C API to write (copy) the contents
 /// of the read-only pointer into the mutable pointer.
 #[no_mangle]
-pub unsafe extern "C" fn gosub_render_tree_get_node_data(
-    node: *const Node,
-    node_data: *mut CTextNode,
-) {
+pub unsafe extern "C" fn gosub_render_tree_get_node_data(node: *const Node, c_node: *mut CNode) {
```
Kiyoshika, line 93: new struct `CNode` introduced to more properly wrap the `node_t` struct from the C API. This is more generic than the previous `CTextNode` when we add more types

crates/gosub-bindings/tests/render_tree_test.c
```diff
@@ -23,13 +23,22 @@ int main() {
   node = render_tree_next(&render_tree);
   assert(node->type == NODE_TYPE_ROOT);
 
+  const double tol = 0.00001;
+
+  // TODO: it'll be good at some point in the future to have the
+  // margins to compute the expected_y position instead of manually
+  // doing math. This will make the tests more robust if we change
+  // margins/etc. in the engine.
```
Kiyoshika, line 31: converting this to issue

crates/gosub-bindings/src/wrapper/node.rs
```diff
@@ -0,0 +1,40 @@
...
+            position: Position::new(),
+            data: CNodeData::Root(true),
+        }
+    }
+}
+
+impl CNode {
+    pub fn new_root() -> Self {
+        Self::default()
+    }
```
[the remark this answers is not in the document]
Kiyoshika: Although now that I think about it, for the purpose of the `RenderTree`, the root node is technically not a visible node so should probably be skipped in the iterator itself

crates/gosub-bindings/include/nodes.h
```diff
@@ -13,15 +13,23 @@ void render_tree_node_free(struct node_t **node);
...
+  struct position_t position;
   union data {
     bool root;               // NODE_TYPE_ROOT
     struct node_text_t text; // NODE_TYPE_TEXT
   } data;
 };
 
 struct node_t *render_tree_node_init();
+double render_tree_node_get_x(const struct node_t *node);
+double render_tree_node_get_y(const struct node_t *node);
```
Sharktheone, line 32: Maybe it makes sense to attach these functions to the node_t structs. The only problem I'd see with attaching them to the node_t is, that it could make it harder to call them.

Kiyoshika: What do you mean exactly? These already take a node_t

Sharktheone: Ah, what i meant is a c++ only feature thought, i did this once in C. Also, i thought i had  seen this once in a c library.

Kiyoshika: Oh, yea c++ allows methods in structs since they’re a special case of a class in that case

Kiyoshika: Some people put function pointers in structs but I only do that if it’s really necessary

Sharktheone: Ya, I thought this was possible because C had no classes. But probably when c had this, there would be no c++

emwalker (approved):

PR #297, opened 2023-12-02
create initial contribution guide

Kiyoshika:
This is the initial contribution guide that I put together. I wrote this between 12am - 1am and am pretty tired so there may be some things that don't make sense. Please feel free to suggest any additions/edits/corrections or spelling issues, etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .SILENT:
 
-SHELL=/usr/bin/env bash -O globstar
+#SHELL=/usr/bin/env bash -O globstar
 
 all: help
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
 .SILENT:
 
-SHELL=/usr/bin/env bash -O globstar
+#SHELL=/usr/bin/env bash -O globstar
```
jaytaph, line 3: I don't know exactly why this was needed. Probably from another Makefile (we use the same layout a lot) where this happens to be needed. So it's ok to remove it

jaytaph (approved):

Makefile
Kiyoshika: Lol I’ve been commenting and uncommenting this out for months because having this line doesn’t work on my Mac and I guess I forgot to remove it here. But if we remove it then it makes it easy for me :P

PR #300, opened 2023-12-02
change default compilation mode to release mode for C API

Kiyoshika:
Originally I was trying to add a variable to determine which mode to use but Make is very annoying and convoluted sometimes that this just made things more complicated than they needed to be. Defaulting to release mode so when the user agent(s) build the bindings, they don't have to change the mode

### DIFF
--- a/crates/gosub-bindings/Makefile
+++ b/crates/gosub-bindings/Makefile
@@ -1,15 +1,37 @@
 # "global" compile settings
-INCLUDE_DIR := include
-SRC_DIR := src
-CPPFLAGS := -I$(INCLUDE_DIR) -I$(INCLUDE_DIR)/nodes
+
+# compilation mode Release or Debug
+MODE?=Release
+
 CFLAGS_DEBUG := -std=c99 -g -Wall -Wextra -O0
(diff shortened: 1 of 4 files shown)


crates/gosub-bindings/Makefile
```diff
@@ -1,15 +1,16 @@
...
+CFLAGS := $(CFLAGS_RELEASE) # *** CONFIGURE RELEASE OR DUBUG MODE ***
+
 INCLUDE_DIR := include
 SRC_DIR := src
 CPPFLAGS := -I$(INCLUDE_DIR) -I$(INCLUDE_DIR)/nodes
-CFLAGS_DEBUG := -std=c99 -g -Wall -Wextra -O0
-CFLAGS_RELEASE := -std=c99 -Wall -Wextra -O2
-CFLAGS := $(CFLAGS_DEBUG)
 CC := gcc
 TARGET_DIR := ../../target/debug
```
Kiyoshika, line 10: After some testing we also need a way to change this to `../../target/release` depending on the mode as well...

Kiyoshika:
Ah, my editor was adding random spaces and also make doesn't allow you to tab in conditionals when it's outside of a rule... the format is very picky but I figured it out now

PR #301, opened 2023-12-02
add var in C API makefile to change compilation mode

Kiyoshika:
This adds a `MODE` variable to the makefile in the `gosub-bindings` to compile in different modes.

By default it will use `Release` mode (for the user agents when they build the bindings) and for development we can switch to `Debug`.

`make bindings` will now also build the crate in the specified mode so we don't have to manually `cargo build` ourselves


crates/gosub-bindings/README.md
```diff
@@ -0,0 +1,18 @@
+# Gosub Bindings
+These are the bindings the expose some of Gosub's engine to the world via a C API. Typically these will be used by user agents.
```
emwalker, line 2:
```suggestion
These are bindings that expose some of Gosub's engine to the world via a C API. Typically these bindings will be used by user agents.
```

emwalker (approved):

PR #310, opened 2023-12-06
Optimize release and benchmark builds

Sharktheone:
I've added some options in `Cargo.toml` to optimize the performance of release and benchmark builds.

- `lto` or link time optimizations, allows optimizations across codegen units and even languages (for example when we are using v8). It also inlines most of the functions. See [this](https://www.reddit.com/r/rust/comments/ijwya5/whats_the_deal_with_lto/) and [this](https://doc.rust-lang.org/cargo/reference/profiles.html#lto)
- `codegen-units` Disallows the compiler to split code in multiple codegen-units, which results in a bit better performence.
- `opt-level` Level at which the compiler optimizes things. (e.g unrolling loops, ...)

This of course causes the build to take much longer. From my measurements a full release build takes about 23.3s instead of 10.8s on my machine. A rebuild (with some dependencies already built) takes 17.3s instead of 2.8s.

Development builds are untouched, so they only take 6.2s with or without these options.

The benchmark results go up ~10-11%. 

<details>

<summary> Benchmark results </summary>

## TreeIterator - wikipedia

TreeIterator

## TreeIterator - stackoverflow

TreeIterator

## TreeConstructor

TreeConstructor

## Tokenizer

Tokenizer

</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,3 +78,16 @@ test-case = "3.3.1"
 [features]
 # Enables extended debugging information during parsing.
 debug_parser = []
+
+
+[profile.release]
+lto = "fat"
+opt-level = 3
+codegen-units = 1
+#rustflags = ["-C", "target-cpu=native"] # Currently unstable and it can cause problems on older CPUs (don't have newer CPU instructions).
+
+[profile.bench]
+lto = "fat"
+opt-level = 3
+codegen-units = 1
+


jaytaph:
pr author check fail is fine. This probably needs finetuning or we might need to get rid of it if it doesn't work out well enough..

Sharktheone:
> pr author check fail is fine. This probably needs finetuning or we might need to get rid of it if it doesn't work out well enough..

Probably because I didn't add myself to the file.

jaytaph:
maybe, but that check is still broken :-)

PR #321, opened 2023-12-24
Add an basic system information fetcher, an basic WebSocket in C and C++ and an Error handler for web pages in C

AR-DEV-1:
I have added an basic system information fetcher for different OS (Mobiles included) like the one in chrome where it displays your OS Information, also an basic WebSocket in C and C++ and also an error handler for web pages in C


AR-DEV-1:
Added an logger based on :
https://github.com/gosub-browser/gosub-engine/discussions/242

AR-DEV-1:
Add client side data based on:
https://github.com/gosub-browser/gosub-engine/discussions/243

AR-DEV-1:
Add an configuration store :
https://github.com/gosub-browser/gosub-engine/discussions/240

jaytaph:
@AR-DEV-1 Thanks for your work, but we are a little bit confused on what it exactly is that you are committing? Gosub is a rust based browser engine, and it is not expected to have much (except for some bindings) to include C(++) code. There also seems to be a lot of things in here that we don't understand? What is a config_store? What is client_storage, logger, websocket etc? 

Before continuing, it might be a good idea to join us on the zulip developer chat (chat.developer.gosub.io) and discus what it is you are building, and see if it's something we can use. It would be a waste if you continue your work when it is something we cannot merge into our engine.

PR #322, opened 2023-12-24
First JS Implementation

Sharktheone:
This PR adds a initial JS implementation. This still has many todos, but i think, most of them are outside of the scope from this PR.

I've hacked together some tests, but V8 is not thread-safe and since cargo runs tests in their own thread, they need to be in one large test for now.

In the next PR I probably make the API thread safe (basically they call into the dedicated JS management thread)

What's also still a big TODO is documenting the code. I might do that also today as well as squashing the commits.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,6 +62,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "aliasable"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
+
+[[package]]
 name = "anes"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -389,7 +395,7 @@ dependencies = [
  "clap",
  "criterion-plot",
  "is-terminal",
- "itertools",
+ "itertools 0.10.5",
  "num-traits",
  "once_cell",
  "oorandom",
@@ -410,7 +416,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6b50826342786a51a89e2da3a28f1c32b06e387201bc2d19791f622c673706b1"
 dependencies = [
  "cast",
- "itertools",
+ "itertools 0.10.5",
 ]
 
 [[package]]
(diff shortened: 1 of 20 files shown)


Sharktheone:
I think, this should be ready now. The last thing would be squashing all the commits.

jaytaph:
@Sharktheone would you mind doing the squash first?

Sharktheone:
> @Sharktheone would you mind doing the squash first?

At least with `git rebase` git doesn't like it because of the merge commits. I guess, the simplest thing is to squash them at merge time of this PR.

Sharktheone:
Okay, got it with `git merge --squash` onto another branch. Then resetting the `js` branch and rebasing the other branch again onto `js`

Sharktheone:
Interesting, clippy doesn't print out the [errors](https://github.com/gosub-browser/gosub-engine/actions/runs/7359002095/job/20033098983?pr=322#step:7:392) that are shown in the CI Run. And why does it show errors in the html5 parser?

Is this maybe a newer version of clippy with more checks?

Sharktheone:
> Is this maybe a newer version of clippy with more checks?

Yup, after updating i get the same warnings

jaytaph:
@Sharktheone The clippy issues are because of warnings that are detected in rust v1.75 while in development we are using v1.73. I got the same issue with some other branches. 

I'm not 100% sure what the best way forward would be for this. Should we keep our release (as stated in cargo.yaml) up to date with stable? Or do we keep 1.73 as a baseline (and if so, why 1.73?). Maybe @emwalker or @Kiyoshika have an idea?

jaytaph (approved):

emwalker:
@jaytaph regarding upgrading to 1.75, that shouldn't be a problem.  Here's what I did:

```sh
$ rustup toolchain install stable
```

<details><summary>Code changes</summary>
<p>

```patch
diff --git a/Cargo.lock b/Cargo.lock
index 85e6754..6f7d3f2 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -823,7 +823,7 @@ version = "0.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5444c27eef6923071f7ebcc33e3444508466a76f7a2b93da00ed6e19f30c1ddb"
 dependencies = [
- "windows-sys",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -1134,7 +1134,7 @@ dependencies = [
  "proc-macro-error",
  "proc-macro2",
  "quote",
- "syn 2.0.39",
+ "syn 2.0.44",
 ]
 
 [[package]]
diff --git a/Cargo.toml b/Cargo.toml
index def01c5..28531de 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "gosub-engine"
 version = "0.1.0"
 edition = "2021"
-rust-version = "1.73"
+rust-version = "1.75"
 authors = ["Gosub Community <info@gosub.io>"]
 description = "An HTML5 browser engine written in Rust."
 license = "MIT"
```

</p>
</details>

PR #326, opened 2023-12-29
use chardet to detect encoding

cnsky1103:
use chardet to detect encoding, and also make Confidence::Tentative contain an f32, which is the probability of the encoding

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -155,7 +155,7 @@ impl CharIterator {
         match encoding.1 {
             p if p >= 0.99 => self.confidence = Confidence::Certain,
             p => self.confidence = Confidence::Tentative(p),
-        }
+        };
     }
 
     /// Returns true when the stream pointer is at the end of the stream
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -16,7 +16,7 @@ pub enum Encoding {
 #[derive(PartialEq)]
 pub enum Confidence {
     /// This encoding might be the one we need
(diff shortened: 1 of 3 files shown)


jaytaph:
@cnsky1103 Nice addition!  Could you rebase this PR as this will fix the clippy issues

jaytaph (approved):

src/bytes.rs
```diff
@@ -16,7 +16,7 @@ pub enum Encoding {
 #[derive(PartialEq)]
 pub enum Confidence {
     /// This encoding might be the one we need
-    Tentative,
+    Tentative(f32),
```
Sharktheone, line 19: probably you could save some memory here with a u8, but it is only marginal, so feel free to ignore

Sharktheone (approved):

jaytaph:
@cnsky1103 First PR is always the hardest, but we're almost there. You are signing your commits now, so that's perfect. The problem is that there is still an unsigned commit in the code. We can fix that by rewriting a little bit of history on your side:

      git rebase --exec "git commit --amend --no-edit -n -S" -i f5458c7

This will sign all commits from f5458c7 onwards and will take care of your unsigned commit.

However, this changes the history of your PR. That's ok, but it means that when you push it to github, you must use the following:

     git push origin -f

The -f tells that you want to overwrite the current PR that is on github.

I think if you do that, the PR will be ok.  Thanks for you patience!

PR #327, opened 2023-12-29
impl Iterator for CharIterator

cnsky1103:


jaytaph:
@cnsky1103 Thanks for your PR. 

clippy fails because of we're now using rust 1.75 instead of the configured 1.73 rust version on development. Could you do a rebase on main that fixes these small issues. Then we can merge this PR.

jaytaph (approved):

src/bytes.rs
```diff
@@ -120,6 +120,31 @@ impl Default for CharIterator {
...
 }
 
+impl Iterator for CharIterator {
+    type Item = char;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.eof() || self.position.offset >= self.length {
+            return None;
+        }
+
```
Sharktheone, line 130: Can you maybe add a comment here, saying that this is safe to do because `self.buffer` has the same length as `self.u8_buffer` and it is checked above?

I always think this is blowing up, when we are using `[]` to index on an vec/array

Sharktheone (approved):

jaytaph:
@cnsky1103 very close!  Still a small clippy issue with `return Some(c);` that it wants to be `Some(c)`.

PR #335, opened 2024-01-04
impl Iterator for CharIterator, and add an encoding detect

cnsky1103:


jaytaph (approved):

jaytaph:
@cnsky1103 Thanks for the PR!

PR #336, opened 2024-01-04
Fix id attribute validation

Celisium:
I noticed a few things about this function to validate the `id` attribute of an element:
- The comment links to the outdated W3C HTML5 spec rather than the WHATWG HTML spec.
- The function requires at least one alphabetic character to be present, when no such requirement actually exists. In particular, the spec states that the value can consist of 'just digits' or 'just punctuation', for example. I'm guessing this stems from a misinterpretation of what is meant by 'character'.
- The spec states that the id may not contain any *ASCII* whitespace, but the validation function uses the `is_whitespace` function, which also detects non-ASCII whitespace characters. This means that certain valid IDs would be incorrectly rejected.

### DIFF
--- a/src/html5/parser.rs
+++ b/src/html5/parser.rs
@@ -4369,7 +4369,6 @@ mod test {
         let mut chars = CharIterator::new();
         chars.read_from_str(
             "<div id=\"my id\"></div> \
-             <div id=\"123\"></div> \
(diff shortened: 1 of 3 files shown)


jaytaph (approved):

jaytaph:
@Celisium Thanks for the PR. Looks good!

PR #351, opened 2024-01-25
Initial styles draft

jaytaph:
Tries to setup a system that will be able to compute styles and attach them to nodes.


src/bin/gosub-parser.rs
```diff
@@ -41,14 +39,22 @@ fn main() -> Result<()> {
...
 
-    let document = DocumentBuilder::new_document();
-    let parse_errors = Html5Parser::parse_document(&mut chars, Document::clone(&document), None)?;
+    // Create a new document that will be filled in by the parser
+    let handle = DocumentBuilder::new_document();
+    let parse_errors = Html5Parser::parse_document(&mut chars, Document::clone(&handle), None)?;
 
-    println!("Generated tree: \n\n {document}");
+    let mut binding = Document::clone(&handle);
+    let mut handle_mut = binding.get_mut();
```
Sharktheone, line 47: You can remove one clone here, however you'd need to make handle mutable.

```suggestion
    let mut handle_mut = handle.get_mut();
```
if you don't want to do it, maybe rename the variable.

src/html5/node.rs
```diff
@@ -169,6 +172,7 @@ impl Clone for Node {
             data: self.data.clone(),
             document: Weak::clone(&self.document),
             is_registered: self.is_registered,
+            style: RefCell::new(CssMap::new()),
```
Sharktheone, line 175: You probably should also clone the style here

src/html5/parser/document.rs
```diff
@@ -277,9 +285,9 @@ impl Document {
         DocumentHandle(Rc::clone(&handle.0))
     }
 
-    pub(crate) fn print_nodes(&self) {
-        self.arena.print_nodes();
-    }
+    // pub(crate) fn print_nodes(&self) {
+    //     self.arena.print_nodes();
+    // }
```
Sharktheone, line 290: dead code

src/html5/parser/document.rs
```diff
@@ -457,6 +472,44 @@ impl Document {
...
+
+    pub fn add_useragent_stylesheet(&mut self, stylesheet: CssNode) {
+        self.useragent_stylesheets.push(stylesheet);
+    }
+}
+
+// Walk the document tree with the given visitor
+pub fn visit(doc: &DocumentHandle, visitor: &mut Box<dyn Visitor<Node>>) {
+    let binding = doc.get();
+    let root = binding.get_root();
```
Sharktheone, line 485:
```suggestion
    let root = doc.get();
    let root = root.get_root();
```

I won't make more comments than this one on the `binding` variables, it's better to rename them, however you can leave it like this if you want.

src/js/v8.rs
Sharktheone, line 1: I hope this won't conflict with my PR, it should have a clippy attribute, so it ignores unused variables / unused muts.

Sharktheone (changes requested):

src/css3/walker.rs
```diff
@@ -76,8 +76,8 @@ fn inner_walk(node: &Node, depth: usize, f: &mut dyn Write) -> Result<(), std::i
             }
         }
         NodeType::Comment { .. } => {}
-        NodeType::Cdo => {}
-        NodeType::Cdc => {}
+        // NodeType::Cdo => {}
+        // NodeType::Cdc => {}
```
Sharktheone, line 80: Dead/Duplicated code. You commented these lines out, just to add them below (L270-271)

src/html5/node.rs
```diff
@@ -337,6 +338,65 @@ impl Node {
...
             && ["mi", "mo", "mn", "ms", "mtext"].contains(&self.name.as_str())
     }
+
+    /// Returns true if the node is an element node
+    pub fn is_element(&self) -> bool {
+        if let NodeData::Element(_) = &self.data {
+            return true;
+        }
+
+        false
```
Sharktheone, line 348: That's basically the `matches!` macro
```suggestion
        matches!(self.data, NodeData::Element(_))
```

src/html5/parser/document.rs
```diff
@@ -459,6 +482,39 @@ impl Document {
...
 }
 
+// Walk the document tree with the given visitor
+pub fn visit(doc: &DocumentHandle, visitor: &mut Box<dyn Visitor<Node>>) {
+    let binding = doc.get();
+    let root = binding.get_root();
+    internal_visit(doc, root, visitor);
+}
+
+fn internal_visit(doc: &DocumentHandle, node: &Node, visitor: &mut Box<dyn Visitor<Node>>) {
```
Sharktheone, line 492: You can get rid of the dyn-Box by using impl:

```suggestion
pub fn visit(doc: &DocumentHandle, visitor: &mut impl Visitor<Node>) {
    let binding = doc.get();
    let root = binding.get_root();
    internal_visit(doc, root, visitor);
}

fn internal_visit(doc: &DocumentHandle, node: &Node, visitor: &mut impl Visitor<Node>) {
```

src/html5/parser/document.rs
```diff
@@ -512,7 +568,7 @@ impl Document {
             buffer.push_str("├─ ");
         }
 
-        // buffer.push_str(&format!("({:?}) ", node.id.as_usize()));
+        // buffer.push_str(format!("{} ", node.id).as_str());
```
Sharktheone, line 571: Dead code (as well as 30 lines below)

src/pipeline.rs
```diff
@@ -0,0 +1,60 @@
...
+        return;
+    }
+
+    // Check CSS styles and remove if not renderable
+
+    // Iterate all children from this node
+    for &child_id in &node.children {
+        remove_unrenderable_nodes(rendertree_handle, child_id, _calculator);
+    }
+}
```
Sharktheone, line 60: I'm not sure, if we should always remove style / scripts from the render pipeline. I've seen people giving `<style>` the shown and content-editable attribute to directly edit them in the browser.

src/styles.rs
```diff
@@ -0,0 +1,536 @@
...
+            actual: "".into(),
+        }
+    }
+}
+
+/// Map of all declared values for a single node
+pub type CssMapEntry = HashMap<String, ValueEntry>;
+
+/// Map of all declared values for all nodes in the document
+pub type CssMap = HashMap<NodeId, CssMapEntry>;
```
Sharktheone, line 120: I guess, this can be optimized when we just use one HashMap and basically do this
```suggestion
/// Map of all declared values for all nodes in the document
pub type CssMap = HashMap<(NodeId, String), ValueEntry>;
```

src/styles.rs
Sharktheone, line 1: In this file is like dead code everywhere

src/styles.rs
```diff
@@ -0,0 +1,536 @@
...
+                    let wanted_attr_name = part.name.clone();
+                    let mut wanted_attr_value = part.value.clone();
+                    let mut got_attr_value = current_node
+                        .get_attribute(&wanted_attr_name)
+                        .unwrap_or(&"".to_string())
+                        .to_string();
+
+                    if !current_node.has_attribute(&wanted_attr_name) {
+                        return false;
+                    }
```
Sharktheone, line 337:
```suggestion
                    let wanted_attr_name = part.name.clone();
                    
                    if !current_node.has_attribute(&wanted_attr_name) {
                        return false;
                    }
                    
                    let mut wanted_attr_value = part.value.clone();
                    let mut got_attr_value = current_node
                        .get_attribute(&wanted_attr_name)
                        .unwrap_or(&"".to_string())
                        .to_string();


```

src/styles/converter.rs
```diff
@@ -0,0 +1,378 @@
...
+                        selector.parts.push(CssSelectorPart {
+                            type_: CssSelectorType::Attribute,
+                            name: attr_selector.0.clone(),
+                            matcher: MatcherType::Equals, // @todo: this needs to be parsed
+                            value: attr_selector.2.clone(),
+                            flags: attr_selector.3.clone(),
+                        });
+                    } else {
+                        panic!("Unknown selector type: {:?}", node);
+                    }
```
Sharktheone, line 349: Why not use a match statement here?

```rust
let part = match &*node.node_type {
    NodeType::Ident { value} => CssSelectorPart {
        type_: CssSelectorType::Type,
        value: value.clone(),
        ..Default::default()
    },
    _ => {
        todo!("Implement the rest of the node types");
    }
};

selector.parts.push(part);
                    
```

Sharktheone (changes requested):
I can't say too much about the technical side, as I didn't read through the whole CSS spec

src/styles/converter.rs
jaytaph: how would one do this, when there are multipe types available?

src/pipeline.rs
jaytaph: There is probably some documentation on what to remove, but this is just a simple list that suffice for the time being

src/css3/cssom/cssrule.rs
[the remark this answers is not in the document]
jaytaph: yes, not everything is used yet.

src/bin/style-parser.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+
+struct RgbColor {
+    r: u8,
+    g: u8,
+    b: u8,
+}
+
+fn get_css_color(props: &HashMap<String, String>, prop_name: &str) -> Option<RgbColor> {
```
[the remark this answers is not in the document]
jaytaph: can you give an example for this?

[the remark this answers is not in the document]
Sharktheone: I mean, for example to do something like this:

```rust
struct CSSMap(HashMap<String, String>);

impl CSSMap {
    fn get_color(&self, name: &str) -> Option<RgbColor> {
        ...
    }
}

File: src/styles/converter.rs (hunk 12 above)
Comment (Sharktheone) [reply src/styles/converter.rs:353]: What do you mean?

File: src/styles.rs (hunk 10 above)
Comment (Sharktheone) [reply src/styles.rs:121]: Okay, then it's probably not quite possible. I'm actually not sure, if a HashMap is best here. Maybe we can use a custom data structure for that

File: src/bin/gosub-parser.rs [hunk 14]
```diff
@@ -6,6 +6,8 @@ use gosub_engine::{
 };
 use std::fs;
 use std::process::exit;
+use std::str::FromStr;
+use url::Url;
 
 fn bail(message: &str) -> ! {
```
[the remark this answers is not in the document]
jaytaph: yes, but I think we do this in more spots.. I'll add a ticket for this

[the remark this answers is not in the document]
jaytaph: i'm keeping this for now, added a ticket #382  for it

src/styles/converter.rs
jaytaph: ok.. fixed

jaytaph:
Closed in favor for #392

PR #372, opened 2024-02-14
Updated gitignore file

jaytaph:


Sharktheone:
I'm not sure about that. I generally like to have also editor specific things in `.gitignore`. Else, it can lead to some mistakes by accidentally committing a file you didn't wanted to.
Yes you can edit your `.gitignore` file locally, but normally when you just do one PR, that's not something you normally

jaytaph:
The problem with this is that there are very many different IDE's out there which each got their own set of files that can be either in the root directory (.idea, .vscode etc) or scatered around the source tree (.DS_Store). Instead of having a big ignore file with ALL possibilities, it's usually cleaner to have your IDE be ignored locally. This saves the user time, as we do not need to add your IDE files to EACH project you touch,.. AND it saves the project "time" as we do not need to maintain user preferences.

Sharktheone:
I get your point, however I usually have a template which covers most of of the files, so maintaining should only be getting the newest version of the template.

Sharktheone:
Hmm, that's interesting https://github.com/gosub-browser/gosub-engine/actions/runs/7905929253/job/21579625141?pr=372

PR #379, opened 2024-02-20
Moving stuff to subcrates

jaytaph:
Much of the code is now moved to dedicates (sub)crates in the gosub_* prefix. Each subcrate has it's own list of dependencies.

The main gosub_engine crate should have the main gosub engine library (rlib?) and additional binary applications to test things from the commandline.


emwalker (approved):
LGTM, assuming you don't mind a bunch of build checks failing.

Sharktheone (commented):
I'm not sure about the JS part, you called it gosub-js and kept the directory `web_executor`, I have chosen this name because we may have also some things dedicated to WASM there.

jaytaph:
> I'm not sure about the JS part, you called it gosub-js and kept the directory `web_executor`, I have chosen this name because we may have also some things dedicated to WASM there.

@Sharktheone do you have an alternative name for it:  gosub_webexecutor for instance?

jaytaph:
> LGTM, assuming you don't mind a bunch of build checks failing.

@emwalker fixed the build. I was a bit enthusiastic with removing dependencies which are needed for tests

Sharktheone:
> @Sharktheone do you have an alternative name for it: gosub_webexecutor for instance?


Yes, so either `gosub_webexecutor` or we can keep `gosub_js` and you then just remove the web_executor folder. When we have code for WASM, we add another crate `gosub_wasm`

What might even make sense is to have maybe `gosub_js` which has the traits and some code to connect it with the rest of the engine, but we also have `gosub_v8` which implements the traits from `gosub_js`

jaytaph:
> > @Sharktheone do you have an alternative name for it: gosub_webexecutor for instance?
> 
> Yes, so either `gosub_webexecutor` or we can keep `gosub_js` and you then just remove the web_executor folder. When we have code for WASM, we add another crate `gosub_wasm`
> 
> What might even make sense is to have maybe `gosub_js` which has the traits and some code to connect it with the rest of the engine, but we also have `gosub_v8` which implements the traits from `gosub_js`

Ok if i go for gosub_webexecutor for now. We can move to seperate crates when we implement wasm at a later stage, or even decide to use the same crate for it

blaumeise20:
Hello everyone, I've been silent for quite some time because of personal time reasons, but I have email notifications enabled for this project and I have a question about the naming of the folders: why do you use `gosub-xxx` (with a dash) instead of `gosub_xxx`? Wouldn't it be better to keep the folder name the same as the crate name itself?

jaytaph:
@blaumeise20 I've created a discussion on this: https://chat.developer.gosub.io/#narrow/stream/6-gosub/topic/Crate.20name.20matching.20path/near/5105

Sharktheone (approved):

PR #381, opened 2024-02-21
CI: add print CPU info on tests

Sharktheone:
Print CPU info of the CI runner because one [test](https://github.com/gosub-browser/gosub-engine/actions/runs/7905929253/job/21579625141) has failed, probably because of some (missing) CPU features.
If it ever occurs again, we can check.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,6 +46,8 @@ jobs:
       matrix:
         rust_version: [ stable, beta, nightly, 1.73.0, "stable minus 1 release", "stable minus 2 releases" ]
     steps:
+      - name: Print CPU info
+        run: lscpu
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@master
         with:


jaytaph (approved):

jaytaph:
I'm not sure why the PR author check fails :/

Sharktheone:
I'm actually not sure if this step should be at the beginning

Sharktheone:
> I'm not sure why the PR author check fails :/

Haven't added myself to the `AUTHORS` file

jaytaph:
It's running parallel, but displayed alphabetically..   the reason it fails is because the PR source (yours) is untrusted, and thus is readonly. It cannot create a comment in the PR itself.   I think it might be best to either remove the check and find out how other repo's are dealing with this.

Sharktheone:
What do you think about the order of the `Print CPU info` step though?

jaytaph:
I think the idea is to capture some of the CPU features into a env var so you can use them later on in different steps/jobs?

Sharktheone:
I assume, it's okay for you.

PR #386, opened 2024-02-25
Removed own defined bail and using anyhow instead

### LINKED ISSUE #382: Use bail! macro instead of our own custom bail method
There is a bail! macro we can use but sometimes we use our own custom bail! function that we define. We should get rid of those and use bail! everywhere

jaytaph:
Fixes #382

### DIFF
--- a/src/bin/gosub-parser.rs
+++ b/src/bin/gosub-parser.rs
@@ -16,7 +16,11 @@ fn main() -> Result<()> {
         )
         .get_matches();
 
-    let url: String = matches.get_one::<String>("url").expect("url").to_string();
+    let url = matches
+        .get_one::<String>("url")
+        .ok_or("Missing url")
+        .unwrap()
+        .to_string();
 
     let html = if url.starts_with("http://") || url.starts_with("https://") {
         // Fetch the html from the url
--- a/src/bin/gosub-parser.rs
+++ b/src/bin/gosub-parser.rs
@@ -1,22 +1,28 @@
-use anyhow::{bail, Result};
+use anyhow::bail;
 use gosub_html5::parser::document::{Document, DocumentBuilder};
 use gosub_html5::parser::Html5Parser;
 use gosub_shared::bytes::{CharIterator, Confidence, Encoding};
+use gosub_shared::types::Result;
 use std::fs;
 
 fn main() -> Result<()> {
-    let url = std::env::args()
-        .nth(1)
-        .unwrap_or_else(|| bail!("Usage: gosub-parser <url>"));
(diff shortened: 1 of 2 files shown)


src/bin/gosub-parser.rs
```diff
@@ -1,25 +1,19 @@
...
 
     let html = if url.starts_with("http://") || url.starts_with("https://") {
         // Fetch the html from the url
         let response = ureq::get(&url).call()?;
         if response.status() != 200 {
-            bail(&format!(
+            bail!(&format!(
                 "Could not get url. Status code {}",
                 response.status()
             ));
```
blaumeise20, line 19: The bail macro supports formatting arguments so you don't have to explicitly use the format macro.
(See https://docs.rs/anyhow/latest/anyhow/macro.bail.html)
```suggestion
            bail!("Could not get url. Status code {}", response.status());
```

src/bin/gosub-parser.rs
```diff
@@ -1,28 +1,28 @@
...
+        .version("0.1.0")
+        .arg(
+            clap::Arg::new("url")
+                .help("The url or file to parse")
+                .required(true)
+                .index(1),
+        )
+        .get_matches();
+
+    let url: String = matches.get_one::<String>("url").expect("url").to_string();
```
blaumeise20, line 19: I'd probably use `.ok_or("missing url")?` here instead of `.expect` since it fits the `-> Result<()>` main function better.

src/bin/gosub-parser.rs
```diff
@@ -1,28 +1,28 @@
...
-    let url = std::env::args()
-        .nth(1)
-        .unwrap_or_else(|| bail("Usage: gosub-parser <url>"));
+    let matches = clap::Command::new("Gosub parser")
+        .version("0.1.0")
+        .arg(
+            clap::Arg::new("url")
+                .help("The url or file to parse")
+                .required(true)
+                .index(1),
```
blaumeise20, line 15: Trailing comma looks weird in this case with a single argument (in my opinion).

blaumeise20 (approved):
You could change some small details, everything else looks great 👍

src/bin/gosub-parser.rs
jaytaph: fmt wants to have a comma here.

src/bin/gosub-parser.rs
jaytaph:
```
    let url: String = matches
        .get_one::<String>("url")
        .ok_or("Missing url")
        .unwrap()
        .to_string();
```

Sharktheone (approved):

PR #391, opened 2024-02-26
CI: fix ci not running on all targets

### LINKED ISSUE #390: [CI] It doesn't run all tests anymore
Since #379 CI doesn't run all tests, only the tree construction and tokenizer ones
For example [here](https://github.com/gosub-browser/gosub-engine/actions/runs/8045061843/job/21969685135?pr=383#step:8:450)

Many tests also don't compile anymore, because the use statements aren't working anymore. I noticed this where I added more tests for V8 (#383)

Sharktheone:
Since #379 CI doesn't run all tests, only the tree construction and tokenizer ones
For example [here](https://github.com/gosub-browser/gosub-engine/actions/runs/8045061843/job/21969685135?pr=383#step:8:450)

Many tests also don't compile anymore, because the use statements aren't working anymore. I noticed this where I added more tests for V8 (#383)

fixes #390

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,7 +37,7 @@ jobs:
           path: target
           key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}-${{ matrix.rust_version }}
       - name: Build
-        run: cargo build --verbose
+        run: cargo build --verbose --all --all-features
 
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
           path: target
           key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}-${{ matrix.rust_version }}
       - name: Run tests
-        run: cargo test --verbose
+        run: cargo test --verbose --all --no-fail-fast --all-features
```
Sharktheone, line 71: I'm not sure about the `--no-fail-fast` and `--all-features` though.

jaytaph: that's fine for now

jaytaph (approved):

Sharktheone:
What we maybe could also look into, I think, caching doesn't really work correctly here for the registry. It always needs to download packages. Maybe we can also update that.

Sharktheone:
@jaytaph Can you maybe look at this? https://github.com/gosub-browser/gosub-engine/actions/runs/8086610458/job/22096738161?pr=391#step:8:895

It's failing somewhere in the tree iterator test. Probably a rerun should fix it, but maybe there is something we can fix.

PR #415, opened 2024-03-21
Messed some things up with the render tree

Sharktheone:
Messed some things up with the render tree for example the children. The second thing was if a node isn't a element it doesn't get added.

I think, there are still a few things, but I don't know where.


Sharktheone:
Already fixed in #416

PR #416, opened 2024-03-22
Rendertree fixups

jaytaph:
- colors are moved from u8 to f32 so we can use decimals (in case of hsl).
- using color transform to automatically convert hsl() and rgb() type colors
- fixes up some issues with the render-tree


crates/gosub_styling/src/render_tree.rs
```diff
@@ -11,72 +16,68 @@ use crate::css_values::{
...
+    /// Returns the root node of the render tree
+    pub fn get_root(&self) -> &RenderTreeNode {
+        self.nodes.get(&self.root).expect("root node")
+    }
 
-impl RenderTreeNode {
-    /// Returns true if the node is an element node
-    pub fn is_element(&self) -> bool {
-        matches!(self.data, NodeData::Element(_))
+    /// INserts a new node into the render tree, note that you are responsible for the node id
```
Sharktheone, line 52:
```suggestion
    /// Inserts a new node into the render tree, note that you are responsible for the node id
```

Sharktheone (approved):

PR #431, opened 2024-04-10
wasm bindings

Sharktheone:
This PR allows the engine to run inside wasm, which means we can run our browser inside of another browser.

image

Some components unfortunately needed to be disabled, for example V8, networking and timing.
V8 doesn't seem to be able to run inside wasm, however mozjs and quickjs can. Networking is as easy as replacing ureq with the browser's `fetch` api. For now I leave it up to the caller to provide the HTML/CSS stream. Timing probably also just needs to be ported to make use of JS APIs.

I want to merge the first renderer, because it will have some conflicts and this already has some conflicts. Additionally they are more easily to fix here than in the other PR.


Sharktheone:
Also works now for the renderer
image

jaytaph (approved):

Sharktheone:
Now it has conflicts again because it still has the last commit from #438

PR #461, opened 2024-05-04
Trying a new css system

jaytaph:
This (draft) PR implements a CSS system that allows to easily fetch properties from a renderer perspective.

We already are able to parse (internal and external) stylesheets into CSS ASTs. These are automatically attached to a document(handle). 

From this point, a useragent can generate a render-tree which is a document but without any invisible or unrenderable nodes. This render tree should have for each node its CSS properties so you can easily find any CSS property on that node.

If there are no properties for a specific node, it will check if the queried property is inheritable (like `font`). if so, it will traverse down the nodes to find the property there. If none are found, or the property is not inheritable, it will use the default value for that property as defined.

In order to make sure that properties have correct values, we can use the [CSS documention's Value Syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax), where many properties are defined. This means that whenever we have a syntax for a property, we can check if any given value that is defined for that property actually matches the syntax  (for instance: the `color` does not understand `3em`).

The system is also able to generate shorthand properties. For instance, "border" is a property that is shorthand for `border-color`, `border-size` and `border-style`. These properties are again even shorthand for `border-top-*`, `border-right-*`, etc. 

So when have defined a `border: 1px solid black`, we can fetch `border-bottom-style`, and retrieve the value "`solid`".

To do or to check:
- [X] Syntax compilation
- [ ] Value to syntax matching
- [ ] Inheritance
- [ ] Initial values
- [ ] Shorthand
- [ ] API to fetch values

Future work:

In a later PR, we will change this API to actually return enums that allows to do better matching. For instance, `border-bottom-style` property will return not a `CssValue()`, but a `CssBorderStyle` enum which could be something like:

```
enum CssBorderStyle {
   Solid,
   Striped,
   ...
   Width(float),
}
```

This way, we have the rust compiler to check if we are forgetting and not handling certain CSS3 values for given properties.


Sharktheone:
Is this an updated version of #433 or something completely new?

jaytaph:
> Is this an updated version of #433 or something completely new?

This is new.. I'll close the other one

PR #467, opened 2024-05-11
make clippy more strict

Sharktheone:
This PR makes clippy lints more strict, I didn't (yet) fix all errors that were created by changing the configuration.


jaytaph:
Lets close this one for now. I think we cannot reasonably salvage this

Sharktheone:
Yes, and it has so many conflicts. 

But maybe we should keep it as an reminder. The main problem is that Clippy can't auto fix lints if there are more than 10 or so. I think we should open an issue on the Clippy side.

And I mean we probably have a couple thousand with this project.

jaytaph:
Setting up a new issue #752

PR #492, opened 2024-06-19
Changing the bytes/CharIterator to the ByteStream

### LINKED ISSUE #384: use gosub_shared::ByteStream instead of gosub_shared:Byte
The gosub_shared::ByteStream works a bit better (?) than the byte structure. The CSS3 uses the ByteStream, but the HTML5 parser still uses Bytes (and maybe some other places too)..  We should move all streams to the ByteStream, and later on improve this stream reader (for instance, to work better with encoding)

jaytaph:
This commit mostly removes all the bytes::CharIterator code and replaces it with the ByteStream. This is a more clean setup of reading a stream of characters taken from a stream of bytes. Now both the HTML5 and CSS3 parser will work with the same bytestream system which makes things a little more clear in the end.

As an additional bonus: attributes are now checked during testing, and all tests seem to pass!

Fixes #384


crates/gosub_css3/src/tokenizer.rs
```diff
@@ -906,37 +896,29 @@ impl<'stream> Tokenizer<'stream> {
...
+        self.stream.offset()
     }
 
-    pub fn slice(&self, start: usize, end: usize) -> String {
+    pub fn slice(&mut self, start: usize, end: usize) -> String {
+        let old_pos = self.stream.offset();
+        self.stream.seek(start);
+
+        // todo: this is not efficient
         let mut s = String::new();
```
Sharktheone, line 907: You could use `String::with_capacity(end - start)` (or maybe even only do the calculation once (or maybe the compiler does that?))

crates/gosub_html5/src/tokenizer.rs
```diff
@@ -337,25 +323,28 @@ impl<'stream> Tokenizer<'stream> {
...
+                    match c {
+                        Ch('/') => {
+                            self.temporary_buffer.clear();
+                            self.state = State::RCDATAEndTagOpen;
+                        }
+                        _ => {
+                            self.consume('<');
+                            self.stream_prev();
+                            self.state = State::RCDATA;
+                        }
```
Sharktheone, line 347: I guess, this should be a `if let` match pattern again, in my clippy PR, this actually will be an warning.

crates/gosub_html5/src/tokenizer.rs
```diff
@@ -419,32 +408,32 @@ impl<'stream> Tokenizer<'stream> {
...
+                    match c {
+                        Ch('/') => {
+                            self.temporary_buffer.clear();
+                            self.state = State::RAWTEXTEndTagOpen;
+                        }
+                        _ => {
+                            self.consume('<');
+                            self.stream_prev();
+                            self.state = State::RAWTEXT;
+                        }
```
Sharktheone, line 436: same here

crates/gosub_html5/src/tokenizer.rs
```diff
@@ -1899,66 +1952,77 @@ impl<'stream> Tokenizer<'stream> {
...
-                        self.state = State::CDATASection;
+                    match c {
+                        Ch(']') => {
+                            self.state = State::CDATASectionEnd;
+                        }
+                        _ => {
+                            self.consume(']');
+                            self.stream_prev();
+                            self.state = State::CDATASection;
+                        }
```
Sharktheone, line 2009: should also be a if let

Sharktheone (changes requested):
Only some minor things, but the rest looks good.

crates/gosub_html5/tests/data/html5lib-tests/tokenizer/contentModelFlags.test
```diff
@@ -1,52 +1,5 @@
...
-"output":[["Character", "foo</xmp"]]},
-
-{"description":"End tag closing RCDATA or RAWTEXT (ending with slash)",
-"initialStates":["RCDATA state", "RAWTEXT state"],
-"lastStartTag":"xmp",
-"input":"foo</xmp/",
-"output":[["Character", "foo"]],
-"errors":[
-    { "code": "eof-in-tag", "line": 1, "col": 10 }
-]},
```
emwalker, line 49: What happened to these test files?

emwalker (approved):
Seems good to me.  A pity to see the test files go away.

crates/gosub_html5/tests/data/html5lib-tests/tokenizer/contentModelFlags.test
jaytaph: that should not be... I'll check what is going on..

jaytaph: @emwalker I've added them back.. I accidentally committed edits in the tests and failed to restore them first

crates/gosub_css3/src/tokenizer.rs
jaytaph: fixed

crates/gosub_html5/src/tokenizer.rs
jaytaph: fixed

crates/gosub_html5/src/tokenizer.rs
jaytaph: fixed

crates/gosub_html5/src/tokenizer.rs
jaytaph: fixed

Sharktheone (approved):

PR #509, opened 2024-07-01
Added different platforms to matrix

### LINKED ISSUE #508: Make sure we do CI also on windows (and mac) platforms

jaytaph:
~~Fixes #508~~

Actually, removes a few more platforms from the matrix in order to speed up CI but mostly not to run out of CI/minutes


Sharktheone (approved):

Sharktheone:
Maybe we should make the rustversion matrix smaller, so we don't have so many checks?

.github/workflows/ci.yaml
```diff
@@ -80,8 +84,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust_version: [ stable, beta, nightly, 1.73.0, "stable minus 1 release", "stable minus 2 releases" ]
-        os: [ ubuntu-latest, windows-latest, macos-latest ]
+        rust_version: [ stable, nightly, 1.73.0 ]
```
Sharktheone, line 87: I guess clippy is okay if it there only is stable and nightly, since they probably don't remove checks or at least not without a reason

```suggestion
        rust_version: [ stable, nightly ]
```

Sharktheone (changes requested):

.github/workflows/ci.yaml
Sharktheone: hmm, actually I'm not even sure if clippy has that much of a difference between stable and nightly, since you can enable and disable different categories. So nightly is probably just `clippy::nursery`

Sharktheone (approved):

PR #536, opened 2024-08-04
Put Layouting behind a trait for modularity

Sharktheone:
This PR add the Layout trait, it is currently only implemented by taffy, other systems may follow, maybe we can even write our own at some point.

Note: This PR still need some fixes here and there because of the rebase with main.


jaytaph:
I think we can squash this before merging

Sharktheone:
> I think we can squash this before merging

I'm not sure, I've already squashed some commits, you can see them in #543, I guess we could squash the layout trait, the "moving away" from "raw" taffy, and the implementation of the layout trait for taffy commits together, but I'd like to keep them separate, so it is easier to revert them if it is needed.

The two other commits, probably should also be separate, since the fmt commit would make it more difficult to see the "main" change form the other commits.

jaytaph (approved):

PR #543, opened 2024-08-06
Support Inline elements

Sharktheone:
This PR tries to more correctly support inline elements. Currently inline elements are just shown below each other (behave like block elements). This PR at least puts them on the same line, it still isn't right most of the time, for example it would just overflow and not try to position it's elements on the next line. Text also doesn't get spaces between text in `<span>` and a `<p>` tags or text that isn't wrapped in any additional tag.

image

Note: This PR contains work from #536 and so requires it to be merged before this PR, then this PR will also be way smaller.

### DIFF
--- a/crates/gosub_html5/src/parser/document.rs
+++ b/crates/gosub_html5/src/parser/document.rs
@@ -287,6 +287,10 @@ impl Document {
         DocumentHandle(Rc::new(RefCell::new(Self::new(location))))
     }
 
+    pub fn peek_next_id(&self) -> NodeId {
+        self.arena.peek_next_id()
+    }
+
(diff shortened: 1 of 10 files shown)


jaytaph (approved):

jaytaph:
This needs a rebase before we merge it

Sharktheone:
Yeah, the actual change from this PR are the last 5 commits, the rest is basically just #536, but I wanted to have this separate, since it isn't strongly related to each other.

I haven't squashed the 5 other commits yet, because I want to throw away the work from the trait thing and pull it in from main again, then I can squash them. I just wanted to create the PR already, so it is "out"

jaytaph (approved):

PR #552, opened 2024-08-17
Added alias table for deprecated css properties

### LINKED ISSUE #551: Add aliased properties
```
Sharktheone: We also have another problem, there are some properties that are aliases for new properties, for example font-stretch this is a legacy alias for font-width.

It looks like, these aliases have an empty syntax value, so maybe we can get a list of them via searching for empty syntaxes and make a list to what they resolve manually.

Sharktheone: There are 71 of them... But most of them are -webkit-* i guess, we could just remove the -webkit- and have the new property
```

jaytaph:
Fixes #551 

- sorting properties alfabetically,
- added an alias table so we can fetch the syntax from the deprecated properties
- hardcoded skipping some properties (stop-color, stop-opacity)
-

### DIFF
--- a/crates/gosub_styling/tools/generate_definitions/main.go
+++ b/crates/gosub_styling/tools/generate_definitions/main.go
@@ -47,23 +47,9 @@ func main() {
 	})
 
 	for _, property := range webrefData.Properties {
-
 		if property.Syntax == "" {
-			alias, err := webref.GetAlias(property.Name)
-			if err != nil {
-				log.Panic("failed to get alias syntax for property: " + property.Name)
-			}
-
-			for wdp := range webrefData.Properties {
-				if webrefData.Properties[wdp].Name == alias {
-					property.Syntax = webrefData.Properties[wdp].Syntax
-					break
-				}
-			}
-
-			if property.Syntax == "" {
-				log.Panic("failed to get alias syntax for property: " + property.Name)
-			}
+			// Skip any empty syntax, as they will be filled later with alias table
+			continue
 		}
 
 		prop := utils.Property{
@@ -124,6 +110,23 @@ func main() {
 			Descriptors: descriptors,
 			Values:      atRule.Values,
 		})
+
+		for _, property := range webrefData.Properties {
+			if property.Syntax != "" {
+				// Skip properties with syntax
+				continue
+			}
+
+			alias, err := webref.GetAlias(property.Name)
+			if err != nil {
+				log.Panic("failed to get alias syntax for property: " + property.Name)
+			}
+
+			data.PropAliases = append(data.PropAliases, utils.PropAlias{
+				Name: property.Name,
+				For:  alias,
+			})
+		}
 	}
 
 	data.Selectors = webrefData.Selectors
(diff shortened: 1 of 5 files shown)


crates/gosub_styling/tools/generate_definitions/main.go
```diff
@@ -41,7 +42,30 @@ func main() {
 		len(webrefData.Properties), len(webrefData.Values), len(webrefData.AtRules), len(webrefData.Selectors),
 	)
 
+	sort.Slice(webrefData.Properties, func(i, j int) bool {
+		return webrefData.Properties[i].Name < webrefData.Properties[j].Name
+	})
+
```
Sharktheone, line 48: I'm not sure if we want to sort the properties alphabetically, it might make more sense to sort them after its spec.

crates/gosub_styling/tools/generate_definitions/main.go
```diff
@@ -41,7 +42,30 @@ func main() {
...
+		if property.Syntax == "" {
+			alias, err := webref.GetAlias(property.Name)
+			if err != nil {
+				log.Panic("failed to get alias syntax for property: " + property.Name)
+			}
+
+			for wdp := range webrefData.Properties {
+				if webrefData.Properties[wdp].Name == alias {
+					property.Syntax = webrefData.Properties[wdp].Syntax
```
Sharktheone, line 59: I wouldn't straight up copy the syntax, I think it makes sense to also add a property that it is an alias and which prop it aliases. So we can copy the value in the matching part directly to the aliased prop.

So when we have `-webkit-mask`, we copy the property to `mask` in the matcher. Basically at the same point where we resolve shorthands.

This makes it easier in the mask implementation, since we only need to check for `mask-clip` and not also `-webkit-mask-clip`

crates/gosub_styling/tools/generate_definitions/webref/webref.go
```diff
@@ -444,3 +450,83 @@ func ProcessExtraValues(values []WebRefValue, pd *ParseData) {
...
+	"-webkit-transition-delay":           "transition-delay",
+	"-webkit-transition-duration":        "transition-duration",
+	"-webkit-transition-property":        "transition-property",
+	"-webkit-transition-timing-function": "transition-timing-function",
+	"-webkit-user-select":                "user-select",
+	"font-stretch":                       "font-width",
+	"grid-column-gap":                    "column-gap",
+	"grid-gap":                           "gap",
+	"grid-row-gap":                       "row-gap",
+}
```
Sharktheone, line 524: Might make sense to put those into a json file.

Sharktheone (changes requested):

crates/gosub_styling/tools/generate_definitions/webref/webref.go
jaytaph: i don't want to tag along resource files if it's just a single one. We can change that later

crates/gosub_styling/tools/generate_definitions/main.go
jaytaph: that might make sense

crates/gosub_styling/tools/generate_definitions/main.go
jaytaph: how do you mean?

crates/gosub_styling/tools/generate_definitions/main.go
jaytaph: i'll move the aliases to a separate "alias" listing, and pick them up in the matcher

crates/gosub_styling/tools/generate_definitions/main.go
Sharktheone: When you are looking for a specific prop, and it's syntax, the computed props might be nearer then, since they are all defined in the same spec.

I guess for expanded props it isn't really an issue, since of `grid` or `mask`, they all also start with `grid-` or `mask`.

PR #558, opened 2024-08-19
Add `Fetcher` which automatically expands urls

Sharktheone:
The Fetcher automatically handles URLs expanding and requesting, so when it encounters a relative URL, it will make use of the base URL and "expands" the URL.

It also handles different schemes, like `https`, `http` and `file` we can add more later on.

For now, I only make use of it in the image requesting. I guess in the future, we want to build a fetcher, when we download the HTML and pass it to the different subcomponents, like drawing and so on.

The fetcher can then also make some caching, but only for the lifetime of the page, so if we have for example the same image twice on the page, it doesn't try to download it two times but only once.

### DIFF
--- a/crates/gosub_net/src/http/headers.rs
+++ b/crates/gosub_net/src/http/headers.rs
@@ -12,8 +12,18 @@ impl Headers {
         }
     }
 
-    pub fn set(&mut self, key: &str, value: &str) {
+    pub fn with_capacity(capacity: usize) -> Headers {
+        Headers {
+            headers: HashMap::with_capacity(capacity),
+        }
+    }
+
+    pub fn set_str(&mut self, key: &str, value: &str) {
         self.headers.insert(key.to_string(), value.to_string());
+    }
+
+    pub fn set(&mut self, key: String, value: String) {
+        self.headers.insert(key, value);
     }
 
     pub fn get(&self, key: &str) -> Option<&String> {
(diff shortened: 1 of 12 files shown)


jaytaph (approved):

crates/gosub_net/src/http/headers.rs
```diff
@@ -12,10 +12,20 @@ impl Headers {
...
     }
 
-    pub fn set(&mut self, key: &str, value: &str) {
+    pub fn with_capacity(capacity: usize) -> Headers {
+        Headers {
+            headers: HashMap::with_capacity(capacity),
+        }
+    }
+
+    pub fn set_str(&mut self, key: &str, value: &str) {
```
blaumeise20, line 21: Hello hello, once again a random comment after long silence 😉 I am actually very actively following the activity here on GitHub, but don't have a lot of time for communication. This has already been merged, but a little note for the future: Rust allows you to use `impl Trait` types as parameters, so you don't need separate methods `set` and `set_str`. That is widely used in the standard library too. The method would look like this:
```rs
pub fn set(&mut self, key: impl AsRef<str>, value: impl AsRef<str>) {
    self.headers.insert(key.as_ref().to_string(), value.as_ref().to_string());
}
```
When calling it with owned strings, the compiler will completely optimize this away into just `self.headers.insert(key, value);`. Because `impl` is just syntax sugar for bounded generics, the method actually gets monomorphized, resulting in the same 2 methods, but you are able to "generate" more by calling it with `&str, String` for example.
```rs
req.header.set("Content-Type", format!("text/{extension}")); // bad, don't do this, just an example
```

Sharktheone: Ah, yeah. That looks interesting. I guess we could also just directly use `impl Into<String>`. I might change that when I'm adding more functionality to the fetcher. But thanks for the suggestion!

PR #560, opened 2024-08-21
Make matcher 120000% faster

Sharktheone:
The current matcher takes 55secs to validate the css from news.ycombinator.com (most of it is from useragent.css). Now it only takes ~40ms, and it takes 2.5 secs from start to the first rendered frame.

### DIFF
--- a/crates/gosub_styling/src/property_definitions.rs
+++ b/crates/gosub_styling/src/property_definitions.rs
@@ -1,10 +1,13 @@
+use std::collections::HashMap;
+use std::sync::LazyLock;
+
+use log::warn;
+
+use gosub_css3::stylesheet::CssValue;
+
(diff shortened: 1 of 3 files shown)


crates/gosub_styling/src/property_definitions.rs
```diff
@@ -324,9 +327,13 @@ impl CssDefinitions {
     }
 }
 
+pub static CSS_DEFINITIONS: LazyLock<CssDefinitions, fn() -> CssDefinitions> =
+    LazyLock::new(pars_definition_files);
+
 /// Parses the internal CSS definition file
-#[memoize]
-pub fn parse_definition_files() -> CssDefinitions {
+fn pars_definition_files() -> CssDefinitions {
```
jaytaph, line 334: typo for parse_definition_files() ?

PR #562, opened 2024-08-24
Added improvements on the CSS3 user stylesheet parser

jaytaph:
- The bytestream has an option configuration you can give it.
- removed confidence property for the bytestream. 
- optimized the bytestream so it doesn't need any more copies
- bytestream can deal with different encodings (UTF8 / ASCII / UTF16LE / UTF16BE)
- Implemented some missing parser syntax
- Fixed location of tokens (they had the wrong line/col)
- html5 tokenizer and parser returns location information


Sharktheone (approved):

crates/gosub_shared/src/lib.rs
```diff
@@ -2,6 +2,9 @@
 //!
 //! This crate supplies a lot of shared functionality in the gosub engine.
 //!
+
+extern crate core;
+
```
Sharktheone, line 7: Why is this needed? Shouldn't you be able to just use the `core` crate? If not, it also should be available through the std crate

Sharktheone:
Hmm... looks like the parse_url test is sus
https://github.com/gosub-io/gosub-engine/actions/runs/10576700889/job/29303510579?pr=562#step:7:617

crates/gosub_shared/src/lib.rs
jaytaph: it's an automatic thing that goland seems to add. It also does not do a very good check, as sometimes it adds it twice.  I can remove this.

PR #583, opened 2024-09-04
Layout/text

Sharktheone:
We now use parley for text layout. So, we can now do like wrapping and so on. It still isn't 100% correct, since there are things I haven't yet implemented:
- flattening of nested inline trees
  > This should fix two issues, first line wrapping of nested text, eg when you have
  > ```html
  >    <p>Hello, world!<span> some other very very long text, that could be partially wrapped to another line</span>
  >    </p>
  > ```
  > currently if the width of the whole span is to wide to fit, it would get it's own line, we would not try to still fit the first few words of the span on the same line of "Hello, world!"

- Text Baseline offset
  > if you have a inline-block box with text, it tries to align the text inside the inline-block box with the text on the outside. Our current implementation can't do this.

I will implement both of these two things in a next PR, but for now I have enough of text rendering/layouting 😅


Sharktheone:
Ah, this also adds the ability to use `bold` and italics and all other variable font axes. 

I don't know if I implemented all css shortcuts, but weight and style should work. `font-variation-settings` is also still left to "parse" the css values into something that it can use.

jaytaph (approved):

PR #587, opened 2024-09-07
disable padding for inline elements, only inline-block elements get padding

Sharktheone:
Many (nested) inline elements got too large because they got padding from somehwere, this is wrong. Inline elements aren't affected by margin, padding, width and height properties. This PR fixes some of those issues with padding.

Note: This is more of a temporary fix, since the current implementation still lays out inline elements with the block algorithm


crates/gosub_taffy/src/style/parse_properties.rs
```diff
@@ -20,11 +20,16 @@ pub fn parse_display(node: &mut impl Node) -> (Display, crate::Display) {
         return (Display::Block, crate::Display::Taffy);
     };
 
+    println!("display: {}", value);
```
jaytaph, line 23: This needed?

Sharktheone: Eh, no. I probably missed this

jaytaph (approved):

PR #597, opened 2024-09-13
Let images set their own size

Sharktheone:
Images can now set their own size, by their actual size.

This is also the first time we can update the style after the initial render 🎉

image

In the screenshot the image is obviously too large for the whole screen, but it's size is inferred from the actual size of the png image

This PR also fixes the layouting of inline boxes that are alone in an inline context. Previously those weren't updated.


crates/gosub_taffy/src/compute/inline.rs
```diff
@@ -102,12 +104,18 @@ pub fn compute_inline_layout<LT: LayoutTree<TaffyLayouter>>(
             tree.update_style(*child);
 
             let size = if let Some(cache) = tree.0.get_cache(*child) {
+                dbg!(&cache.style);
+
+                dbg!(cache.display);
```
jaytaph, line 109: these still need to be here for now?

Sharktheone: No, let me remove those...

src/bin/renderer.rs
```diff
@@ -17,6 +17,10 @@ type Drawer = TreeDrawer<Backend, Layouter>;
 type Tree = RenderTree<Layouter>;
 
 fn main() -> Result<()> {
+    unsafe {
+        std::env::remove_var("WAYLAND_DISPLAY"); //DO NOT COMMIT!
```
jaytaph, line 21: DO NOT COMMIT   :p

jaytaph (approved):

PR #599, opened 2024-09-13
Layout/text inline improvements

Sharktheone:
Some improvements for inline and text layout


Sharktheone:
#600 should be merged before to fix the tests

jaytaph (approved):

PR #609, opened 2024-09-19
Implement text variation

Sharktheone:
The `font-variation-settings` property can now be used.

This PR requires both #608 and #607 to actually work.


jaytaph (approved):

Sharktheone:
Just noticed that this also requires #608 for merging, because it needs the `is_comma` method on a Value.

PR #633, opened 2024-10-15
issue 393: Use less unwrap and for loops in `colors.rs`

NicholasQNguyen:
Hi there! This is my first PR so please let me know if I need to improve anything.

This PR changes a lot of the `.unwrap()` called in `colors.rs`. It instead puts everything together in a method and called `expect()` with an error message (hopefully the expect shouldn't fail either!).

It also gets rid of some of the for loops in the code for iters. Just cleaner and easier to understand I think.

This is part of issue 393 about getting rid of the unwraps in the codebase. This is a step in that process!

Always open to feedback!

### DIFF
--- a/crates/gosub_css3/src/colors.rs
+++ b/crates/gosub_css3/src/colors.rs
@@ -5,7 +5,6 @@ use std::str::FromStr;
 use colors_transform::Color;
 use colors_transform::{AlphaColor, Hsl, Rgb};
 use lazy_static::lazy_static;
-
 // Values for this table is taken from https://www.w3.org/TR/CSS21/propidx.html
 // Probably not the complete list, but it will do for now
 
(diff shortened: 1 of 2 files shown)


jaytaph (approved):

crates/gosub_css3/src/colors.rs
```diff
@@ -97,17 +96,12 @@ impl From<&str> for RgbColor {
...
 }
 
 fn get_hex_color_from_name(color_name: &str) -> Option<&str> {
-    for entry in crate::colors::CSS_COLORNAMES.iter() {
-        if entry.name == color_name {
-            return Some(entry.value);
-        }
-    }
-    None
+    Some(crate::colors::CSS_COLORNAMES.iter().find(|entry| entry.name == color_name)?.value)
```
Sharktheone, line 104: The only thing that might be a bit stylistic unpretty is this here, the line is relatively long. But the rest looks good!

Sharktheone (approved):

Sharktheone:
It seems like most checks are failing. The first and easiest is the fmt check. You can fix it with just running `cargo fmt` the other ones are clippy, these are some checks. You can see what it doesn't like by running `cargo clippy --all` and seeing what lints there are.

Sharktheone:
The build checks are a bit annoying currently. Github updated their Ubuntu image, which removed the libsqlite package by default, this is something I need to fix.

NicholasQNguyen:
Cool let me get on that! Also I plan to squash my commits to no worries about there being a million commits lol

crates/gosub_css3/src/colors.rs
jaytaph: I think it would be a bit nicer if it's separated on different lines in this case.. but we should use 120char (or even 140 ftm) line lengths..  we should really need to see if we can somehow enforce this in the linting check on CI probably..   It works quite well in go, but somehow it doesn't play nice with rust

Sharktheone:
The last thing it doesn't like is that your commits aren't signed. In case you need help, here is how you can do it: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

NicholasQNguyen:
alright spent way too long but figured out how to verify my commits lol. Please rereview and then hopefully this is good!

Sharktheone (approved):

jaytaph (approved):

NicholasQNguyen:
one last cleanup merge. also for the ci pipeline to run again

jaytaph:
@NicholasQNguyen Thanks for the PR!

I need to manually merge this because it uses some old required tests we don't use anymore (ubuntu-latest vs ubuntu-24.04).

PR #634, opened 2024-10-15
fix CI on ubuntu 24 by installing libsqlite3

Sharktheone:


jaytaph (commented):
I would prefer pinning the ubuntu version..  so instead of latest use ubuntu-22.04  or ubuntu-24.04  (in the last case, we at least need to install libsqlite, and maybe others)

jaytaph (approved):

PR #635, opened 2024-10-15
Fix scrolling behavior for short pages

### LINKED ISSUE #618: Scrolling a page that isn't 100vh high, causes it to stuck at the bottom after the first scroll
Pre scroll:
image
After scroll:
image

DynamiteC:
Fixes #618 

Addressed an issue where scrolling a page shorter than 100vh would cause it to get stuck at the bottom after the first scroll. Thanks to the maintainers for guiding me to the correct rendering logic where adjustments were needed to properly calculate scroll limits. 

- Updated scroll handling logic in `draw` function.


DynamiteC:
fixes #618

PR #636, opened 2024-10-15
Fixes scrolling behavior for short pages

### LINKED ISSUE #618: Scrolling a page that isn't 100vh high, causes it to stuck at the bottom after the first scroll

DynamiteC:
Fixes #618 

Addressed an issue where scrolling a page shorter than 100vh would cause it to get stuck at the bottom after the first scroll. Thanks to the maintainers for guiding me to the correct rendering logic where adjustments were needed to calculate scroll limits properly.

- Updated scroll handling logic in draw function.
- I added .idea to .gitignore to avoid adding JetBrains files to commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ settings.db
 
 # C API tests (ignore everything but source files)
 *.dSYM
+
+#ignore jetbrains ide folders
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,3 @@ settings.db
 
 # C API tests (ignore everything but source files)
 *.dSYM
-
-#ignore jetbrains ide folders
-.idea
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -13,3 +13,6 @@ settings.db
 
 # C API tests (ignore everything but source files)
 *.dSYM
+
+#ignore jetbrains ide folders
+.idea
```
Sharktheone, line 18: I think we specifically removed IDE specific folders in the git ignore for some reason

DynamiteC: In rustRover While I was trying to commit, it accidentally added .idea to stage so I had to stash that. So should I remove that?

jaytaph: yes please..  IDE and OS specific files should be added to your own gitignore instead of doing this in the project's.

DynamiteC: Understood, I have reverted.

Sharktheone (approved):

jaytaph (approved):

jaytaph:
LGTM. Thank @DynamiteC for the PR..

DynamiteC:
@jaytaph @Sharktheone Thank You!

Issue #393, opened 2024-02-29
Deal with unwrap()


jaytaph:
There are currently 108 `unwrap()` calls found in the code. They should be removed and replaced with better checking before using the unwrapped values, as this can go wrong

Sharktheone:
This probably shouldn't be only about `unwrap()`, but also about `panic!()`. It doesn't need to be completely `no_panic` as i suggested in #380, since that also restricts the usage of `array[index]`.

jaytaph:
Closing as duplicate of #959, which supersedes this with a more recent and comprehensive tracking of unwrap() removals.

PR #638, opened 2024-10-16
#393 Cleaning up some methods in color.rs and getting rid of more unwraps

NicholasQNguyen:
Slowly clanking out a lot of the unwraps used in the codebase. Decided to ignore all unwraps used in tests since that's fine.

a lot of these were `.lock().unwrap()` which i just changed to an `.expect("Poisoned")` since that's the only possible error. Just think it's a little clearer what is going on. But of course always open to feeback


NicholasQNguyen:
it didn't pass some tests so imma take this back and cook on it some more

Issue #521, opened 2024-07-21
Test fixture root is not correct for some tests


jaytaph:
When running `make test`, the system will run ./src/bin/parser-test.rs. This tries to find the tests in FIXTURE_ROOT, but they are not present there. However, the tests in the gosub_html5 crate, do find these tests at the correct position, mainly because the FIXTURE_ROOT is a relative path.

We somehow must make this more consistent so we can run both `make test`, and have the CI pass

aemreaydin:
Hello everyone!
First time here, I would like to work on this to get familiar with the project.

I think there are a couple ways this could be done reliably that I can think of.
1. Change FIXTURE_ROOT variable to use `env!("CARGO_MANIFEST_DIR")` and concat the rest of the path
2. Find base project root (possibly the Cargo.lock) and use it as the base path and generate the FIXTURE_ROOT from there.

jaytaph:
Both options seem reasonable. I would opt for the least amount of change for the enduser (so option 2),.. but having option 1 as a backup (so you can override if the system cannot find it for whatever reason) might be nice too.. I just don't know if that is a lot of work or not..

aemreaydin:
I'm just able to respond as I was away last week. Honestly the simplest change is the first one as it is basically changing the FIXTURE_ROOT variable.
Second one would be more robust way to do it but if we're not using it anywhere else I don't see the need to add the functionality.

aemreaydin:
I was reading the docs and I saw in the binaries docs [parser-test](https://github.com/gosub-io/gosub-engine/blob/main/docs/binaries.md#parser-test), this issue is still listed. Should I remove it in this PR?

jaytaph:
@aemreaydin I didn't catch your comment until now.  Happy to receive a PR for this!

jaytaph:
Closing as fixed. The parser-test binary now delegates to gosub_html5::testing::tree_construction::fixture::read_fixtures(), which uses FIXTURE_ROOT = concat!(env!("CARGO_MANIFEST_DIR"), "/./tests/data/html5lib-tests") — a compile-time absolute path anchored to the crate manifest directory. This works correctly regardless of the working directory, so the original inconsistency between make test and the crate tests no longer exists.

PR #659, opened 2024-10-28
#521 - Updated FIXTURE_ROOT path

aemreaydin:
This change makes the test fixture path resolution more robust by using Cargo's manifest directory environment variable instead of a relative path. This ensures tests can be run from any directory while correctly locating the test data files.


jaytaph (approved):

jaytaph:
Fixes #521

PR #666, opened 2024-11-04
Add Token tests

yt2b:
I've added the fllowing Token function tests.
- is_mixed
- is_mixed_null
- get_location
- set_location

PR #667, opened 2024-11-05
Add Token tests

yt2b:
Same PR(#666) with signed commits.

I've added the following Token function tests.
- is_mixed
- is_mixed_null
- get_location
- set_location


jaytaph:
@yt2b Thanks for the PR!

PR #669, opened 2024-11-11
Refactor LocationHandler

yt2b:
I've refactored LocationHandler.
- changed line_endings's data type from HashMap to Vec.
- renamed line_endings  column_stack.

And I've added test code.

### DIFF
--- a/crates/gosub_shared/src/byte_stream.rs
+++ b/crates/gosub_shared/src/byte_stream.rs
@@ -599,7 +599,7 @@ impl LocationHandler {
             self.cur_location.offset -= 1;
         } else if self.cur_location.line > 1 {
             self.cur_location.line -= 1;
-            self.cur_location.column = self.column_stack.pop().expect("column_stack is empty");
+            self.cur_location.column = self.column_stack.pop().unwrap_or(1);
             self.cur_location.offset -= 1;
         }
     }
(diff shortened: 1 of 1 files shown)


crates/gosub_shared/src/byte_stream.rs
```diff
@@ -598,12 +597,9 @@ impl LocationHandler {
...
             self.cur_location.column -= 1;
             self.cur_location.offset -= 1;
-            return;
-        }
-
-        if self.cur_location.line > 1 {
+        } else if self.cur_location.line > 1 {
             self.cur_location.line -= 1;
-            self.cur_location.column = self.line_endings[&self.cur_location.line];
+            self.cur_location.column = self.column_stack.pop().expect("column_stack is empty");
```
Sharktheone, line 602: Does this work? `pop` removes the last item, shouldn't this be `last`?
Maybe you can replace the expect with a `.unwrap_or(1)`

Sharktheone (changes requested):

crates/gosub_shared/src/byte_stream.rs
yt2b: It does work. Column size is pushed when moving to next line.
So last column size must be removed when moving to previous line.
`last` returns last element But doesn't remove element.

To avoid panic, I'll replace `expect` with `.unwrap_or(1)`.

jaytaph (approved):

crates/gosub_shared/src/byte_stream.rs
Sharktheone: Ah, I thought it needs to be removed to actually have a stack of offsets to lines.

Sharktheone (approved):

jaytaph:
Thanks @yt2b  for the PR!

PR #677, opened 2024-11-11
support more encoding formats

Sharktheone:


jaytaph:
@Sharktheone is this still relevant?

Sharktheone:
It kind of is, but this probably has a lot of conflicts with the new gosub net impl

PR #682, opened 2024-11-17
fix specificity in cascaded value calculation & fix specified value calculation

Sharktheone:
There are still some issues with shorthands because we currently don't store their specificity and so on, but that's something for later

### DIFF
--- a/crates/gosub_css3/src/matcher/styling.rs
+++ b/crates/gosub_css3/src/matcher/styling.rs
@@ -446,34 +446,19 @@ impl CssProperty {
     }
 
     fn find_cascaded_value(&self) -> Option<CssValue> {
-        let mut declared = self.declared.clone();
-
-        declared.sort();
-        declared.sort_by(|a, b| {
-            if a.priority() == b.priority() {
-                return Ordering::Equal;
-            }
-
-            a.specificity.cmp(&b.specificity)
-        });
-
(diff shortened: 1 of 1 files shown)


crates/gosub_css3/src/matcher/styling.rs
```diff
@@ -446,36 +446,25 @@ impl CssProperty {
...
-            a.specificity.cmp(&b.specificity)
-        });
-
-        declared.last().map(|d| d.value.clone())
+        self.declared
+            .iter()
+            .max_by(|a, b| match a.cmp(b) {
+                Ordering::Greater => Ordering::Greater,
+                Ordering::Less => Ordering::Less,
+                Ordering::Equal => a.specificity.cmp(&b.specificity),
```
blaumeise20, line 454: Hellohello, another random comment from me after months of silence... You can use `Ordering::then_with` instead of manually matching against `Equal`. It would look like `a.cmp(b).then_with(|| a.specificity.cmp(&b.specificity))`.
I'm following this project a bit through email notifications, it is amazing what you already managed to do!

jaytaph (approved):

crates/gosub_css3/src/matcher/styling.rs
Sharktheone: Interesting, will change that. Thanks!

PR #688, opened 2024-11-21
begin migration to engine v2

Sharktheone:
Implement the [engine v2 POC](https://github.com/gosub-io/engine_v2) in the actual engine!


crates/gosub_shared/Cargo.toml
```diff
@@ -17,6 +17,10 @@ chardetng = "0.1.17"
 encoding_rs = "0.8.35"
 derive_more = {version = "1.0.0", features = ["display"]}
 log = "0.4.22"
+raw-window-handle = "0.6.2"
```
jaytaph, line 20: are these needed?

I'd rather not have the shared crate depend on a non-shared crate (typeface in this case) just to make sure we don't get any conflicts with cyclic dependencies

Sharktheone: Sadly, yes. At least the raw-window-handle one. The gosub_typeface is probably a candidate that could be moved into gosub_shared. This is because all trait definitions now need to live in gosub_shared

jaytaph (approved):

PR #703, opened 2024-12-06
Initial setup of cairo rendering backend

jaytaph:
Initial setup of the backend renderer that uses the cairo toolkit instead of vello.

- Some generic toolkits are used (skrifa, kurbo, peniko), which are also present in vello. These crates are installed separately.
- Rendering is not done immediately onto a cairo context, but instead stored as commands in a Scene (just like vello). Later, the render() function will apply the scene commands onto the cairo context.


crates/gosub_cairo/src/debug/text.rs
```diff
@@ -0,0 +1,92 @@
...
+use crate::elements::color::GsColor;
+use crate::elements::transform::GsTransform;
+use crate::Scene;
+use gosub_shared::render_backend::{Brush as _, Color as _, Transform as _};
+use gosub_shared::types::Point;
+use gosub_typeface::ROBOTO_FONT;
+use peniko::{Blob, Font};
+use std::sync::{Arc, LazyLock};
+
+static FONT: LazyLock<Font> = LazyLock::new(|| Font::new(Blob::new(Arc::new(ROBOTO_FONT)), 0));
```
Sharktheone, line 11: I guess this can be removed, since we can't use this with cairo

crates/gosub_cairo/src/elements/border.rs
```diff
@@ -0,0 +1,555 @@
...
+            width: border_width,
+            join: Join::Bevel,
+            miter_limit: 0.0,
+            start_cap: cap,
+            end_cap: cap,
+            dash_pattern,
+            dash_offset: 0.0,
+        };
+
+        todo!("scene.stroke()");
```
Sharktheone, line 345: I guess, we should not panic here, instead we should just print a message

crates/gosub_cairo/src/elements/text.rs
```diff
@@ -0,0 +1,233 @@
...
+        let font_face = unsafe {
+            LIB_FONT_FACE.with(|ctx_mutex| {
+                let mut ctx = ctx_mutex.lock().unwrap();
+
+                let ft_face = ctx.find_face_family("sans-serif");
+                let ft_face_ptr = ft_face.raw_mut() as *mut _ as *mut std::ffi::c_void;
+                let ff = cairo::ffi::cairo_ft_font_face_create_for_ft_face(ft_face_ptr, 0);
+                cairo::FontFace::from_raw_full(ff)
+            })
+        };
```
Sharktheone, line 184: I'm not 100% sure if this code is sound because you're locking the Mutex, creating a pointer in the FontFace, then dropping the Mutex lock, while still having the pointer to something in the Mutex.

I mean sure, this should always work, since it is a thread_local, so no one will move in the mutex for the next two lines.

I assume it should be fine, as long as cairo does not need the pointer to the font after the set_font_face call

crates/gosub_cairo/src/elements/text.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        &mut self.default_face
+    }
+}
+
+thread_local! {
+    /// We use a thread-local lazy static to ensure the font context is initialized once per thread
+    /// and is dropped when the thread exits. We need this because the FreeType library cannot be dropped
+    /// while any faces are still alive, so all is managed within this struct.
+    static LIB_FONT_FACE: Lazy<Mutex<GosubFontContext>> = Lazy::new(|| {
```
Sharktheone, line 95: Why is this a `Mutex`? A `RefCell` should work fine here, since it is a `thread_local`

crates/gosub_cairo/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    /// @todo: find out when this is called
+    fn suspend_window(
+        &mut self,
+        _handle: impl WindowHandle,
+        _data: &mut Self::ActiveWindowData<'_>,
+        _window_data: &mut Self::WindowData<'_>,
+    ) -> Result<()> {
+        Ok(())
+    }
```
Sharktheone, line 95: These two methods are called when the window is switching from shown to hidden or the other way around. I assume this is to save some memory by destroying the surfaces of the rendered window, which can be quite large in terms of memory.

I assume in cairo we can just ignore this, as we won't get this event by gtk and gtk should handle all of that internally, however this might not be always the case, for example when we use winit and vello

crates/gosub_cairo/src/scene.rs
```diff
@@ -0,0 +1,278 @@
...
+};
+use std::fmt::{Debug, Formatter};
+
+/// A scene command that can be executed onto a cairo context.
+#[derive(Clone)]
+pub enum SceneCommand {
+    // Draw a rectangle, including rounded corners and border
+    Rectangle(Box<RenderRect<CairoBackend>>),
+    // Draw a text
+    Text(Box<RenderText<CairoBackend>>),
```
Sharktheone, line 17: Is it really worth it to put those inside a Box? I mean, we will mainly use these, so it should be fine if the size of `SceneCommand` is as big as the `RenderRect` or `RenderText` itself are

crates/gosub_cairo/src/scene.rs
```diff
@@ -0,0 +1,278 @@
...
+                .finish(),
+        }
+    }
+}
+
+/// A scene holds all the commands that must be executed onto the given cairo context.
+#[derive(Debug)]
+pub struct Scene {
+    /// Scene identifier. We can remove later after development/debugging
+    pub(crate) id: uuid::Uuid,
```
Sharktheone, line 68: smells like a feature flag for me

crates/gosub_cairo/src/scene.rs
```diff
@@ -0,0 +1,278 @@
...
+        // if let Some(border) = &rect.border {
+        //     let opts = BorderRenderOptions {
+        //         border,
+        //         rect: &rect.rect,
+        //         transform: rect.transform.as_ref(),
+        //         radius: rect.radius.as_ref(),
+        //     };
+        //
+        //     Border::draw(&mut self.context, opts);
+        // }
```
Sharktheone, line 177: dead code

crates/gosub_cairo/src/render/window.rs
```diff
@@ -0,0 +1,10 @@
+use crate::Scene;
+
+pub struct WindowData {
+    pub _cr: Option<cairo::Context>,
```
Sharktheone, line 4: I think this can be removed, since it is replaced by the scene?

examples/gtk-renderer/README.md
```diff
@@ -0,0 +1,66 @@
+# GTK Renderer Example
+
+This example demonstrates how to use the GTK renderer to render a website in a GTK window.
+
+## Usage
+
+```bash
+cargo run --release --package gtk-renderer -- <URL>
```
Sharktheone, line 8: `cargo run -r -p gtk-renderer -- <URL>`?

Sharktheone (dismissed):

crates/gosub_cairo/src/scene.rs
jaytaph: yes.. otherwise the enum differences are too big (800 bytes, vs 400 bytes, vs a lot less)... cargo complains about this

crates/gosub_cairo/src/scene.rs
jaytaph: no, we can remove this

crates/gosub_cairo/src/debug/text.rs
jaytaph: FONT is being used in the render_text(). For now it doesn't do much though

crates/gosub_cairo/src/lib.rs
jaytaph: i don't think this information should leak into the rendering backend to begin with.. for now, its fine as we don't use this in the cairo backend

crates/gosub_cairo/src/render/window.rs
jaytaph: The scene does not know anything about the cairo context.

Due to the current setup of the renderer, we need to work with ActiveWindowData and WindowData. We do not have a generic way to send over the cr in the render function.  

Normally, you would only need to pass the cairo context (or whatever it needs to do the actual painting) at the render() step, but vello works differently.. Since there is also a `activate_window` function that creates an ActiveWindowData from WindowData, we probably need a copy of cr in the window data before we can copy it to activewindowdata..

we really should redesign the renderer so it makes more sense again for both systems.

crates/gosub_cairo/src/scene.rs
Sharktheone: Yes, but I assume 95% of the scene commands are Rectangle and Text Nodes, if we have another one, so SimpleText or Group, we are wasting a bit of memory, but it probably is worth it

crates/gosub_cairo/src/lib.rs
Sharktheone: i guess we could change it, so the tab just renders to a scene and the rest is handled by the implementation, I mean in the future the tab doesn't get the full surface anyway

jaytaph: yeah, there will be some more iterations on the rendering stuff later on.. at least we now have two different implementations so we can generalize the interface a bit better..

crates/gosub_cairo/src/scene.rs
```diff
@@ -0,0 +1,232 @@
...
+            (x + tl_rx) as f64,
+            (y + tl_ry) as f64,
+            tl_rx.min(tl_ry) as f64,
+            std::f64::consts::PI,
+            3.0 * std::f64::consts::FRAC_PI_2,
+        );
+    }
+
+    cr.close_path();
+}
```
Sharktheone, line 232: It seems like this isn't used? There is another impl of that in `elements/border.rs` I assume this is for later, just pointing it out in case one of them should be removed.

Sharktheone (approved):
Okay, looks fine now. Just one thing where I'm not 100% sure of. I assume you do the border stuff later, but found that thing.

PR #721, opened 2024-12-20
moved vello to separate example

jaytaph:
- examples are now working properly with `cargo run --example vello-renderer"
- updated some readme

One thing missing: wasm seems not to function anymore, as it seems to use things from the vello renderer..


docs/crates.md
```diff
@@ -64,4 +60,65 @@ Implementation of a RenderBackend for the `vello` crate
...
+    13 -> 5 [ ]
+    13 -> 4 [ ]
+    13 -> 3 [ ]
+    13 -> 6 [ ]
+    13 -> 7 [ ]
+    13 -> 8 [ ]
+    13 -> 1 [ ]
+    13 -> 9 [ ]
+    13 -> 12 [ ]
+}
```
Sharktheone, line 121: It seems like this diagram is broken

Sharktheone (changes requested):

docs/crates.md
jaytaph: it's graphviz instead of mermaid syntax.. this has been fixed now

Sharktheone (approved):

PR #729, opened 2024-12-23
Added more to the prelude

jaytaph:


Sharktheone (commented):
I'm not sure about those. I'm currently in the proccess of removing the `WindowedEventLoop` which means that we probably also don't need to worry about the `ImageBuffer`

The `WindowData` and `ActiveWindowData` will probably fall out of the public API and the rendering proccess will just output one `Scene` in the end

Sharktheone (approved):

jaytaph:
We will be moving more things out of the prelude once the final gosub_engine api materializes. At this moment, we have none so we will be using mostly bits and pieces for now..

PR #730, opened 2024-12-24
Added functionality to render tree from html-source instead of url

jaytaph:
The gtk browser alraedy fetches the URL before the drawer. So instead of letting the drawer fetching the url, we provide the html source

### DIFF
--- a/crates/gosub_interface/src/draw.rs
+++ b/crates/gosub_interface/src/draw.rs
@@ -36,7 +36,7 @@ pub trait TreeDrawer<C: HasDrawComponents> {
         layouter: C::Layouter,
         // Debug flag
         debug: bool,
-    ) -> impl Future<Output = gosub_shared::types::Result<Self>>
+    ) -> gosub_shared::types::Result<Self>
     where
         Self: Sized;
 
--- a/crates/gosub_interface/src/draw.rs
+++ b/crates/gosub_interface/src/draw.rs
@@ -27,6 +27,19 @@ pub trait TreeDrawer<C: HasDrawComponents> {
     where
         Self: Sized;
 
+    fn from_source(
+        // The initial url that the source was loaded from
+        url: Url,
+        // Actual loaded source HTML
+        source_html: &str,
+        // Layouter that renders the tree
+        layouter: C::Layouter,
+        // Debug flag
+        debug: bool,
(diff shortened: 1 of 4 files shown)


crates/gosub_interface/src/draw.rs
```diff
@@ -27,6 +27,19 @@ pub trait TreeDrawer<C: HasDrawComponents> {
...
+    fn from_source(
+        // The initial url that the source was loaded from
+        url: Url,
+        // Actual loaded source HTML
+        source_html: &str,
+        // Layouter that renders the tree
+        layouter: C::Layouter,
+        // Debug flag
+        debug: bool,
+    ) -> impl Future<Output = gosub_shared::types::Result<Self>>
```
Sharktheone, line 39: I don't think this needs to be a future, since the only thing that is async in there is fetching the HTML, which was already done

crates/gosub_renderer/src/render_tree.rs
```diff
@@ -10,40 +10,31 @@ use gosub_shared::byte_stream::{ByteStream, Encoding};
...
     let fetcher = Fetcher::new(url.clone());
 
-    let rt = load_html_rendertree_fetcher::<C>(url, &fetcher).await?;
+    let rt = match source {
+        Some(source) => load_html_rendertree_source::<C>(url, source).await?,
+        None => load_html_rendertree_fetcher::<C>(url, &fetcher).await?,
+    };
 
     Ok((rt, fetcher))
 }
```
Sharktheone, line 28: I think this should be removed, instead callers should create their fetcher by temselves and give it `load_html_rendertree_fetcher`

Sharktheone (changes requested):

crates/gosub_interface/src/draw.rs
jaytaph: actually, no.. the parsing is also async.. so this needs to be async as well.

crates/gosub_renderer/src/render_tree.rs
jaytaph: Added a new issue for this:

https://github.com/gosub-io/gosub-engine/issues/732

Sharktheone (approved):

PR #733, opened 2024-12-24
begin event loop and `EngineInstance`

Sharktheone:


crates/gosub_instance/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+use gosub_shared::geo::SizeU32;
+use gosub_shared::types::Result;
+use std::sync::mpsc::Sender as SyncSender;
+use tokio::runtime::{Builder, Runtime};
+use tokio::sync::mpsc::{Receiver, Sender};
+use tokio::task;
+use tokio::task::{spawn_blocking, LocalSet};
+use url::Url;
+
+pub struct EngineInstance<C: ModuleConfiguration> {
```
jaytaph, line 16: I don't know exactly what an engine instance is... but from context I don't understand why it needs a title or url?

crates/gosub_interface/src/config/chrome.rs
```diff
@@ -0,0 +1,6 @@
+use crate::chrome::ChromeHandle;
+use crate::config::HasRenderBackend;
+
+pub trait HasChrome: HasRenderBackend + Sized {
```
jaytaph, line 4: Please describe what chrome is and what it does...

crates/gosub_instance/src/lib.rs
Sharktheone: It is basically a tab. It will manage the event loop, the web event loop and everything else.

It needs the title and url, so JS can ask for it

crates/gosub_interface/src/config/chrome.rs
Sharktheone: Yeah, I still need many comments. I am also not completely happy about the `HasChrome` thingy.

jaytaph (approved):

PR #740, opened 2024-12-27
Updates cargo depenencies for packaging crates

jaytaph:
All crates can now be packaged separately, including the main gosub_engine crate on registry.gosub.io


Sharktheone:
I'm not sure if I like this, as you are now required to set up the registry and it probably also must work. If we decide to publish our packages on crates.io we can no longer just try out an older version

jaytaph:
@Sharktheone nothing changes. It will still use the path definitions inside the Cargo.toml files.. We just have a second source (version + registry) in order to enable usages as crates itself..

so:

 1. you can still use gosub_engine the regular way, by having the crates in ./crates and modify them
 2. when writing external applications (which is the whole reason of the engine), you can use the gosub_engine (or any other crate) as dependency from the repository.

Sharktheone:
Yes, but when you have the `registry = "gosub"` thing, then cargo wants to know about the registry and potentially also tries to request it (not sure), so if the registry is offline at some point in the future you can't test the versions with the `registry = "gosub"`

Sharktheone:
https://chat.developer.gosub.io/#narrow/stream/6-gosub/topic/crate.20registry/near/13872

.github/workflows/ci.yaml
```diff
@@ -40,6 +40,12 @@ jobs:
           key: ${{ matrix.os }}-cargo-target-${{ matrix.rust_version }}-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: |
             ${{ matrix.os }}-cargo-target-${{ matrix.rust_version }}-
+      - name: Add Gosub crate registry
+        run: |
+          cat <<EOF > ~/.cargo/config.toml
+          [registries]
+          gosub = { index = "sparse+https://registry.gosub.io/" }
+          EOF
```
Sharktheone, line 48: this should be just in `gosub-engine/.cargo/config.toml` not in `~/.cargo/config.toml` so when you clone the engine you don't need to modify something in your home folder

crates/gosub_render_utils/Cargo.toml
Sharktheone, line 1: ~~What happend here? Is this no longer needed? If so you probably should also delete the code in the render_utils folder. But it seems like this is still needed?~~

Ah, it was moved to `gosub_rendering`

Sharktheone (changes requested):

Sharktheone (approved):

PR #741, opened 2024-12-28
Begin WebEventLoop

Sharktheone:


crates/gosub_web_platform/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
```
jaytaph, line 1: what does this to / what is it for?

Sharktheone: This will implement the "Web Platform", so things that a js or lua script can interact with. This might be timers, and Web APIs

Sharktheone:
This diagram describes the architecture that already is in the engine or gets introduced with this PR.

webeventloop_architecture

jaytaph (approved):

Issue #761, opened 2025-01-12
perf: use cow-utils instead


heygsc:
We can improve performance by using `cow_utils`, and automatically check in clippy.

We have done similar work before, and if necessary and possible, I would like to provide pr for boa.

Related issue: 
https://github.com/gosub-io/gosub-engine/issues/752
https://github.com/oxc-project/oxc/issues/5586

jaytaph:
Hi!.  This is something we really want as well.. The html parser for instance is written at a time there wasn't too much experience with Rust, so there is a lot of cloning and copying of data which probably shouldn't happen.  @Sharktheone your thoughts?

Sharktheone:
I think this is something we can do. Maybe we can save this as a more general string optimization. For example, in some places, we should replace the std `String` with some `SmallString` variant. Or more specifically, in the HTML parser we are currently not sharing the namespace. And I think we also store the source file name, that should also be shared.

PR #762, opened 2025-01-12
use `cow_to_ascii_lowercase` instead

heygsc:
part of #761 

https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_methods

If this PR can be merged, I will continue to use more cow methods in clippy.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,6 +6,7 @@
 Ahmed Ibrahim <me@ahmedibrahim.dev>
 DynamiteC <shethshails@gmail.com>
 Emre AYDIN <aeaydin1@gmail.com>
+heygsc <1596920983@qq.com>
 Joshua Thijssen <jthijssen@noxlogic.nl>
 Nicholas Nguyen <nicholas.nguyen@chargeitspot.com>
 Niklas Scheerhoorn <sinner1991@gmail.com>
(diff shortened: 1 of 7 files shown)


jaytaph:
> By the way, running clippy in make fix-format may cause some code to automatically change, which I did not submit because the clippy in make check-format allows it to pass.

This is good. The PR should only reflect your changes, not unrelated format changes.  I'm not sure why this is happening though.. The CI should automatically reject clippy issues that are not resolved, so everybody should have the same formatting. I've noticed this happens on occasion with other PR's, so it might be a good idea to look into this issue as well (later)..

jaytaph (approved):

jaytaph:
Looks good.. Thanks @heygsc  for the PR!

jaytaph:
Ah, we found a small issue concerning your git commits. They need to be signed commits. For more info about signed commits see here: https://docs.github.com/authentication/managing-commit-signature-verification/about-commit-signature-verification

You should be able to either rebase your PR with a signed commit, or reopen a new PR with signed commits so we can merge.

heygsc:
Okay, I did neglect to carefully read the contribution document. I only realized it before modifying the AUTHORS, so the second commit has a gpg sign.

I will open a new PR later.

PR #771, opened 2025-01-13
Fixed clippy error on stable

jaytaph:


Sharktheone (approved):
I think we even have more of them. For instance, currently, the `examples` are not checked by clippy.

PR #772, opened 2025-01-13
use `cow_to_ascii_lowercase` instead

heygsc:
same pr: #762

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,6 +6,7 @@
 Ahmed Ibrahim <me@ahmedibrahim.dev>
 DynamiteC <shethshails@gmail.com>
 Emre AYDIN <aeaydin1@gmail.com>
+heygsc <1596920983@qq.com>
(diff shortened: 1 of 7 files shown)


jaytaph:
@heygsc believe it or not,.. but it seems i've just committed a change in a code-line you touched as well :/

I fixed a clippy issue here: https://github.com/gosub-io/gosub-engine/pull/771/files#diff-3c1cd73766aa14d7d38d3068dcb66112bcce6c37de5478884f2dd4123bc106e4L14

It might be possible that this fix is obsolete because of your PR.. in that case, you can resolve the branch conflict with your version.

heygsc:
@jaytaph   I also noticed that it happened to be very close, and at first I hesitated whether to ask you which one to merge first.🤪

It has already been modified.

jaytaph (approved):

jaytaph:
@heygsc  looks good!  Thanks for the PR!

PR #773, opened 2025-01-13
use `cow-utils` instead

heygsc:
close: #761


jaytaph (approved):

jaytaph:
@heygsc Would it be possible somehow to see any benchmarking of this?  We have a bencher.dev setup, but i'm not sure if that works correctly yet, and I would love to see if and how performance would increase this way

Sharktheone (approved):
I hope this doesn't interfere with my latest PR. But I guess the conflicts would be relatively minor

heygsc:
> @heygsc Would it be possible somehow to see any benchmarking of this? We have a bencher.dev setup, but i'm not sure if that works correctly yet, and I would love to see if and how performance would increase this way

I am also very curious about how much performance it can improve, but unfortunately I am not very familiar with benchmark related things.🥲

jaytaph:
@heygsc Thanks for the PR!

PR #838, opened 2025-04-14
Fixed css syntax where literals where parsed as tokens

jaytaph:
- Fixed css parsing as there was a minor issue with multipliers
- Fixed the definition generator to deal with literals like , ; ()
- Fixed the output sorting of the definition generator so elements are
  always correctly ordered, which makes it easier to use in version
  control.

### DIFF
--- a/crates/gosub_css3/resources/definitions/definitions.json
+++ b/crates/gosub_css3/resources/definitions/definitions.json
@@ -1,10 +1,327 @@
 {
   "properties": [
     {
-      "name": "border-clip-top",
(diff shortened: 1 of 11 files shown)


crates/gosub_css3/resources/definitions/definitions.json
Sharktheone, line 1: I think this file should not be edited as it is generated from the spec, so we should either file a spec PR or somehow change it in the parsing

Sharktheone (changes requested):

Sharktheone (changes requested):

Sharktheone (approved):

PR #893, opened 2026-03-31
Big component merge

jaytaph:
This is a big component merge assisted by claude to merge the new engine api system AND the renderer pipeline into the current engine.

Currently we have merged it into a state that everything could be merged, but we still need to do more work here:

  - Go over the code (it's a massive merge!), and spot any issues.
  - Check new and old examples and tooling to see if they still work in the new setup. Make sure old tooling uses the new engine api.
  - The new engine api has documentation, which we should structurize better and place in a seperate /doc directory.
  - We should NOT change anything that is not needed for the merge. There is still a lot of things that can and must be improved, but for now this should only focus on changing things for the merge.


crates/gosub_engine_core/examples/egui_vello/main.rs
```diff
@@ -0,0 +1,273 @@
...
+
+// ---------------------------------------------------------------------------
+// WgpuContextProvider implementation backed by eframe's render state
+// ---------------------------------------------------------------------------
+
+struct EframeWgpuContext {
+    // eframe hands us non-Arc references to device/queue; we clone (unsafe-free)
+    // by holding the render state alive for the app lifetime. Here we store
+    // raw pointers and convert them back inside the impl — this is safe as long
+    // as eframe's render state outlives this struct (it does: both live in App).
```
Sharktheone, line 35: Are we sure that there isn't a better alternative?

crates/gosub_engine_core/examples/egui_vello/main.rs
```diff
@@ -0,0 +1,273 @@
...
+            sample_count: 1,
+            dimension: wgpu::TextureDimension::D2,
+            format,
+            usage: wgpu::TextureUsages::RENDER_ATTACHMENT | wgpu::TextureUsages::TEXTURE_BINDING,
+            view_formats: &[],
+        });
+        let view = texture.create_view(&wgpu::TextureViewDescriptor::default());
+
+        let id = {
+            let mut next = self.next_id.lock().unwrap();
```
Sharktheone, line 89: Not sure what we should do about unwrap in examples. I guess they are fine?

crates/gosub_engine_core/examples/egui_vello/main.rs
```diff
@@ -0,0 +1,273 @@
...
+        guard.get(&id).map(|(t, v)| {
+            // SAFETY: we cast the texture/view references into owned values via
+            // an unsafe clone.  This is acceptable here because both the
+            // returned values and the stored entries are kept alive for the
+            // same duration (the lifetime of EframeWgpuContext).
+            //
+            // A production integration would use Arc<Mutex<...>> or wgpu's
+            // TextureId registry instead.
+            let t2 = unsafe { std::ptr::read(t as *const wgpu::Texture) };
+            let v2 = unsafe { std::ptr::read(v as *const wgpu::TextureView) };
```
Sharktheone, line 113: This is just wrong. wgpu::Texture does implement Clone.

crates/gosub_engine_core/examples/egui_vello/main.rs
```diff
@@ -0,0 +1,273 @@
...
+
+impl eframe::App for GosubApp {
+    fn update(&mut self, ctx: &egui::Context, frame: &mut eframe::Frame) {
+        // Initialise the Vello backend on first frame (when wgpu is available)
+        if self.backend.is_none() {
+            if let Some(wgpu_state) = frame.wgpu_render_state() {
+                // SAFETY: wgpu_state is kept alive by eframe for the app lifetime.
+                let wgpu_ctx = unsafe {
+                    EframeWgpuContext::new(&wgpu_state.device, &wgpu_state.queue)
+                };
```
Sharktheone: Just clone the device and queue here

Sharktheone (changes requested):

crates/gosub_engine_core/examples/egui_vello/main.rs
jaytaph: probably.. but at the moment it's what we've got..

crates/gosub_engine_core/examples/egui_vello/main.rs
jaytaph: I would rather we keep unwrap()s out of non-test modules..  I think we can enforce this

Sharktheone: there is a clippy rule to do that

jaytaph:
I'm closing this one in favor of the new merge attempt: #895

PR #923, opened 2026-04-28
Feat/net update

jaytaph:
Updated the net code so we got rid of the "old" code, and use the new code, also in the current engine. 

We removed ureq and using reqwest for now, but we clean that up in the next PR.


crates/gosub_net/src/http/fetcher.rs
```diff
@@ -1,36 +1,26 @@
...
-    fn new() -> Self;
-    fn get(&self, url: &str) -> impl Future<Output = Result<Response>>;
-    fn get_req(&self, req: &Request) -> impl Future<Output = Result<Response>>;
-}
-
 #[derive(Debug)]
 pub struct Fetcher {
     base_url: Url,
-    client: RequestImpl,
+    #[cfg(not(target_arch = "wasm32"))]
```
Sharktheone: I don't think this should be feature gated here. reqwest also works on wasm. We might only need to enable some feature

crates/gosub_net/src/http/fetcher.rs
```diff
@@ -39,41 +29,80 @@ impl Fetcher {
...
+        let body = resp.bytes().await?.to_vec();
+        Ok(Response {
+            status,
+            body,
+            ..Response::default()
+        })
+    }
+
+    #[cfg(target_arch = "wasm32")]
+    pub async fn get_url(&self, url: &Url) -> Result<Response> {
```
Sharktheone: I would just let reqwest handle this part, no anointing handling with js_sys and web_sys

Sharktheone (changes requested):

crates/gosub_net/src/http/fetcher.rs
jaytaph: should we remove the !wasm target in all instances here?

crates/gosub_net/src/http/fetcher.rs
```diff
@@ -41,39 +29,76 @@ impl Fetcher {
...
+        let status = resp.status().as_u16();
+        let body = resp.bytes().await?.to_vec();
+        Ok(Response {
+            status,
+            body,
+            ..Response::default()
+        })
+    }
+
+    pub async fn get_url(&self, url: &Url) -> Result<Response> {
```
Sharktheone: This is now duplicated, you need to remove this function, not just the cfg attribute

jaytaph: yes, i see now..   i'm trying to see if i compile and test wasm separately

jaytaph: wasm compiles and uses plain reqwest now

Sharktheone (approved):

PR #927, opened 2026-04-28
Documentation update

jaytaph:


.github/workflows/ci.yaml
```diff
@@ -23,6 +23,10 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust_version }}
+      - name: Free disk space
+        run: |
+          sudo rm -rf /usr/local/lib/android /usr/share/dotnet /opt/ghc /usr/local/.ghcup
+          df -h
```
Sharktheone: What? Why do we need that?

jaytaph: It is not 100% needed but it gets rid of some unused packages. We were hitting a diskspace issue (through our cache) but this is cleaned up and should not fill up too much. It saves us about 1.5gb of space. But id reckon we should monitor our caches as well so we dont reach limits, for instance if github decides to decrease worker diskspace

Sharktheone: Huh, are we really caching the full system? I thought we'd only cache our own project dirs. Maybe that is a better solution

jaytaph: We cache only our own dirs.

jaytaph: @Sharktheone i've changed the current caching system to a different system.. this would keep the caches to a reasonable size

Sharktheone (approved):

Sharktheone (approved):

PR #929, opened 2026-04-30
Storage agnostic interface

jaytaph:
Rewrite document system to be storage-agnostic via NodeId handles

Follows the same handle pattern used throughout the engine (TabId, ZoneId, etc.): instead of handing out references to concrete Node structs, all DOM access now goes through opaque NodeId handles. The Document<C> trait is the only interface callers may use.

Before:
    let node: &NodeImpl = doc.node_by_id(id)?;
    let name = node.element_data().name.as_str();

After:
    let name = doc.tag_name(id);          // NodeId in, &str out


crates/gosub_css3/src/matcher/styling.rs
```diff
@@ -46,124 +43,78 @@ fn consume<'a, T>(this: &mut &'a [T]) -> Option<&'a T> {
...
+            doc.node_type(current_id) == NodeType::ElementNode && doc.tag_name(current_id).is_some_and(|t| t == name)
         }
         CssSelectorPart::Class(name) => {
-            if !current_node.is_element_node() {
-                return false;
-            }
-            current_node.get_element_data().unwrap().classlist().contains(name)
+            doc.node_type(current_id) == NodeType::ElementNode
+                && doc
+                    .attribute(current_id, "class")
```
Sharktheone: Here you use the class attribute, but there still is the classlist()

crates/gosub_html5/src/document/document_impl.rs
```diff
@@ -551,1077 +671,38 @@ impl<C: HasDocument> Iterator for TreeIterator<'_, C> {
...
+) {
+    let root = doc.get_root();
+    internal_visit(doc, root, visitor);
+}
 
-        assert_eq!(expected_order, traversed_nodes);
+fn internal_visit<C: HasDocument<Document = DocumentImpl<C>>>(
+    doc: &DocumentImpl<C>,
+    node: &NodeImpl,
+    visitor: &mut Box<dyn Visitor>,
```
Sharktheone: I think this should be `visitor: impl Visitor` or at least `visitor: &mut dyn Visitor`

crates/gosub_interface/src/layout.rs
```diff
@@ -167,7 +168,7 @@ pub trait Layout: Default + Debug {
 }
 
 pub trait LayoutNode<C: HasLayouter>: HasTextLayout<C> {
-    fn get_property(&self, name: &str) -> Option<&C::CssProperty>;
+    fn get_property(&self, name: &str) -> Option<&<C::CssSystem as CssSystem>::Property>;
```
Sharktheone, line 171: For example here, since you removed the Property type in the CssSystem, you now have to do `<C::CssSystem as CssSystem>::Property>` idk if I want that over `C::CssProperty`

crates/gosub_renderer/Cargo.toml
```diff
@@ -8,6 +8,7 @@ description = "Gosub Renderer"
 
 [dependencies]
 gosub_rendering = { version = "0.1.1", registry = "gosub", path = "../gosub_rendering" }
+gosub_html5 = { version = "0.1.1", registry = "gosub", path = "../gosub_html5" }
```
Sharktheone: Why do we need gosub_html5 here? The whole point of the modular config is that  we don't have to do that and can swap out everything

Sharktheone (changes requested):
So, not sure if we actually want to drop the shortcuts in the `Has*` traits, it has some syntax changes which makes things more unclean

crates/gosub_css3/src/matcher/styling.rs
jaytaph: I think we should add a more generic has_class() to the document. This way we don't need to check element types. Internally, it can use the classlist

crates/gosub_html5/src/document/document_impl.rs
jaytaph: :+1:

crates/gosub_renderer/Cargo.toml
jaytaph: yes, it shouldn't

jaytaph: fixed by using document trait instead of the implementation DocumentBuilderImpl

crates/gosub_interface/src/layout.rs
jaytaph: I'm happy to bring back the "HasProperty" and others if this makes it more readable

Sharktheone: Sure, do we want to do that in this PR or a different one?

jaytaph: I think we can try another PR for that.. we might want to do an inventory on which ones we need?

Sharktheone: I think that makes sense. Maybe we should open an issue for that

Sharktheone (approved):

PR #936, opened 2026-05-02
Audit fix

jaytaph:
## Chores
- Updated `indicatif` development dependency to version 0.18.
- Strengthened continuous integration audit job permissions for enhanced security checks.
- Implemented explicit security advisory management to address known vulnerabilities and maintain code integrity.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -86,6 +86,8 @@ jobs:
 
   audit:
     runs-on: blacksmith-4vcpu-ubuntu-2404
+    permissions:
+      checks: write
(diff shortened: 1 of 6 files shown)


Sharktheone (changes requested):
Let's add a note to where we are pulling in V8 to say that we need to update it and then remove the advisory. Same for paste, we should state that we need to remove that.

jaytaph:
I'll add the note to cargo.toml, but i don't think we need to remove the advisory ignore, because it will trigger an audit failure..

Sharktheone (approved):

PR #950, opened 2026-05-04
Bytestream refactoring

jaytaph:
The following has been refactored on the bytestream system:

An incremental rewrite of gosub_shared::ByteStream where some of the functionality is replaced by pre-existing libs (for instance, for handling multi-byte sequences)

- Integrate LocationHandler into ByteStream. read_and_next auto-advances line/col, prev auto-decrements. Add mark()/reset_to_mark(). 
- Remove all manual location_handler.inc/dec() calls from both tokenizers
- Drop Cell/RefCell — mutating Stream methods now take &mut self; 
- remove Character::StreamEmpty (callers can use exhausted() && !closed() to detect open-but-drained)
- rename Encoding::UNKNOWN to Unknown, ASCII to Latin1
- add ByteStream::from_str() constructor and updated around 30 calls

Tests have been expanded from 11 to 31, now covering mark/reset, location tracking, look-ahead, get-slice, UTF-16 LE/BE,
   surrogate pairs, BOM detection, Latin-1 high bytes, tell_bytes/seek_bytes with multi-byte UTF-8.

## Summary by CodeRabbit

* **New Features**
  * Improved charset handling with Latin‑1 mapping and broader encoding support for inputs.

* **Bug Fixes**
  * More accurate byte/character position and end‑of‑stream behavior across parsing and tokenization.

* **Refactor**
  * Unified stream/location model and simplified input initialization used throughout parsers and tools.

* **Tests**
  * Updated suites to cover encodings, position/mark-reset behavior, and revised stream semantics.


crates/gosub_shared/src/byte_stream.rs
```diff
@@ -115,374 +109,380 @@ impl Default for Config {
...
-    buffer_pos: RefCell<usize>,
-    /// True when the buffer is empty and not yet have a closed stream
+    /// Pre-decoded characters (filled by decode_buffer)
+    chars: Vec<Character>,
+    /// Byte offset of each decoded character in `buffer`
+    char_byte_offsets: Vec<usize>,
+    /// char_pos of the first character on each line (index 0 = line 1)
+    line_starts: Vec<usize>,
+    /// Current position in the decoded chars array
+    char_pos: usize,
```
Sharktheone, line 121: I do we need those in the byte stream? I'd guess that might add a lot of overhead

Sharktheone: Maybe we can separete this to a different byte stream which specifically can do all of this position stuff

jaytaph: yes, we need these for now, as we want to be able to detect locations (line, char) for mostly debug and logging purposes.

Sharktheone (approved):

PR #958, opened 2026-05-10
Compiling under windows

jaytaph:
## Summary by CodeRabbit

* **New Features**
  * Example app exposes loaded page HTML per pane and triggers immediate UI repaint when navigation finishes.

* **Improvements**
  * Safer, thread-friendly render state synchronization and more robust surface/texture handling with graceful failure paths.
  * Simplified example rendering flow, async tab creation, and event-driven redraws for improved responsiveness.

* **Chores**
  * Platform-gated GTK/Cairo components to Unix targets and pinned revisions for rendering crates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gosub-io/gosub-engine/pull/958)

### DIFF
--- a/crates/gosub_engine/src/render/backends/vello.rs
+++ b/crates/gosub_engine/src/render/backends/vello.rs
@@ -7,8 +7,7 @@ use crate::render::backends::vello::text_renderer::{TextKey, TextRenderer};
 use crate::render::DisplayItem;
 use anyhow::{anyhow, Result};
 use std::any::Any;
-use std::cell::RefCell;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use vello::kurbo::Affine;
 use vello::peniko::{Color, Fill};
 use vello::wgpu;
@@ -33,10 +32,10 @@ pub struct VelloBackend<C: WgpuContextProvider + Send + Sync> {
     /// The wgpu context provider that we can use for device, queue, and texture management.
     context: Arc<C>,
     /// The Vello renderer instance.
-    renderer: RefCell<Renderer>,
-    text_renderer: RefCell<TextRenderer>,
-    font_manager: RefCell<FontManager>,
-    font_cache: RefCell<FontCache>,
+    renderer: Mutex<Renderer>,
+    text_renderer: Mutex<TextRenderer>,
+    font_manager: Mutex<FontManager>,
+    font_cache: Mutex<FontCache>,
 }
 
 impl<C: WgpuContextProvider + Send + Sync> VelloBackend<C> {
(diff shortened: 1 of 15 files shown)


crates/gosub_engine/src/render/backends/vello.rs
```diff
@@ -33,10 +32,10 @@ pub struct VelloBackend<C: WgpuContextProvider + Send + Sync> {
...
     context: Arc<C>,
     /// The Vello renderer instance.
-    renderer: RefCell<Renderer>,
-    text_renderer: RefCell<TextRenderer>,
-    font_manager: RefCell<FontManager>,
-    font_cache: RefCell<FontCache>,
+    renderer: Mutex<Renderer>,
+    text_renderer: Mutex<TextRenderer>,
+    font_manager: Mutex<FontManager>,
+    font_cache: Mutex<FontCache>,
```
Sharktheone, line 39: Not sure if we want this to be Mutex's. They have way more overhead than RefCells for multiple reasons

crates/gosub_engine/src/render/backends/vello.rs
```diff
@@ -169,9 +168,9 @@ impl<C: WgpuContextProvider + Send + Sync> RenderBackend for VelloBackend<C> {
 
         // Generate a scene which contains the gpu render commands
         let scene = {
-            let mut tr = self.text_renderer.borrow_mut();
-            let mut fm = self.font_manager.borrow_mut();
-            let mut fc = self.font_cache.borrow_mut();
+            let mut tr = self.text_renderer.lock().unwrap();
+            let mut fm = self.font_manager.lock().unwrap();
+            let mut fc = self.font_cache.lock().unwrap();
```
Sharktheone: Here we have unwraps now, I think in new code we should not have any unwraps

crates/gosub_engine/src/render/backends/vello.rs
Sharktheone: There might be a reason to use them though, but maybe we can then have that only in that case instead of always

Sharktheone (changes requested):

crates/gosub_engine/src/render/backends/vello.rs
jaytaph: I think we might want to use parkinglot instead?

jaytaph: The current code already uses parkinglot at some points. At other places, we use lock().unwrap(), which is basically what parkinglot does (it ignores poisoning).. 

my suggestion would be to for now to change to parkinglot, and have another issue open to move EVERYTHING that is currently lock().unwrap() to parkinglot. Functional wise it doesn't change anything, and we can later see if we want to make sure we can handle poisoning correctly (maybe we need to setup a correct engine result system)

crates/gosub_engine/src/render/backends/vello.rs
jaytaph: let's keep them here until we have a good reason to change.

Sharktheone (approved):
Approved from my side. Idk if the coderabbit things are outdated or not

PR #960, opened 2026-05-11
Refactored lock().unwrap() to parkinglot

jaytaph:
## Summary by CodeRabbit

* **Refactor**
  * Replaced assorted std synchronization primitives with a single locking implementation across the project, removing unwrap/poison-handling patterns and improving runtime robustness.
  * Updated examples and docs to reflect the new lock API.

* **Chores**
  * Added the new locking library as a dependency where required.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gosub-io/gosub-engine/pull/960)


jaytaph:
@coderabbitai help

jaytaph:
@coderabbitai resolve

Sharktheone (commented):
This in itself looks good, but I'm not fully sure if we want to actually use parking lot. We can have non-panicking `Mutex` locks with the `std::sync::Mutex` with

```rs
let x = Mutex::new(1);
            
let x = match x.lock() {
    Ok(guard) => guard,
    Err(poisoned) => poisoned.into_inner(),
};

dbg!(x); // always 1
```

jaytaph:
I think the best way is to deal with the poisoning through a Result<>. But i think this needs a bit more organization and the need for more proper Result usage. I would suggest that we would:

- merge this with parkinglot
- move all other mutexes (for now) to parking lot (there shouldn't be much left)
- setup a decent Result system for all crates so they work independently from eachother, but still be able to use them from other components.
- change the parkinglot mutexes so that poisoned entries will be returned with ?

Sharktheone:
I think this is okay, but I am not fully sure what other trade-offs we get by switching to parkinglot. It seems that parkinglot also uses futex, but also has a spin loop?

Sharktheone:
okay, well, the stdmutex also has a spin loop. I guess this is just fine

Sharktheone (approved):

PR #968, opened 2026-05-13
Stricter linting checks

jaytaph:
Added more strict linting checks to some of the crates (not all though, that needs a bit more refactoring).

Basically, any non-production call like unwrap(), expect(), unimplemented!, unsafe{} etc are not allowed in the lib.rs of the crate. This should take care of accidentially adding these into production code.

## Summary by CodeRabbit

* **Refactor**
  * Replaced panic-prone locking with non-panicking primitives and removed unwrap/expect paths to improve runtime stability.

* **Bug Fixes / Stability**
  * Event loop and HTTP client construction now surface errors instead of panicking.
  * SVG rendering for the Vello backend returns a clear "not implemented" error.

* **Chores**
  * Added a workspace lint configuration and a new synchronization dependency for consistent code quality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gosub-io/gosub-engine/pull/968)

### DIFF
--- a/crates/gosub_instance/src/lib.rs
+++ b/crates/gosub_instance/src/lib.rs
@@ -1,9 +1,3 @@
-#![forbid(unsafe_code)]
-#![deny(clippy::todo)]
-#![deny(clippy::unimplemented)]
-#![deny(clippy::dbg_macro)]
-#![cfg_attr(not(test), deny(clippy::unwrap_used, clippy::expect_used, clippy::panic))]
-
 use gosub_interface::chrome::ChromeHandle;
 use gosub_interface::config::{HasTreeDrawer, ModuleConfiguration};
 use gosub_interface::draw::TreeDrawer;
--- a/crates/gosub_instance/src/lib.rs
+++ b/crates/gosub_instance/src/lib.rs
@@ -1,3 +1,9 @@
+#![forbid(unsafe_code)]
+#![deny(clippy::todo)]
+#![deny(clippy::unimplemented)]
+#![deny(clippy::dbg_macro)]
+#![cfg_attr(not(test), deny(clippy::unwrap_used, clippy::panic))]
+
 use gosub_interface::chrome::ChromeHandle;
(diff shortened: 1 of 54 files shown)


crates/gosub_instance/src/lib.rs
```diff
@@ -1,3 +1,9 @@
+#![forbid(unsafe_code)]
+#![deny(clippy::todo)]
```
Sharktheone: same here, I think todo could be fine.
Also, I'd rather have this as something we define in Cargo.toml. It's also possible to make them only be crate wise there.

Sharktheone (changes requested):
In our root Cargo.toml, we define the following:

```toml
[workspace.lints.clippy]
panic = "deny"
expect_used = "deny"
unwrap_used = "deny"
unimplemented = "deny"
dgb_macro = "deny"
```

In the individual crate Cargo.toml's we then add:

```
[lints]
workspace = true
```

I think this is a much nicer approach than using the macros.

jaytaph:
> I think this is a much nicer approach than using the macros.

The problem is the last item:

`#![cfg_attr(not(test), deny(clippy::unwrap_used, clippy::expect_used, clippy::panic))]`

that is not possible in a cargo.toml file. So we still need to have that added to the lib.rs.  So instead of having things at two different spots, i'd rather have it in one file (per crate)

Sharktheone:
> > I think this is a much nicer approach than using the macros.
> 
> The problem is the last item:
> 
> `#![cfg_attr(not(test), deny(clippy::unwrap_used, clippy::expect_used, clippy::panic))]`
> 
> that is not possible in a cargo.toml file. So we still need to have that added to the lib.rs. So instead of having things at two different spots, i'd rather have it in one file (per crate)

no, you can do that via the clippy.toml as

```toml
allow-unwrap-in-tests = true
allow-expect-in-tests = true
```

and probably also for panics

Sharktheone (approved):

PR #983, opened 2026-05-17
Render pipeline merge

jaytaph:
## Summary by CodeRabbit

* **New Features**
  * Optional render pipeline with rasterized tile blitting; Cairo backend now supports blitted tiles.
  * New screenshot examples that render HTML to PNG.

* **Improvements**
  * Layout engine upgraded to a newer Taffy.
  * Cleaner stylesheet error reporting (no stdout noise).
  * GTK Cairo example accepts an initial URL and enables startup logging.

* **Refactor**
  * Replaced global browser/media/texture singletons with explicit shared per-instance state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gosub-io/gosub-engine/pull/983?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


crates/gosub_engine/src/engine/context.rs
```diff
@@ -184,3 +192,221 @@ impl BrowsingContext {
...
+    }
+    log::info!(
+        "[pipeline] stage 5 painting:      {:>6.1}ms  ({} tiles painted, {} commands total)",
+        t.elapsed().as_secs_f64() * 1000.0,
+        painted_tiles,
+        painted_commands
+    );
+
+    // Stage 6: rasterize
+    #[cfg(feature = "backend_cairo")]
```
Sharktheone, line 323: do we really want cfg flags here? shouldn't this all be done via the module config?

crates/gosub_engine/src/render/backends/cairo.rs
Sharktheone, line 1: this probably also is a todo for some time. I don't think the gosub_engine crate should know anything about the supported backends or something

Sharktheone (changes requested):

crates/gosub_engine/src/engine/context.rs
jaytaph: yes, but that is currently not attached yet..  This actually uses the "new" renderers for now..   the connection will be added when the renderer is in place properly..

crates/gosub_engine/src/render/backends/cairo.rs
jaytaph: yes, these items will be removed to separate crates later on

Sharktheone (approved):

PR #991, opened 2026-05-19
[render-pipeline-docs]

jaytaph:
Part of a stacked diff series.

*Created by pancake.*

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive documentation for the render pipeline, covering the seven-stage architecture (render tree building through compositing), multiple backend implementations (Cairo, Vello, Null), core data structures used across stages, and algorithms for layout, layering, tiling, painting, and rasterization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gosub-io/gosub-engine/pull/991?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #992, opened 2026-05-19
[timings]

jaytaph:

*Created by pancake.*

Added timings to the different parts of the render engine and other parts. This allows us to easily see the timings and find bottlenecks quickly. It uses a separate http server at port 9090, which is behind a compile feature flag (`metrics`). 

image

## Summary by CodeRabbit

* **New Features**
  * Added optional metrics feature enabling HTTP metrics server at `127.0.0.1:9090/metrics` to expose engine performance statistics.
  * Introduced metrics CLI tool with support for querying metrics, resetting counters, JSON output, and watch mode.
  * Enhanced timing instrumentation across rendering pipeline and network operations for detailed performance analysis.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gosub-io/gosub-engine/pull/992?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/gosub_engine/examples/metrics_cli.rs
+++ b/crates/gosub_engine/examples/metrics_cli.rs
@@ -13,34 +13,32 @@ use std::time::Duration;
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
-    let args: Vec<String> = std::env::args().collect();
-
     let mut host = "127.0.0.1".to_string();
     let mut port: u16 = 9090;
     let mut reset = false;
     let mut json_output = false;
     let mut watch_interval: Option<u64> = None;
 
-    let mut i = 1;
-    while i < args.len() {
-        match args[i].as_str() {
+    let mut args = std::env::args().skip(1);
+    while let Some(arg) = args.next() {
+        match arg.as_str() {
             "--host" => {
-                i += 1;
-                host = args.get(i).cloned().unwrap_or(host);
+                if let Some(v) = args.next() {
+                    host = v;
+                }
             }
             "--port" => {
-                i += 1;
-                port = args.get(i).and_then(|p| p.parse().ok()).unwrap_or(port);
+                if let Some(v) = args.next() {
+                    port = v.parse().unwrap_or(port);
+                }
             }
             "--reset" => reset = true,
             "--json" => json_output = true,
             "--watch" => {
-                i += 1;
-                watch_interval = args.get(i).and_then(|s| s.parse().ok()).or(Some(1));
+                watch_interval = args.next().and_then(|s| s.parse().ok()).or(Some(1));
             }
             _ => {}
         }
-        i += 1;
     }
 
     let base = format!("http://{host}:{port}");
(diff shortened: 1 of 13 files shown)


jaytaph:
📚 **Stack:**
- #991 `render-pipeline-docs`
- #992 `timings` ← you are here
- #993 `quick-optimizations`
- #995 `backend-splits`

crates/gosub_engine/examples/metrics_cli.rs
```diff
@@ -0,0 +1,145 @@
...
+//!   cargo run --example metrics_cli -- --watch 2      # poll every 2 seconds
+
+use std::collections::BTreeMap;
+use std::time::Duration;
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    let args: Vec<String> = std::env::args().collect();
+
+    let mut host = "127.0.0.1".to_string();
```
Sharktheone, line 16: not sure if it really makes sense to call .to_string here

crates/gosub_engine/examples/metrics_cli.rs
```diff
@@ -0,0 +1,145 @@
...
+            "--reset" => reset = true,
+            "--json" => json_output = true,
+            "--watch" => {
+                i += 1;
+                watch_interval = args.get(i).and_then(|s| s.parse().ok()).or(Some(1));
+            }
+            _ => {}
+        }
+        i += 1;
+    }
```
Sharktheone: what's the issue with a for loop here? (It can also use &str refs)

Sharktheone: Also, then you don't even have to use args().collect()

crates/gosub_engine/examples/metrics_cli.rs
```diff
@@ -0,0 +1,145 @@
...
+
+fn fmt_us(us: u64) -> String {
+    if us < 1_000 {
+        format!("{us}µs")
+    } else if us < 1_000_000 {
+        format!("{:.1}ms", us as f64 / 1_000.0)
+    } else {
+        format!("{:.2}s", us as f64 / 1_000_000.0)
+    }
+}
```
Sharktheone, line 143: how about just using `Duration.to_string` for that?

Sharktheone (changes requested):
Btw, this also has a few other unrelated changes. Not sure if that is a bug in the stacked pr thing

jaytaph:
> Btw, this also has a few other unrelated changes. Not sure if that is a bug in the stacked pr thing

No, this is a metric cli for benchmarking. It should be here

crates/gosub_engine/examples/metrics_cli.rs
jaytaph: It needs to be a string as it is overwritten with the args (cloned())

crates/gosub_engine/examples/metrics_cli.rs
jaytaph: to_string() does not implement Display, only Debug

Sharktheone (approved):

PR #993, opened 2026-05-19
[quick-optimizations]

jaytaph:

*Created by pancake.*

Optimized the style caching so it won't rebuild on each fetch. Saves a factor 2500(!)

## Summary by CodeRabbit

* **New Features**
  * Added a new metrics CLI example tool.

* **Performance**
  * Improved style computation with per-node caching to speed up lookups and rendering.

* **Chores**
  * Updated example build/dev tooling requirements so example code compiles with HTTP+JSON support.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gosub-io/gosub-engine/pull/993?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


jaytaph:
📚 **Stack:**
- #993 `quick-optimizations` ← you are here
- #995 `backend-splits`
- #997 `new-render-example`

crates/gosub_pipeline/src/common/document/pipeline_doc.rs
```diff
@@ -126,14 +128,30 @@ where
...
+    }
+
+    /// Returns the full computed style list for `id`, fetching from cache when possible.
+    fn cached_styles(&self, id: NodeId) -> Arc<StylePropertyList> {
+        if let Some(cached) = self.style_cache.lock().get(&id) {
+            return cached.clone();
+        }
+        let list = Arc::new(self.compute_styles(id));
+        self.style_cache.lock().insert(id, list.clone());
+        list
```
Sharktheone, line 154: does this have to be Arc? I assume all of this is single threaded? I mean yeah, we can do some things on different thread, but I don't think the caching should be part of that

Sharktheone (changes requested):

crates/gosub_pipeline/src/common/document/pipeline_doc.rs
Sharktheone: btw, then we also don't need the mutex

jaytaph: The arc is not needed, but the Mutex is since the PipelineDocument requires Send + Sync

Sharktheone: Why is PipelineDocument Send+Sync? When looking at other browsers, the document should always stay on the main thread

jaytaph: I would not want to make that assumption at this point of time.

Sharktheone (approved):

Sharktheone (approved):
Still the sync primitives that need to go at some point, but for now this is okay

PR #995, opened 2026-05-20
[backend-splits] Separated the render backends from the pipeline

jaytaph:
The pipeline should be completely backend agnostic and all specific render code for each backend are now stored in seprate cargo_renderer_* crates.

---

*Created by pancake.* 🥞


jaytaph:
📚 **Stack:**
- #993 `quick-optimizations`
- #995 `backend-splits` ← you are here
- #997 `new-render-example`

crates/gosub_pipeline/Cargo.toml
```diff
@@ -36,27 +26,12 @@ file_type = { version = "0.8.1", features = ["httpd"] }
 anyhow = "1.0.102"
 cow-utils = "0.1"
 parking_lot = "0.12"
+parley = "0.3.0"
```
Sharktheone, line 29: why did we go from optional parley to forced parley? I think the pipeline should not use this?

Sharktheone (approved):

crates/gosub_pipeline/Cargo.toml
jaytaph: no that is correct. We shouldn't force parley.. I just added it to make things work, but we should get rid of this as soon as the rendering system is working properly (i think that won't take long).. i'll add an issue for it

Sharktheone (approved):

crates/gosub_pipeline/Cargo.toml
jaytaph: https://github.com/gosub-io/gosub-engine/issues/1007

PR #998, opened 2026-05-21
[css-stylesheet-lookups]

jaytaph:
Updated the stylesheet handling. It's much better as before:

image

Part of a stacked diff series.

*Created by pancake.*

PR #997, opened 2026-05-21
[new-render-example]

jaytaph:
Part of a stacked diff series.

*Created by pancake.*

## Summary by CodeRabbit

* **New Features**
  * New GTK4 browser example demonstrating Cairo-backed rendering and interactive UI (address bar, scrolling, resize).
  * System UI font initialization entrypoint for the Cairo backend.

* **Bug Fixes**
  * More robust error handling across renderers: surface/context failures and unimplemented paint paths no longer crash.
  * Safer fallback rendering with light-grey placeholder when frames are unavailable.

* **Refactor**
  * Font utilities exposed for broader use; internal module reorganization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gosub-io/gosub-engine/pull/997?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1006, opened 2026-05-26
[pipeline-screenshot]

jaytaph:
Added a screenshot tool that will render a complete full website to a PNG file. You can define the width of the viewport, but the height depends on the actual size of the page. (later on we need to see how this would work with things like doomscrolling)

*Created by pancake.* 🥞

## Summary by CodeRabbit

* **New Features**
  * Added a headless screenshot tool for rendering websites to PNG images without a graphical interface

* **Developer Tools**
  * Added a CSS property comparison utility for validating rendered styles against reference browsers

* **Improvements**
  * Enhanced error handling and diagnostic logging in Cairo and Vello renderers

[Review Change Stack](https://app.coderabbit.ai/change-stack/gosub-io/gosub-engine/pull/1006?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,10 @@ name = "egui-vello"
 path = "examples/egui-vello/main.rs"
 
 [[example]]
+name = "pipeline-screenshot"
+path = "examples/pipeline-screenshot/main.rs"
+
+[[example]]
 name = "metrics-cli"
 path = "examples/metrics_cli.rs"
 
--- a/crates/gosub_pipeline/tools/compare_css.py
+++ b/crates/gosub_pipeline/tools/compare_css.py
@@ -1,0 +1,158 @@
+#!/usr/bin/env python3
+"""
+Compare computed CSS properties between gosub and Firefox.
+
+Usage:
+    python3 compare_css.py <gosub.json> <firefox.json>
+
+gosub.json  — produced by running with GOSUB_DUMP_CSS=/tmp/gosub.json
+firefox.json — produced in the browser console:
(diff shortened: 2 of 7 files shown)


jaytaph:
📚 **Stack:**
- #997 `new-render-example`
- #1006 `pipeline-screenshot` ← you are here

crates/gosub_css3/src/system.rs
```diff
@@ -103,7 +103,11 @@ impl CssSystem for Css3System {
                                 // CSS but happen not to have their own PropertyDefinition entry)
                                 // still reach the style consumer.
                                 let value = if let CssValue::List(mut v) = value {
-                                    if v.len() == 1 { v.pop().expect("unreachable") } else { CssValue::List(v) }
+                                    if v.len() == 1 {
+                                        v.pop().expect("unreachable")
```
Sharktheone, line 126: I guess this should be handled somehow differently

Sharktheone (changes requested):

crates/gosub_css3/src/system.rs
jaytaph: i'm not sure why .expect is falling through.. i though we had denied that.

jaytaph: @Sharktheone The linting was not added to this crate as there are 211 violations in it.. i would suggest to leave it for now, but i've will add linting checks for other crates that are compliant in a later PR.

There is an issue to get rid of these and to set up the linting checks for all crates.

Sharktheone (approved):

PR #1028, opened 2026-06-04
Pipeline performance improvements

jaytaph:
## Summary

- Add static font manager
- Move pixel buffers to Arc to reduce cloning
- Add parallel rasterizers for tiles
- Remove redundant pixel buffer copies
- Fix code warnings

> **Note:** This PR is stacked on top of #1027 (`example-fixes`) and should be merged after that one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Skia-based text rasterization implemented across renderers.

* **Bug Fixes**
  * Removed text-rendering placeholder; text now renders correctly and wraps.

* **Performance**
  * Per-tile pixel caching to avoid redundant rasterization.
  * Per-thread font manager caching to speed text layout/render.
  * Shared, zero-copy pixel buffers to reduce memory use and copying.

### DIFF
--- a/crates/gosub_engine/src/engine/context.rs
+++ b/crates/gosub_engine/src/engine/context.rs
@@ -844,7 +844,7 @@ fn pipeline_composite(cache: &PipelineCache, scroll_x: f64, scroll_y: f64, vp_w:
             y: (tile.page_y - scroll_y) as f32,
             w: tile.width,
             h: tile.height,
-            data: (*tile.data).clone(),
+            data: Arc::clone(&tile.data),
         });
         blits += 1;
     }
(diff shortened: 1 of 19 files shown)


crates/gosub_engine/src/engine/context.rs
```diff
@@ -458,6 +479,185 @@ impl RenderContext for BrowsingContext {
 }
 
 /// Runs pipeline stages 1–6 for the **entire page** (all tiles, not just the viewport slice)
+/// Compute a stable cache key for a tile: (page_x bits, page_y bits, layer_id, content hash).
+/// The content hash covers all paint commands so any visual change produces a different key.
+#[cfg(any(feature = "backend_cairo", feature = "backend_skia"))]
+fn tile_cache_key(tile: &gosub_render_pipeline::tiler::Tile) -> TileCacheKey {
```
Sharktheone, line 485: I'm not sure if that makes sense here. We build a TileCacheKey which we then cache again afterwards. I feel like there are better options to do that.

PR #1034, opened 2026-06-04
gosub_config: fixed saving JSON 

jaytaph:
JSON saving works properly

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty map configuration values now parse/serialize correctly and round-trip without producing spurious entries.

* **Changes**
  * Default values updated for select configuration entries.
  * Map formatting/display simplified and serialized output aligned with display formatting.
  * Configuration updates are persisted to storage immediately on change.

* **Breaking Changes**
  * Removed public metadata field `last_accessed`.

* **Tests & Docs**
  * Added round-trip tests and clarified persistence behavior in docs.

### DIFF
--- a/crates/gosub_config/src/storage/json.rs
+++ b/crates/gosub_config/src/storage/json.rs
@@ -8,9 +8,8 @@ use std::fs;
 use std::fs::File;
 use std::io::{Read, Write};
 
-/// JSON file-backed storage adapter. All settings are held in memory and written to a JSON
-/// file on the filesystem. Note: `set` currently only updates the in-memory cache; call
-/// `write_file` explicitly to persist changes to disk.
+/// JSON file-backed storage adapter. All settings are held in memory and written through to the
+/// backing file on every `set` call.
 pub struct JsonStorageAdapter {
     path: String,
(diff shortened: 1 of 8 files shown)


crates/gosub_config/src/storage/json.rs
```diff
@@ -48,8 +47,8 @@ impl StorageAdapter for JsonStorageAdapter {
     }
 
     fn set(&self, key: &str, value: Setting) -> Result<()> {
-        let mut lock = self.elements.lock();
-        lock.insert(key.to_owned(), value);
+        self.elements.lock().insert(key.to_owned(), value);
```
Sharktheone, line 50: I think we should just insert a `key: String` here and let the caller clone the value if necessary

Sharktheone (changes requested):

crates/gosub_config/src/storage/json.rs
jaytaph: if we are going to use key: String, then the caller would always need to clone the string, even when the implementation might not need it (for instance, when we would hash the string). In that case, i'd prefer that the caller sends a regular &str, and we clone only when the implementation needs it.

PR #1121, opened 2026-07-12
perf(bytestream): decode incrementally on append instead of re-scanning

jaytaph:
## Summary by CodeRabbit

* **Performance**
  * Improved incremental text decoding when appending new data, reducing unnecessary reprocessing.
  * Better behavior for incomplete trailing UTF-8 sequences during streaming, continuing when more input arrives and completing on stream close.
  * Enhanced line/column position tracking across incremental updates, including CR/LF boundary handling.

* **Bug Fixes**
  * Fixed decoding of multibyte characters split across appended chunks.
  * Ensured consistent decoded characters and location reporting between incremental and complete decoding.

* **Tests**
  * Added regression coverage for incremental append equivalence and matching location results.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -14,6 +14,11 @@ ignore = [
     # (resvg -> usvg -> fontdb) and font shaping (ab_glyph -> owned_ttf_parser).
     # Nothing we control depends on it directly.
     "RUSTSEC-2026-0192", # ttf-parser unmaintained
+    # `rustybuzz` is unmaintained (maintainer declared it EOL) but pulled in transitively
+    # by SVG rendering (resvg -> usvg -> rustybuzz). No safe upgrade is available; the
+    # maintained alternative (harfrust) requires resvg/usvg to migrate first. Nothing we
+    # control depends on it directly.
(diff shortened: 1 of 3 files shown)


deny.toml
```diff
@@ -14,6 +14,11 @@ ignore = [
     # (resvg -> usvg -> fontdb) and font shaping (ab_glyph -> owned_ttf_parser).
     # Nothing we control depends on it directly.
     "RUSTSEC-2026-0192", # ttf-parser unmaintained
+    # `rustybuzz` is unmaintained (maintainer declared it EOL) but pulled in transitively
+    # by SVG rendering (resvg -> usvg -> rustybuzz). No safe upgrade is available; the
+    # maintained alternative (harfrust) requires resvg/usvg to migrate first. Nothing we
+    # control depends on it directly.
+    "RUSTSEC-2026-0206", # rustybuzz unmaintained
```
Sharktheone, line 21: I guess this just means we should switch to HarfRust if possible

Sharktheone (approved):

PR #1123, opened 2026-07-12
converted the definitions generator from go to rust

jaytaph:
This means we can remove the go language from our toolchain in order to compile the engine.

## Summary by CodeRabbit

* **New Features**
  * Added a Rust-based CSS definition generator.
  * Generates deterministic CSS definition files by combining MDN and Webref data.
  * Supports consolidated or split definition outputs with local caching.

* **Documentation**
  * Updated the generator documentation for the Rust workflow, including data sources, caching behavior, output paths, and determinism guarantees.

* **Refactor**
  * Replaced the legacy Go-based generator, removing its prior patching and data-processing components.

* **Chores**
  * Updated Rust security advisory ignore configuration for a transitive crate.


Sharktheone:
Note here, we do not need go to compile the engine as the definition files are vendored

Sharktheone (approved):

jaytaph:
> Note here, we do not need go to compile the engine as the definition files are vendored

Yes, that is true. But we need to make sure they stay a bit up to date, and running it with rust makes a bit more sense than using an external go toolchain.  Not a critical thing, but still...

PR #1130, opened 2026-07-16
removed em-dashes from most of the comments and texts

jaytaph:
Ok, so this is currently the sad state of AI usage in projects. We/I feel that AI is a very robust tool to help us with defining problems, laying out solutions, and implement them. All under our guidance and inspection.

It works faster and results often in better code than I could personally write myself. Not because I have no clue what i'm doing (i hope i have by now), but mostly because I literally didn't think of solving some problems in a way that an LLM came up with. And sometimes, it finds ways that are too ridiculous and with some rewrites we got it back working. Most of the time, I feel like it uses the same solution i would do myself, only it's 10 times faster in producing it.

Anyhow, there are many different ways on how to use AI, and I think is very dangerous that the current mindset by too many people is that when you use AI, it must be slop. It just screams insecurity in my opinion.

So the first thing that people will check on ANY repo: is there even a miniscule trace of AI that I can detect in this repo? If so, i will not be looking any further and will scream AI-slop at the nearest forum i can find.

This is why i'm removing the em-dashes, the number one telltale on AI generated content/comments. Not because I don't approve the comments itself (i think most of them are fine, but it's a big project, and there are bound to be some spillage here and there that we should change/update), but I don't approve of the way people will NOT read the comments, but rather fall over the em-dash being present.

Without them, it might make people actually read the comments, the code, and decide for themselves if something is AI slop or not.

I don't like that this is the way we have to work at this point in time, but it is the way it is.

## Summary by CodeRabbit

- **Documentation**
  - Standardized punctuation, dash usage, wording, and line wrapping across project documentation and inline comments.
  - Improved clarity and consistency in explanatory notes for CSS, rendering, layout, engine, storage, and example components.
- **Refactor**
  - No functional, API, parsing, rendering, layout, or runtime behavior changes.

### DIFF
--- a/bin/gosub-screenshot/main.rs
+++ b/bin/gosub-screenshot/main.rs
@@ -1,7 +1,7 @@
 //! Headless screenshot tool: loads a URL through the full gosub render pipeline and
 //! saves the result as a PNG without opening a window.
 //!
-//! Uses the Skia backend for **CPU** rasterization — no GPU, no wgpu adapter, and no
+//! Uses the Skia backend for **CPU** rasterization - no GPU, no wgpu adapter, and no
 //! system libraries (skia-safe is statically linked). The page is rasterized into small
 //! cached tiles (`ExternalHandle::TileCache`) which we composite here, so there is no
 //! GPU texture-size limit and pages of any height can be captured.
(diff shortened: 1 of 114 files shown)


eandersons:
Slop is not just badly regurgiated code, it also embodies hidden negative impacts (environmental, social, economical and so on) of the current implementation of LLMs. Even utilising LLMs in code reviews can be enough reason for a lot of people to consider a project being slopped or tainted (read: using LLMs in any way thus adding to already enourmous negative impact).

So I woud say that em-dashes are not the biggest issue in relation to the project being cosidered as slop. A lot of people have been using them long before LLMs started to spit out slop at a scale, so em-dashes alone cannot be considered as enough to judge some project being a slop.

Removing em-dashes from the project just for it not being called slop radiates some insecurities.

PR #1153, opened 2026-08-03
JS: Implemented functionality to the js-api bridge

jaytaph:
More functionality added to the js-api bridge. This means we connected some of the javascript APIs into the engine so we can use it. Even though we don't have v8 or any JS engine hooked into the engine, we still are able to test this part.

We try to use as many WPT tests as possible to make sure we follow the correct standards.

## Summary by CodeRabbit

* **New Features**
  * Added browser-style APIs for Base64, DOM exceptions, headers, text encoding/decoding, URLs, events, abort signals, and Web Storage.
  * Added a Web Platform Test runner with JavaScript API interoperability, timers, console output, and file-backed fetching.
* **Testing**
  * Added optional Web Platform Tests conformance coverage with expected-failure tracking.
  * Expanded unit tests across the new web APIs.
* **Bug Fixes**
  * Improved callback scope handling for more reliable error reporting.


Sharktheone:
I think this might go a bit too far for now. We don't have event stuff yet. I would rather get some more experience about all of this. I need to review more of this, but I already so some things which may have side effects that are not intended

jaytaph:
We dont need to merge this yet. I was working on using the wpt testsuite to test the console api that was already present. Ive added more functionality that could be implemented and tested without an actual js connection. (It merely tests any engine to rust, not the engine to rust to engine. That would indeed come later

Sharktheone:
I think the console API is something we don't really need to test yet. Afaik it mostely contains a unfinished version of console.log

jaytaph:
> I think the console API is something we don't really need to test yet. Afaik it mostely contains a unfinished version of console.log

Actually, it is a almost fully implementation of the console api. Only fmt and table are not completely implemented. It follows the spec completely ( which is why i wanted to test against wpt)

jaytaph:
Fixed in #1242 and #1244

PR #1162, opened 2026-08-09
Rasterize viewport first, extend on scrolling

jaytaph:
When we have a large page, we do not need to create tiles for the whole page yet. We only need to make sure we generate the tiles for the viewport.

To speed things up with scrolling, we also create a few tiles after the viewport so we try to keep ahead of scrolling this way. Old tiles outside the viewport are stored in the cache and uses an LRU system to clear once the cache is full.

Stacked upon #1155

## Summary by CodeRabbit

- **Performance**
  - Improved scrolling through content that has not yet been rasterized.
  - Reuses cached tiles and pixel data for more efficient updates.
  - Defers off-screen rendering and rasterizes content as it enters view.
  - Added memory controls to limit rendering-cache usage while preserving visible content.
- **Bug Fixes**
  - Improved tile ordering and repaint behavior during scrolling and hover updates.
  - Prevented scroll-only updates until required content is ready for display.
  - Improved handling of cached content at page edges and during repeated scrolling.
- **Navigation & Interaction**
  - Added fragment-target navigation support.
  - Improved context-menu hit testing and cursor-shape updates.

### DIFF
--- a/crates/gosub_engine/src/engine/context.rs
+++ b/crates/gosub_engine/src/engine/context.rs
@@ -22,12 +22,21 @@ use crate::html::RenderConfiguration;
 use gosub_interface::css3::{CssSystem, HoverFingerprints};
 use gosub_interface::document::Document as _;
 use gosub_interface::node::NodeType;
+use gosub_render_pipeline::common::browser_state::{BrowserState, WireframeState};
+use gosub_render_pipeline::common::document::pipeline_doc::GosubDocumentAdapter;
+use gosub_render_pipeline::common::geo::{Dimension as PipelineDimension, Rect as PipelineRect};
+use gosub_render_pipeline::common::media::MediaStore;
 use gosub_render_pipeline::common::texture::TilePixels;
-use gosub_render_pipeline::layering::layer::LayerList;
-use gosub_render_pipeline::layouter::LayoutElementId;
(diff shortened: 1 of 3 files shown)


Sharktheone:
is this supposed to have the same changes as the tile cache again?

jaytaph:
> is this supposed to have the same changes as the tile cache again?

Not really.. it is stacked...  github still can't handle that properly...  i've merged the first part, so this will be shown correctly shortly

crates/gosub_engine/src/engine/context.rs
```diff
@@ -944,6 +983,96 @@ fn pipeline_build_cache<C: RenderConfiguration>(
...
+    strategy: RasterStrategy,
+    prev_tile_cache: TilePixelCache,
+    media_store: Arc<gosub_render_pipeline::common::media::MediaStore>,
+    tile_size: f64,
+) -> PipelineCache {
+    use gosub_render_pipeline::common::geo::{Dimension as PipelineDimension, Rect as PipelineRect};
+    use gosub_render_pipeline::tile_budget::defer_tiles_outside_window;
+    use gosub_render_pipeline::tiler::{TileList, TileState};
+    use gosub_shared::{timing_start, timing_stop};
+
```
Sharktheone: not a big fan of imports inside a function

PR #1229, opened 2026-09-05
Fixes ghsa c762 mxfh vwvp

jaytaph:
Since we are currently not live yet and haven't a real current release, we can do this in the open.

This fixes security advisory [GHSA-c762-mxfh-vwvp](https://github.com/gosub-io/gosub-engine/security/advisories/GHSA-c762-mxfh-vwvp) which has been reported by @Szczurowsky and @xProXyx.

The SVG parser has no recursion limit: roxmltree's tokenizer recurses once per level of element nesting and neither its nodes_limit nor usvg own 1024-level guard stops it. A few kilobytes of `<g><g><g>...` exhausts the stack and aborts the process.

Fixed by:                                                                                                                                                                                      
   - Reject SVG nesting deeper than 128 levels before parsing, and refuse entity declarations                                                                                                    
     whose replacement text contains markup. The DOCTYPE is read as a unit with its quoted                                                                                                       
     literals tracked; both review-found bypasses came from not doing that.
   - Grow the stack with `stacker::maybe_grow` for the parse instead of spawning a thread,
     so the depth limit holds regardless of how deep the caller already is.
   - Serialize the document iteratively (inline SVGs).
   - Resolve inherited styles iteratively instead of recursively.
   - Cap layout nesting at 128 levels.
   - Differential test and cargo-fuzz target comparing the scanner against the real parser.

## Bug Fixes
- Improved handling of deeply nested HTML and SVG documents to prevent stack overflows during parsing, rendering, layout, and serialization.
- Added safeguards that reject excessively nested or entity-expanding SVG/SVGZ content with clear parse errors.
- Preserved document order and closing tags when serializing deeply nested content.
- Corrected inherited style resolution, including font-relative units.
- Limited excessively deep layout trees to maintain reliable rendering.

## Tests
- Added regression coverage for nested HTML, SVG, SVGZ, and entity-expansion inputs, plus style and layout behavior.
- Added differential testing and fuzzing coverage for SVG nesting and XML limits.

### DIFF
--- a/crates/gosub_render_pipeline/src/common/media/decoder/svg.rs
+++ b/crates/gosub_render_pipeline/src/common/media/decoder/svg.rs
@@ -1,9 +1,13 @@
 use super::{DecodedMedia, ImageDecodeError, MediaDecoder};
+use gosub_shared::svg_limits::{xml_nesting_depth_exceeds, MAX_SVG_NESTING_DEPTH, SVG_PARSE_STACK_SIZE};
 use resvg::usvg;
 use std::sync::{Arc, OnceLock};
 
 /// Number of leading bytes scanned when sniffing for an SVG root element.
 const SVG_SNIFF_LEN: usize = 1024;
+
+/// Leading bytes of a gzip member, i.e. an SVGZ file.
+const GZIP_MAGIC: [u8; 2] = [0x1f, 0x8b];
 
 /// `usvg::Options` backed by a shared fontdb, built once and reused so system font discovery
 /// happens only once per process.
@@ -59,7 +63,48 @@ impl MediaDecoder for SvgDecoder {
     }
 
     fn decode(&self, bytes: &[u8]) -> Result<DecodedMedia, ImageDecodeError> {
-        let tree = usvg::Tree::from_data(bytes, &svg_options()).map_err(|e| ImageDecodeError::Decode(e.to_string()))?;
-        Ok(DecodedMedia::Vector(Box::new(tree)))
+        Ok(DecodedMedia::Vector(Box::new(parse_svg(bytes)?)))
     }
+}
+
+/// Parse SVG bytes into a `usvg::Tree`, bounding both nesting depth and stack.
+///
+/// `usvg::Tree::from_data` would do the gzip step for us, but then the depth check would run on
+/// the compressed bytes and a few hundred bytes of SVGZ would walk straight past it. Hence
+/// decompress, measure, parse, in that order.
+fn parse_svg(bytes: &[u8]) -> Result<usvg::Tree, ImageDecodeError> {
+    let decompressed;
+    let bytes = if bytes.starts_with(&GZIP_MAGIC) {
+        decompressed = usvg::decompress_svgz(bytes).map_err(|e| ImageDecodeError::Decode(e.to_string()))?;
+        decompressed.as_slice()
+    } else {
+        bytes
+    };
+
+    // Has to happen before `from_str`: the recursion it bounds is inside the parser.
+    if xml_nesting_depth_exceeds(bytes, MAX_SVG_NESTING_DEPTH) {
+        return Err(ImageDecodeError::Decode(format!(
+            "SVG nests elements deeper than the {MAX_SVG_NESTING_DEPTH} level limit"
+        )));
+    }
+
+    let text = std::str::from_utf8(bytes).map_err(|_| ImageDecodeError::Decode("SVG is not valid UTF-8".into()))?;
+
+    // Own thread, so the depth limit above translates into an actual stack budget. Callers
+    // cannot provide one: `<img src=…svg>` decodes on a default-stack fetch thread, an inline
+    // `<svg>` decodes partway down a recursive layout walk on a tokio worker. `svg_options()`
+    // is built inside the closure because `usvg::Options` holds non-`Send` resolver closures;
+    // the fontdb behind it is a `OnceLock`-shared `Arc`, so system fonts are still scanned once.
+    std::thread::scope(|scope| {
+        let parse = std::thread::Builder::new()
+            .name("svg-parse".into())
+            .stack_size(SVG_PARSE_STACK_SIZE)
+            .spawn_scoped(scope, || usvg::Tree::from_str(text, &svg_options()))
+            .map_err(|e| ImageDecodeError::Decode(format!("could not spawn SVG parse thread: {e}")))?;
+
+        parse
+            .join()
+            .map_err(|_| ImageDecodeError::Decode("SVG parser panicked".into()))?
+            .map_err(|e| ImageDecodeError::Decode(e.to_string()))
+    })
 }
--- a/crates/gosub_shared/src/svg_limits.rs
+++ b/crates/gosub_shared/src/svg_limits.rs
@@ -1,0 +1,257 @@
+//! Depth limit for untrusted SVG.
+//!
+//! Parsing SVG costs a stack frame per level of element nesting: roxmltree's tokenizer recurses
+//! between `parse_element` and `parse_content`, and nothing below us bounds that. Its
+//! `nodes_limit` counts nodes rather than depth, and usvg's own 1024-level guard sits in a later
+//! pass the tokenizer never reaches. A few kilobytes of `<g><g><g>...` will therefore abort the
+//! process on a stack overflow (GHSA-c762-mxfh-vwvp).
+//!
+//! [`xml_nesting_depth_exceeds`] rejects those documents before the parser sees them.
+//! [`SVG_PARSE_STACK_SIZE`] is the other half: the depth limit only bounds the number of frames,
+//! and callers cannot say how much of their own stack is already spent.
+
+/// Maximum element nesting depth accepted in an SVG document.
+///
+/// The 5887 SVG files in this machine's icon themes nest at most 8 levels deep, so this is not a
+/// limit real content runs into. Unoptimised, the tokenizer spends roughly 15 KiB of stack per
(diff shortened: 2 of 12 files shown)


crates/gosub_render_pipeline/src/common/media/decoder/svg.rs
```diff
@@ -59,7 +63,48 @@ impl MediaDecoder for SvgDecoder {
...
     }
+
+    let text = std::str::from_utf8(bytes).map_err(|_| ImageDecodeError::Decode("SVG is not valid UTF-8".into()))?;
+
+    // Own thread, so the depth limit above translates into an actual stack budget. Callers
+    // cannot provide one: `<img src=…svg>` decodes on a default-stack fetch thread, an inline
+    // `<svg>` decodes partway down a recursive layout walk on a tokio worker. `svg_options()`
+    // is built inside the closure because `usvg::Options` holds non-`Send` resolver closures;
+    // the fontdb behind it is a `OnceLock`-shared `Arc`, so system fonts are still scanned once.
+    std::thread::scope(|scope| {
```
Sharktheone, line 98: I don't really like this being an own thread I mean we just wait for the other thread to finish, so this is pure overhead. And from the stack perspective, i think there are better options than to have one full thread for a single svg

crates/gosub_shared/src/svg_limits.rs
```diff
@@ -0,0 +1,257 @@
...
+/// sees each declaration once, so a document carrying an internal subset gets its budget divided
+/// by this to stay an upper bound.
+const ENTITY_NESTING_LIMIT: usize = 10;
+
+/// Whether `xml` nests elements deeper than `max`.
+///
+/// Answers the depth question only, and answers it high: anything the scan cannot account for is
+/// counted as nesting rather than skipped, so a document it accepts really is within `max`.
+/// Malformed input may be rejected, which the parser would have done anyway.
+pub fn xml_nesting_depth_exceeds(xml: &[u8], max: usize) -> bool {
```
Sharktheone, line 36: hmm, I would like this to happen during parsing, not beforehand. We should optimize for the happy case here, not for the bad case

crates/gosub_shared/src/svg_limits.rs
```diff
@@ -0,0 +1,257 @@
...
+}
+
+/// Whether the DOCTYPE carries an internal subset (`<!DOCTYPE svg [ ... ]>`), the only place a
+/// document can declare its own entities.
+fn has_internal_dtd_subset(xml: &[u8]) -> bool {
+    let Some(start) = find(xml, b"<!DOCTYPE") else {
+        return false;
+    };
+    // The subset opens before the DOCTYPE's own `>`.
+    xml[start..].iter().take_while(|&&b| b != b'>').any(|&b| b == b'[')
```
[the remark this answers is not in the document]
Sharktheone: I think that makes sense, that would also be one of my considerations here

crates/gosub_svg/src/lib.rs
```diff
@@ -24,11 +25,33 @@ pub struct SVGDocument {
...
+                .name("svg-parse".into())
+                .stack_size(SVG_PARSE_STACK_SIZE)
+                .spawn_scoped(scope, || usvg::Tree::from_str(svg, &svg_options()))
+                .map_err(|e| Error::Parse(format!("could not spawn SVG parse thread: {e}")))?;
+
+            parse
+                .join()
+                .map_err(|_| Error::Parse("SVG parser panicked".into()))?
+                .map_err(|e| Error::Parse(e.to_string()))
+        })?;
```
Sharktheone, line 53: Same here, don't like us spawning a full thread just for parsing an SVG

crates/gosub_shared/src/svg_limits.rs
```diff
@@ -0,0 +1,257 @@
...
+/// if `close` never arrives.
+fn skip_delimited(xml: &[u8], open: &[u8], close: &[u8]) -> Option<usize> {
+    if !xml.starts_with(open) {
+        return None;
+    }
+    Some(match find(&xml[open.len()..], close) {
+        Some(at) => open.len() + at + close.len(),
+        None => xml.len(),
+    })
+}
```
Sharktheone, line 113: I think this has the same comment bug. We cannot really say that without parsing it. I also don't like checking that before parsing since it makes the happy case way more expensive

crates/gosub_shared/src/svg_limits.rs
```diff
@@ -0,0 +1,257 @@
...
+        match b {
+            b'"' | b'\'' => quote = Some(b),
+            b'>' => return (offset + 1, last == b'/'),
+            _ if b.is_ascii_whitespace() => {}
+            _ => last = b,
+        }
+    }
+
+    (xml.len(), false)
+}
```
Sharktheone, line 140: Same here, comments

Sharktheone (changes requested):
We also need a follow up PR which fixes this for other parsers such as HTML, CSS nesting, ...

crates/gosub_render_pipeline/src/common/media/decoder/svg.rs
jaytaph: The reason for a spawned thread is that we are doing the parsing with a clean stack, allowing us to use 128 recursions (limited to that number). This is way more than enough.

Without the thread, we do not have enough room and some tests show we can have a limit of 32 max. I'm happy to see if we can find a way to get rid of the recursion, but we cannot get rid of everything as usvg internally use recursion as well.

Sharktheone: I'm really not a fan of needing a own thread. I guess what we could to is replace the stack of the current thread with such an scope API. I just really don't like needing a full thread and blocking on it.

It would even be better if we'd have a pool of threads which we initialize when we need them and destroy when we expect to not need parsing anymore

crates/gosub_shared/src/svg_limits.rs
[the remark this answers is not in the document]
jaytaph: Fixed

crates/gosub_render_pipeline/src/common/media/decoder/svg.rs
jaytaph: I've moved to `stacker::maybe_grow` that will increase the stack if not enough room is available. We keep the recursionlimit  capped.

crates/gosub_shared/src/svg_limits.rs
jaytaph: It's measured less than 3%.  We should open another ticket if we need to optimize this.

crates/gosub_shared/src/svg_limits.rs
jaytaph: I think it would make more sense to have this fixed in the roxmltree upstream. For now, the happy path is less than 3% increase, so i don't think it should be blocking for this security fix for now.

crates/gosub_svg/src/lib.rs
jaytaph: stacker::maybe_grow is now used instead.

crates/gosub_shared/src/svg_limits.rs
jaytaph: It's measurable not THAT more expensive, but yes, it's a small piece of extra work.  We fixed the docdata issue, which is much more complicated than this. It's tested against a few examples, but we could add fuzzing here to make sure it works with basically everything

jaytaph: We could add a differential test here where we check our depth-matcher against the parsed document.

crates/gosub_shared/src/svg_limits.rs
jaytaph: See https://github.com/gosub-io/gosub-engine/pull/1229#discussion_r4036950485

jaytaph:
Fuzzing ran against the XML checker:

```
#17456176       REDUCE cov: 2431 ft: 11026 corp: 3466/1033Kb lim: 4096 exec/s: 40595 rss: 494Mb L: 15/4080 MS: 1 EraseBytes-
#17458002       REDUCE cov: 2431 ft: 11026 corp: 3466/1033Kb lim: 4096 exec/s: 40600 rss: 494Mb L: 92/4080 MS: 1 EraseBytes-
#17458653       REDUCE cov: 2431 ft: 11026 corp: 3466/1033Kb lim: 4096 exec/s: 40601 rss: 494Mb L: 345/4080 MS: 1 EraseBytes-
#17460928       REDUCE cov: 2431 ft: 11026 corp: 3466/1033Kb lim: 4096 exec/s: 40606 rss: 494Mb L: 775/4080 MS: 5 ChangeBit-ShuffleBytes-ShuffleBytes-ShuffleBytes-EraseBytes-
```

No issues found

jaytaph:
> We also need a follow up PR which fixes this for other parsers such as HTML, CSS nesting, ...

HTML parsing is fixed in this PR.  CSS is checked and was already capped. (blocks, var() and @import depths)

Sharktheone (commented):
I really wonder if the best fix might really just be checking the nesting level and not allocating things over and over. The decision I would make this to is the performance of the html and svg parser and everything else we changed here

jaytaph:
@Sharktheone We are waiting on the fix in roxmltree (https://github.com/RazrFalcon/roxmltree/issues/154). Once that is fixed in the crate, we can upgrade and we can remove the whole scanning.. For now, i don't think we have any issues and i don't think it will take months until we see a new release.

Sharktheone:
I mean for the usvg case, we probably don't have many other options. 

The writer still uses a manually allocated vec![], but maybe that just isn't relevant since the writer won't be used that often  

I like the option with the stack resizing a lot more.

PR #1234, opened 2026-09-10
Data urls

jaytaph:
Implementing data-urls (data:....) inline data. The issue is that we do not need to fetch the resource as the data is already present, but we catch these urls (same as with file://). The handler will just respond by creating a fabricated FetchResult.

## Summary by CodeRabbit

- **New Features**
  - Added direct loading of `data:` URLs, including percent-encoded and Base64-encoded content.
  - Supports case-insensitive Base64 markers, unpadded Base64, escaped padding, and inferred `text/plain` media types.
  - Preserves decoded media types and content for inline resources.

- **Bug Fixes**
  - URL fragments are excluded from inline payloads.
  - Malformed or cancelled inline requests are handled safely.
  - Inline resources no longer trigger unnecessary network fetches.

### DIFF
--- a/crates/gosub_engine/src/net/data_url.rs
+++ b/crates/gosub_engine/src/net/data_url.rs
@@ -1,0 +1,265 @@
+//! `data:` URLs, decoded where they are asked for rather than fetched.
+//!
+//! A `data:` URL carries its own bytes, so there is nothing to request: no connection, no
+//! cache, no referrer, no policy to apply. That makes it the odd one out in a network stack
+//! built around submitting a request and waiting, and the reason it is handled here instead
+//! of in the fetcher: a consumer that hits one decodes it and is done.
+//!
+//! The syntax is RFC 2397: `data:[<mediatype>][;base64],<data>`. An absent media type means
+//! `text/plain;charset=US-ASCII`, and without `;base64` the payload is percent-encoded.
+
+use crate::net::types::{FetchRequest, FetchResult, FetchResultMeta, NetError};
+use base64::Engine as _;
+use bytes::Bytes;
+use gosub_sonar::net::events::NetEvent;
+use gosub_sonar::net::observer::NetObserver;
+use http::HeaderMap;
+use std::sync::Arc;
+use url::Url;
+
+/// The default per RFC 2397, for a URL that names no media type.
+const DEFAULT_MEDIA_TYPE: &str = "text/plain;charset=US-ASCII";
+
+/// Whether this request is for the engine-served `data:` scheme.
+///
+/// gosub-sonar only speaks http(s), so the I/O thread answers these itself, the same way it
+/// answers `file://`. Without this a `data:` stylesheet, script or font reaches the zone
+/// fetcher and fails as an unsupported scheme - only images worked, because the media source
+/// decodes them before ever submitting a request.
+pub fn handles(req: &FetchRequest) -> bool {
+    req.url.scheme() == "data"
+}
+
+/// Answer a `data:` request from the URL itself.
+///
+/// No policy gate, unlike `file://`: the bytes travel inside the URL, so serving one reads
+/// neither the disk nor the network and crosses no privilege boundary. The worst a malformed
+/// one can do is fail to decode.
+pub async fn serve(req: &FetchRequest, observer: Arc<dyn NetObserver + Send + Sync>) -> FetchResult {
+    let url = req.url.clone();
+    observer.on_event(NetEvent::Started { url: url.clone() });
+    let started = std::time::Instant::now();
+
+    let Some((content_type, body)) = decode(&url) else {
+        let err = NetError::Io(Arc::new(std::io::Error::new(
+            std::io::ErrorKind::InvalidData,
+            "malformed data: URL",
+        )));
+        observer.on_event(NetEvent::Failed {
+            url,
+            error: anyhow::anyhow!(err.clone()),
+        });
+        return FetchResult::Error(err);
+    };
+
+    let body = Bytes::from(body);
+    let len = body.len() as u64;
+    let mut headers = HeaderMap::new();
+    if let Ok(v) = content_type.parse() {
+        headers.insert(http::header::CONTENT_TYPE, v);
+    }
+    if let Ok(v) = len.to_string().parse() {
+        headers.insert(http::header::CONTENT_LENGTH, v);
+    }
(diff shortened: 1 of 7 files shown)


crates/gosub_engine/src/net/data_url.rs
```diff
@@ -0,0 +1,336 @@
...
+/// Base64 as a `data:` URL is allowed to write it: padding optional rather than required, and
+/// the leftover bits of an unpadded final chunk dropped rather than rejected. This is the
+/// forgiving-base64 decode of the Infra standard, which is what the data URL processor calls
+/// for - a stricter engine turns URLs every other browser accepts into failed resources.
+const BASE64: base64::engine::general_purpose::GeneralPurpose = base64::engine::general_purpose::GeneralPurpose::new(
+    &base64::alphabet::STANDARD,
+    base64::engine::GeneralPurposeConfig::new()
+        .with_decode_padding_mode(base64::engine::DecodePaddingMode::Indifferent)
+        .with_decode_allow_trailing_bits(true),
+);
```
Sharktheone, line 37: I would prefer importing `GeneralPurpose`, `GeneralPurposeConfig` and `DecodingPaddingMode`

crates/gosub_engine/src/net/data_url.rs
```diff
@@ -0,0 +1,336 @@
...
+    req.url.scheme() == "data"
+}
+
+/// Answer a `data:` request from the URL itself.
+///
+/// No policy gate, unlike `file://`: the bytes travel inside the URL, so serving one reads
+/// neither the disk nor the network and crosses no privilege boundary. The worst a malformed
+/// one can do is fail to decode.
+pub async fn serve(req: &FetchRequest, observer: Arc<dyn NetObserver + Send + Sync>) -> FetchResult {
+    let url = req.url.clone();
```
Sharktheone, line 55: Just for a general optimization, it might make sense to reference count urls in the future

crates/gosub_engine/src/net/data_url.rs
```diff
@@ -0,0 +1,336 @@
...
+pub fn decode(url: &Url) -> Option<(String, Vec<u8>)> {
+    if url.scheme() != "data" {
+        return None;
+    }
+
+    // The payload is the URL serialized with its *fragment excluded*, so a `#` ends the data:
+    // `data:text/plain,a#b` carries `a`. `Url` has already filed the fragment separately, so
+    // excluding it is a matter of not putting it back. The query does serialize, and so is put
+    // back: `data:text/plain,a?b` carries `a?b`.
+    let mut rest = url.path().to_string();
```
Sharktheone, line 123: Do we really need to clone the string here? Maybe we can just use the internal buffer, we might have to switch to iterators depending on how the layout of URL is. 
For the Query below, that would just be an iterator chain

crates/gosub_engine/src/net/data_url.rs
```diff
@@ -0,0 +1,336 @@
...
+    // Percent-decoding comes first and applies to the whole payload, base64 or not. That is
+    // what makes `data:text/plain;base64,SGVsbG8%3D` - an author or a serializer escaping the
+    // padding - decode rather than fail.
+    let body = percent_decode(payload);
+
+    // `;base64` is a parameter name, and parameter names are case-insensitive: `;BASE64` marks
+    // the payload just as well.
+    let (media_type, bytes) = match meta.rsplit_once(';') {
+        Some((head, param)) if param.trim().eq_ignore_ascii_case("base64") => {
+            let cleaned: Vec<u8> = body.into_iter().filter(|b| !b.is_ascii_whitespace()).collect();
```
Sharktheone, line 141: It seems we are doing another clone here, I'd rather not do that. Maybe we can just use the stack for this operation and fallback to the heap if needed

crates/gosub_engine/src/net/data_url.rs
```diff
@@ -0,0 +1,336 @@
...
+            (head.trim_end(), BASE64.decode(cleaned).ok()?)
+        }
+        _ => (meta, body),
+    };
+
+    let media_type = if media_type.is_empty() {
+        DEFAULT_MEDIA_TYPE.to_string()
```
Sharktheone, line 148: We also need to start using Cow<str>, or an optimized string type in general. I'd be up to creating one. Using the std string isn't really optimal for a browser use case

crates/gosub_engine/src/net/data_url.rs
```diff
@@ -0,0 +1,336 @@
...
+pub fn handles(req: &FetchRequest) -> bool {
+    req.url.scheme() == "data"
+}
+
+/// Answer a `data:` request from the URL itself.
+///
+/// No policy gate, unlike `file://`: the bytes travel inside the URL, so serving one reads
+/// neither the disk nor the network and crosses no privilege boundary. The worst a malformed
+/// one can do is fail to decode.
+pub async fn serve(req: &FetchRequest, observer: Arc<dyn NetObserver + Send + Sync>) -> FetchResult {
```
Sharktheone, line 54: Why is this async? It does not need to be


```suggestion
pub fn serve(req: &FetchRequest, observer: Arc<dyn NetObserver + Send + Sync>) -> FetchResult {
```

crates/gosub_engine/src/net/io_runtime.rs
```diff
@@ -127,6 +127,32 @@ impl IoRouter {
...
+            // A navigation cancelled while its request sat in the I/O queue must not go on to
+            // emit network events for it. Decoding never awaits, so there is no point inside
+            // `serve` to cancel at: the check belongs here, ahead of the first event. Dropping
+            // `reply_tx` is how the zone fetcher reports a cancelled fetch too.
+            if cancel.is_cancelled() {
+                return;
+            }
+            let result = crate::net::data_url::serve(&req, observer).await;
+            let _ = reply_tx.send(result);
+        });
```
Sharktheone, line 153: Once serve is sync again, i reckon, we can just remove this spawn directly and handle it fully within serve_data_url_request


```suggestion
        // A navigation cancelled while its request sat in the I/O queue must not go on to
        // emit network events for it. Decoding never awaits, so there is no point inside
        // `serve` to cancel at: the check belongs here, ahead of the first event. Dropping
        // `reply_tx` is how the zone fetcher reports a cancelled fetch too.
        if cancel.is_cancelled() {
            return;
        }
        let result = crate::net::data_url::serve(&req, observer);
        let _ = reply_tx.send(result);
```

